PR #2, opened 2017-12-11
Clang fixes

Artem-B:
Make CUTLASS compilable with clang and tweak few minor things to enforce inlining where necessary.

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -41,51 +41,64 @@ endif
 
 ifeq (70, $(findstring 70, $(SM_ARCH)))
     SM_TARGETS 	+= -gencode=arch=compute_70,code=\"sm_70,compute_70\"
+    CLANG_SM_TARGETS += --cuda-gpu-arch=sm_70
 endif
 ifeq (62, $(findstring 62, $(SM_ARCH)))
     SM_TARGETS 	+= -gencode=arch=compute_62,code=\"sm_62,compute_62\"
+    CLANG_SM_TARGETS += --cuda-gpu-arch=sm_62
 endif
 ifeq (61, $(findstring 61, $(SM_ARCH)))
     SM_TARGETS 	+= -gencode=arch=compute_61,code=\"sm_61,compute_61\"
+    CLANG_SM_TARGETS += --cuda-gpu-arch=sm_61
 endif
 ifeq (60, $(findstring 60, $(SM_ARCH)))
     SM_TARGETS 	+= -gencode=arch=compute_60,code=\"sm_60,compute_60\"
+    CLANG_SM_TARGETS += --cuda-gpu-arch=sm_60
 endif
 ifeq (52, $(findstring 52, $(SM_ARCH)))
     SM_TARGETS 	+= -gencode=arch=compute_52,code=\"sm_52,compute_52\"
+    CLANG_SM_TARGETS += --cuda-gpu-arch=sm_52
 endif
 ifeq (37, $(findstring 37, $(SM_ARCH)))
     SM_TARGETS 	+= -gencode=arch=compute_37,code=\"sm_37,compute_37\"
+    CLANG_SM_TARGETS += --cuda-gpu-arch=sm_37
 endif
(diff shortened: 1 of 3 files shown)


common.mk
```diff
@@ -139,15 +152,29 @@ else
 
 endif
 
+# compiler=clang Enables compilation with clang. 
+
+ifeq ($(compiler), clang)
```
galv: Nitpick: Imagine if someone wanted to use a specific version of clang, e.g. clang-4.0 or clang-3.9.

common.mk
```diff
@@ -139,15 +152,29 @@ else
...
+	BIN_SUFFIX := sm$(SM_ARCH)_clang_$(NVCC_VERSION)
```
galv: It makes sense to annotate which version of the CUDA runtime libraries you're using, though this code is a little misleading as it's using a variable called NVCC_VERSION (as opposed to something like CUDA_VERSION), and also because it could mislead someone into thinking that they're somehow using clang version 9.0 (or whatever CUDA version is being used), even though they're of course not.

You could consider doing something like `clang_$(CLANG_VERSION)_cuda_$(NVCC_VERSION)` instead.

common.mk
Artem-B: IMO the build is so simple here, that it's not worth bothering parametrizing every little knob. If it's needed, it's easy enough to add another config variable to specify specific clang path. At the moment the makefiles don't do that for nvcc or CUDA version, so doing that for clang is an overkill, IMO.

common.mk
Artem-B: I'll buy the argument that build should be annotated with CUDA version as that's really what we care about here. Specific version of nvcc or clang is largely irrelevant here. I'll update the patch to reflect CUDA version.

common.mk
Artem-B: As for performance, clang is somewhat slower than nvcc on most of the tests, ties or wins some.
Here's the run (on P100) of the tests compiled by nvcc and clang: https://pastebin.com/bfWQ9fbv

It's not exactly apples-to-apples comparison as it appears that the data is partitioned somewhat differently due to run-time choices that depend on the code of the kernels generated by compiler (i.e. number of registers used). The heuristics may need to be changed/tuned for clang.

Artem-B:
Ping.

Artem-B:
@dumerrill : Is this the right channel for submitting patches to cutlass?

kerrmudgeon:
Yes, this is the appropriate channel to discuss merge requests.

We thank you for your contribution to enable clang support and will merge
changes as soon as possible.

Artem-B:
Yes. It's done that way to make CUDA_LOG() work even when it's used in the host code.
Unlinke nvcc, clang always sees both host and device side of CUDA compilation. In original code, CUDA_LOG() expanded to print threadIdx. When clang ran into a function where CUDA_LOG was used in a host function, it complained, that one can't use threadIdx on host.

In order to make the macro code work in both host and device functions, I've split its implementation into printf with the message plus two overloaded functions that handle the difference in functionality on host/device. Now, when the macro is expanded in a host function, clang resolves the function to a __host__ one which does nothing. In device function, overload is resolved to the function which prints threadIdx, which is, again, what we want.

I could instead expand CUDA_LOG macro into a pair of variadic functions that incorporate printf, but clang does not support them yet on device side. I could rig something with variadic templates, but that would be an overkill, IMO.

jdemouth:
The issue with your solution is that you have no guarantee that the print with blockIdx/threadIdx will happen immediately before the 2nd call. You may end up with things like (shorten notation):

block=0 tid=0
block=0 tid=32
message X
message Y // good luck figuring which one maps to tid=0 and which one maps to tid=32

Instead, I'd have defined a group of 6 functions:

inline __host__ int get_threadIdx_x() { return 0; }
inline __device__ int get_threadIdx_x() { return threadIdx.x; }

And replaced threadIdx.x and friends with calls to those functions in the macro.

Does that make sense?

Artem-B:
@jdemouth : Good point. I've reworked the macro to print location and the message with one printf.

PR #6, opened 2018-03-09
Fix incorrect behavior of matrix::fill_random() and Hgemm

jeng1220:
1. Correct type conversion for matrix of FP16

Use `host_value_t` instead of `value_t` as conversion type in `matrix::fill_random()`. Otherwise, each element in `_h_data` will be same value and the value is unpredictable when `value_t` is `__half` type which causes verfication process of GEMM is malfunction.

2. Correct multiply and accumulation behavior of hgemm

Only data of the even rows of matrix B is read because only `thread_accumulator::mad_xy_even()` is called in `thread_accumulator::multiply_accumulate()` which causes operand of `HFMA2` is incorrect.

For example:
```
__half matrix_a[2][2] = {{3, 5},
                         {6, 2}};
__half matrix_b[2][2] = {{3, 6},
                         {7, 5}};
```
The result of hgemm_nn should be
```
__half matrix_c[2][2] = {{45, 27},
                         {51, 45}};
```
However, original cutlass' result is
```
__half matrix_c[2][2] = {{45, 27},
                         {45, 27}};
```
Therefore, also call `thread_accumulator::mad_xy_odd()` in `thread_accumulator::multiply_accumulate()` to read data of odd rows.

### DIFF
--- a/cutlass/gemm/thread_accumulator.h
+++ b/cutlass/gemm/thread_accumulator.h
@@ -453,12 +453,17 @@ public:
 
         // Simply traverse the accumulator tile in column-major order
         #pragma unroll
-        for (int x = 0; x < ThreadTilePairsX; ++x)
+        for (int x = 0; x < ThreadTilePairsX; x += 2)
         {
             #pragma unroll
             for (int y = 0; y < ThreadTilePairsY; ++y)
             {
                 mad_xy_even(pairs_tile_a, pairs_tile_b, x, y);
+            }
+            #pragma unroll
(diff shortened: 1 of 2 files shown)


kerrmudgeon:
We will need to close this MR as it applies to CUTLASS v0.1. We have substantially rewritten aspects of CUTLASS that likely render this merge request moot. The present HGEMMs and WMMA GEMMs have been extensively validated. 

We welcome comments on CUTLASS v1.0 as we expect it to be significantly more stable.

Thanks.

PR #47, opened 2019-05-02
Make CUTLASS compileable with Clang.

Artem-B:
Enable compilation with clang with these options:
```
$ cmake -DCUDA_COMPILER=clang -DCMAKE_CXX_COMPILER=/path/to/clang++ [other options]
```

Compilation for sm_75 requires a very recent clang build (r359248 or newer).
Some fp16 unit tests are currently failing, if cutlass is compiled with clang. I didn't find the root cause yet.
Performance will also need fair amount of tuning. Specifically clang requires huge increase to the default unroll thresholds in order to reach reasonable performance, but there's still substantial room for improvement.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,8 +24,9 @@ cmake_minimum_required(VERSION 3.3.0 FATAL_ERROR)
 
 set(CUTLASS_LANGUAGES CXX)
 
+if( CUDA_COMPILER STREQUAL "clang" )
 # CMake 3.9.0 has native support for CUDA without the need of the CUDA package. Use it!
-if(WIN32 AND NOT ${CMAKE_VERSION} VERSION_LESS "3.9.0")
+elseif(WIN32 AND NOT ${CMAKE_VERSION} VERSION_LESS "3.9.0")
   list(APPEND CUTLASS_LANGUAGES CUDA)
   set(CUTLASS_NATIVE_CUDA TRUE)
 
@@ -47,6 +48,32 @@ else()
 endif()
 
 project(CUTLASS ${CUTLASS_LANGUAGES})
+
+if( CUDA_COMPILER STREQUAL "clang" )
+  if( NOT CMAKE_CXX_COMPILER_ID STREQUAL "Clang" )
+    message(FATAL_ERROR "C++ compiler must be Clang. Currently it's ${CMAKE_CXX_COMPILER_ID}" )
+  endif()
+  string(APPEND CLANG_FLAGS " --std=c++11")
+  string(APPEND CLANG_FLAGS " --cuda-path=${CUDA_TOOLKIT_ROOT_DIR}")
+  string(APPEND CLANG_FLAGS " -mllvm -pragma-unroll-threshold=100000")
+  string(APPEND CLANG_FLAGS " -mllvm -unroll-threshold=5000")
+  string(APPEND CLANG_FLAGS " -Wno-unused-command-line-argument")
+  # needed for libcublasLt.so in case it's installed in the same location as libcudart.so
+  # dynamic linker can find it if linker sets RPATH (forced by --disable-new-tags)
+  # Otherwise linker uses RUNPATH and that does not propagate to loaded libs.
+  string(APPEND CLANG_FLAGS " -Wl,--disable-new-dtags")
+
+  link_libraries(${CUDA_CUDART_LIBRARY})
+  # Treat CUDA files as C++ files
+  macro(cutlass_add_executable)
+    foreach(File ${ARGN})
+      if(${File} MATCHES ".*\.cu$")
+        set_source_files_properties(${File} PROPERTIES LANGUAGE CXX)
+      endif()
+    endforeach()
+    add_executable(${ARGN})
+  endmacro()
+endif()
 
 # check if the configuration is supported
 if( NOT CMAKE_SIZEOF_VOID_P EQUAL 8 )
@@ -167,6 +194,7 @@ endif()
 
 # Set NVCC arguments
 foreach(ARCH ${CUTLASS_NVCC_ARCHS})
+  string(APPEND CLANG_FLAGS " --cuda-gpu-arch=sm_${ARCH}")
   if(CUTLASS_NVCC_EMBED_CUBIN)
     string(APPEND NVCC_FLAGS " -gencode arch=compute_${ARCH},code=sm_${ARCH}")
   endif()
@@ -175,12 +203,16 @@ foreach(ARCH ${CUTLASS_NVCC_ARCHS})
   endif()
 endforeach()
 
+if(CUTLASS_NVCC_EMBED_PTX)
+  string(APPEND CLANG_FLAGS " --cuda-include-ptx=all")
+endif()
+
 if (CUTLASS_ENABLE_TENSOR_CORE_MMA)
-  string(APPEND NVCC_FLAGS " -DCUTLASS_ENABLE_TENSOR_CORE_MMA=1")
+  string(APPEND COMMON_FLAGS " -DCUTLASS_ENABLE_TENSOR_CORE_MMA=1")
 endif()
 
 if (CUTLASS_ENABLE_CUBLAS)
-  string(APPEND NVCC_FLAGS " -DCUTLASS_ENABLE_CUBLAS=1")
+  string(APPEND COMMON_FLAGS " -DCUTLASS_ENABLE_CUBLAS=1")
 endif()
 
 if (CUTLASS_EXHAUSTIVE_PERFORMANCE_TEST)
@@ -189,6 +221,7 @@ endif()
 
 if (CUTLASS_NVCC_KEEP)
     string(APPEND NVCC_FLAGS " -keep")
+    string(APPEND CLANG_FLAGS " -save-temps=obj")
 endif()
 
 if (WIN32 AND CUTLASS_NATIVE_CUDA)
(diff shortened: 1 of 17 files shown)


Artem-B:
Ping. Let me know if you have any questions.

CMakeLists.txt
```diff
@@ -48,6 +49,32 @@ endif()
 
 project(CUTLASS ${CUTLASS_LANGUAGES})
 
+if( CUDA_COMPILER STREQUAL "clang" )
+  if( NOT CMAKE_CXX_COMPILER_ID STREQUAL "Clang" )
+    message(FATAL_ERROR "C++ compiler must be Clang. Currently it's ${CMAKE_CXX_COMPILER_ID}" )
+  endif()
+  string(APPEND CLANG_FLAGS " --std=c++11")
```
TimmyLiu: Hi. Thanks for this PR. It was a conscious decision that cutlass 1.3 will not include any features from C++11 and later. Do you think it is possible that we can avoid using C++11 with this PR?

cutlass/gemm/igemm_multiply_add.h
```diff
@@ -82,7 +82,9 @@ struct ThreadMultiplyAdd<ThreadGemmShape_, ThreadsPerWarp_, int8_t, int8_t, int>
     int const* a_int = reinterpret_cast<int const*>(&a[0]);
     int const* b_int = reinterpret_cast<int const*>(&b[0]);
 
+#pragma unroll
```
TimmyLiu: We have defined `CUTLASS_PRAGMA_UNROLL` ( https://github.com/NVIDIA/cutlass/blob/877bdcace6d765359833101f6eabd5036d607a33/cutlass/cutlass.h#L61 ). Do you think it is possible we can try to use that? Also a nit-pick the style of the code is to have `CUTLASS_PRAGMA_UNROLL` aligned with the for loop in the next line.

cutlass/reduction/threadblock_swizzle.h
```diff
@@ -35,7 +35,7 @@ struct DefaultBlockSwizzle {
   CUTLASS_HOST_DEVICE DefaultBlockSwizzle() {}
 
   /// Swizzle the block index.
-  CUTLASS_DEVICE dim3 swizzle() { return blockIdx; }
+  CUTLASS_DEVICE dim3 swizzle() { return {blockIdx.x, blockIdx.y, blockIdx.z}; }
```
TimmyLiu: I guess this change uses a c++11 feature. As state before we would like to avoid that if we can.

cutlass/wmma_matrix.h
```diff
@@ -30,20 +30,20 @@
 #if defined(__CUDACC__) && (!defined(__CUDA_ARCH__) || __CUDA_ARCH__ >= 700)
 #define CUTLASS_USE_WMMA_API
 
-#if defined(__CUDACC__) && (__CUDACC_VER_MAJOR__ >= 10) && (!defined(__CUDA_ARCH__) || __CUDA_ARCH__ >= 720)
+#if defined(__CUDACC__) && (CUDA_VERSION >= 10000) && (!defined(__CUDA_ARCH__) || __CUDA_ARCH__ >= 720)
```
TimmyLiu: I see a few places the CUDA_VERSION is bumped to 10000. Can you elaborate the reasoning behind this? We would like to be able to define `CUTLASS_USE_INT_WMMA` in this case if CUDA_VERSION is larger than 10, for example.

tools/util/half.h
```diff
@@ -209,7 +215,7 @@ std::string lexical_cast<std::string>(cutlass::half_t const& arg);
 
 #define HLF_MANT_DIG 10
 
-namespace std {
+namespace cutlass {
```
TimmyLiu: Can you elaborate the reasoning behind this change? It was a deliberate choice to place the code here under std

tools/util/half.h
```diff
@@ -229,7 +235,10 @@ int fpclassify(cutlass::half_t const&);  /// returns a flag classifying floating
 bool signbit(cutlass::half_t const&);  /// returns true if negative, false if positive
 
 cutlass::half_t sqrt(cutlass::half_t const&);  /// square root of half_t
+cutlass::half_t copysign(cutlass::half_t const&, cutlass::half_t const&);
```
TimmyLiu: I guess this is another c++11 feature?

CMakeLists.txt
Artem-B: What is the minimum version CUTLASS is intended to support?

--std=c++11 is needed by clang in order to make work with the standard C++ library. A lot of it depends in standard library functions being constexpr. 
I did not intend to add anything that specifically requires c++11 features to cutlass, but clang itself does need c++11 it in order to successfully compile nontrivial CUDA code that depends on the standard library. I've just tried to build cutlass with -std=c++03 and it failed miserably with all sorts of errors in libstdc++.

CUTLASS itself does not set any specific C++ standard version during compilation, so, according to the CUDA docs NVCC will be compiling the sources with whatever is the default C++ standard picked by the host compiler.  In my case (gcc 7.3) it's c++14. So clang is effectively compiling with an older standard than nvcc.

cutlass/reduction/threadblock_swizzle.h
Artem-B: blockIdx is special. Clang only allows accessing it as individual fields. 
I can rewrite it as a bit more verbose, but equivalent
```
    dim3 r;
    r.x = blockIdx.x;
    r.y = blockIdx.y;
    r.z = blockIdx.z;
    return r;
```
https://godbolt.org/z/zap7ln

cutlass/wmma_matrix.h
Artem-B: I did *not* bump the version -- it was 10 in the original code.

I've replaced `__CUDACC_VER_MAJOR__` which is not set by clang to `CUDA_VERSION` which matches the intent of this code and is available to both nvcc and clang.

tools/util/half.h
Artem-B: No, it's a forward declaration for the function already implemented at the end of this header. Unlike the sqrt() and other functions above, this one wasn't declared.

I've added the declaration for consistency. AFAICT, it's not strictly necessary and I can remove it, if you prefer.

tools/util/half.h
Artem-B: Good question. It's a leftover from an early attempt to make cutlass compile with clang.
It appears to be unnecessary now. I'll remove it.

cutlass/reduction/threadblock_swizzle.h
TimmyLiu: or could you do `return dim3(blockIdx.x, blockIdx.y, blockIdx.z);`?

cutlass/wmma_matrix.h
TimmyLiu: ah I see the difference now. Thanks for clarification.

tools/util/half.h
TimmyLiu: When I compile the code in this MR with GCC without C++11. It did complain `error: namespace "std" has no member "copysign"`. If you think copysign is not needed or not necessary could we remove it?

Artem-B: Scratch my previous comment. I did add both the declaration and the implementation. I'll need to check tomorrow, but I believe it was needed to make the standard library happy.  I'll check if it's still needed. If it is, I can preprocess it away for C++ standards older than c++11. Will that be acceptable?

cutlass/reduction/threadblock_swizzle.h
Artem-B: Indeed. It's been a long day. Will do.

cutlass/gemm/igemm_multiply_add.h
Artem-B: Will do.

tools/util/half.h
TimmyLiu: hehe. Yes I think that's reasonable.  
It looks like the code as is makes call to std::copysign
```
inline cutlass::half_t copysign(cutlass::half_t const& a,
                                cutlass::half_t const& b) {
  return cutlass::half_t(std::copysign(float(a), float(b)));
}
```
which is a C++11 feature. But I cannot seem to find the code that makes call to it. Anyways, we would like to make sure cutlass 1.3 does not use c++11 feature. I think it is okay for clang compiler to use `--std=c++11` But we want to make sure cutlass can still compile with older compiler that does not support cpp11.

kerrmudgeon: Hi, Artem!

Specifically, this is a library feature added in C++11. I don't think this depends on any actual C++11 language features, so it's okay to use.

I'd recommend contributing this function and others like it to `cutlass/util/platform.h` so we can swap it out at a point in the future when we can fully embrace C++11.

tools/util/half.h
Artem-B: It turns out the issue is a bit more complicated. Could you elaborate on the reasons for using `namespace std` here?

Technically, IIUIC, you are not allowed to extend std:: namespace this way -- it's undefined behavior that happens to work with libstdc++. It fails with libc++, which is what I was running into.
Consider this reduced case: https://godbolt.org/z/29jOBD -- when `namespace std` is used, it compiles with libstdc++ with both nvcc and clang. However, it fails with clang+libc++.

Using `namespace cutlass` here allows the standard library to find the functions via ADL and works with both libc++ and libstdc++.  

@timshen91 -- Tim, what's the official stand on extending standard c++ library this way?

tools/util/half.h
Artem-B: On second thought, quite a few functions above are only in c++11 (e.g. [signbit](https://en.cppreference.com/w/cpp/numeric/math/signbit), [isnan](https://en.cppreference.com/w/cpp/numeric/math/isnan), [isfinite](https://en.cppreference.com/w/cpp/numeric/math/isfinite))

I'm not sure it's worth making a special case just for `copysign.

cutlass/wmma_matrix.h
```diff
@@ -30,20 +30,20 @@
...
 #endif
 
 #include "stdio.h"
 
-#if __CUDACC_VER_MAJOR__ >= 10
+#if CUDA_VERSION >= 10000
 #include <mma.h>
 #else
-#include <crt/mma.h>
+#include <mma.h>
```
Artem-B: I'm not quite sure why 'crt/mma.h' was included originally. The header in CUDA-10.1 has explicit checks to prevent direct inclusion by user.

Should the whole thing be replaced with unconditional `#include <mma.h>` ?

kerrmudgeon: We had the same thought, upon review of this. Yes, I agree that `<mma.h>` is the appropriate include.

cutlass/gemm/igemm_multiply_add.h
Artem-B: Done.

cutlass/reduction/threadblock_swizzle.h
Artem-B: Fixed.

tools/util/half.h
timshen91: According to the C++ standard [namespace.std]:
* Unless otherwise specified, the behavior of a C++ program is undefined if it adds declarations or definitions to namespace std or to a namespace within namespace std.
* Unless explicitly prohibited, a program may add a template specialization for any standard library class template to namespace std provided that (a) the added declaration depends on at least one program-defined type and (b) the specialization meets the standard library requirements for the original template.

That basically says that defining std::isnan/isinf/copysign/etc for half_t causes undefined behavior, but specializing numeric_limits for half_t is fine.

For the practical side, the standard has many reasons to forbid such arbitrary cutomization. For example, libc++ implements everything in `namespace std { inline namespace __1 { ... } }`. In that case, half_t version of isnan doesn't even overload with std::__1::isnan.

The common way of injecting user customization into standard library functions is through ADL (https://en.cppreference.com/w/cpp/language/adl), which requires the customization (isnan) to be in the same namespace as the type (half_t), hence this change.

kerrmudgeon: Thank you for explaining this, Tim and Artem. I'm definitely convinced.  Your proposed solution of defining these functions in namespace `cutlass` sounds like the correct approach.

cutlass/wmma_matrix.h
Artem-B: Fixed.

tools/util/half.h
Artem-B: @TimmyLiu  makes a valid point that I'm also using `std::copysign` which may not be available before c++11. That indeed breaks if nvcc compiles the files using -std=c++03. I've added an  `#if __cplusplus` guard around it.

TimmyLiu: Thank you!

CMakeLists.txt
Artem-B: I think we can call this one resolved. The patch does not add any new dependencies on c++11 and the flag is needed only because standard library support for CUDA compilation in clang is very limited for standard library before c++11.

PR #53, opened 2019-07-10
Performance enhancement for Volta Tensor Cores TN layout

kerrmudgeon:
Performance enhancement for Volta Tensor Cores TN layout. A pointer to shared memory was previously accessed indirectly via an expression based on threadIdx. This patch factors this expression into the iterator's constructor enabling the the pointer array to be stored in registers, thereby eliminating all indirection and register spills.


d-k-b:
Looks good to me!

PR #82, opened 2020-03-27
update tools/library/CMakeLists to require python 3.6 according to #70

thakkarV:
#70 only updates the documentation. This commit reflects this bump in python version to the CMake configuration as well.


thakkarV:
hi @kerrmudgeon , would you mind reviewing this patch?

PR #87, opened 2020-04-19
Fixed FLOPS count for CGEMM

cliffburdick:
Complex matrix multiplies require 8mnk - 2mn operations, which is slightly different from the amount in this file. The 8mnk is from the element-wise multiplies, and the 2mn is from summing the products. The formula can be derived by:

6mnk+ 2mn(k-1) = 8mnk - 2mn

### DIFF
--- a/examples/10_planar_complex/planar_complex.cu
+++ b/examples/10_planar_complex/planar_complex.cu
@@ -186,12 +186,10 @@ struct Options {
 
   /// Compute performance in GFLOP/s
   double gflops(double runtime_s) const {
+    // Number of real-valued multiply-adds. Each complex multiply has 4 real multiplies and 2 real adds.
+    int64_t ops = (8 * (problem_size.m() * problem_size.n() * problem_size.k()) - 2.0 * (problem_size.m() * problem_size.n())) * batch_count;
 
-    // Number of real-valued multiply-adds 
-    int64_t fmas = problem_size.product() * batch_count * 4;
-    
-    // Two flops per multiply-add
-    return 2.0 * double(fmas) / double(1.0e9) / runtime_s;
+    return double(ops) / double(1.0e9) / runtime_s;
   }
 };
 


kerrmudgeon:
Thank you for submitting this, Cliff.

I'm not completely sure I understand the calculation, however. 

The mainloop computes complex multiply-add as follows, with four fused multiply-add instructions and eight flops total (1 fma = 2 flops).
```c++
c_r += a_r * b_r;
c_i += a_i * b_r;
c_r -= a_i * b_i;
c_i += a_r * b_i;
```
This approach differs from an alternative implementation which first computes the complex product then adds to the result. This reduces the total liveness by avoiding extra accumulator state held in registers.

We're not counting multiply-adds to perform the alpha/beta scaling after the matrix product, but these would contribute an additional `mn` operations.

Given this, does it seem reasonable to compute total flops as `8mnk`?

cliffburdick:
> Thank you for submitting this, Cliff.
> 
> I'm not completely sure I understand the calculation, however.
> 
> The mainloop computes complex multiply-add as follows, with four fused multiply-add instructions and eight flops total (1 fma = 2 flops).
> 
> ```c++
> c_r += a_r * b_r;
> c_i += a_i * b_r;
> c_r -= a_i * b_i;
> c_i += a_r * b_i;
> ```
> 
> This approach differs from an alternative implementation which first computes the complex product then adds to the result. This reduces the total liveness by avoiding extra accumulator state held in registers.
> 
> We're not counting multiply-adds to perform the alpha/beta scaling after the matrix product, but these would contribute an additional `mn` operations.
> 
> Given this, does it seem reasonable to compute total flops as `8mnk`?

Hi @kerrmudgeon, I was leaving out the FMA part of it since I thought the typical way Nvidia counts FLOPS is just by number of operations, since even though FMA is a single instruction, it still counts as two operations. 

My reasoning was as follows:
1) These operations would be performed 8mnk times:
`c_r = a_r * b_r - a_i * b_i`
`c_i = a_i * b_r + a_r * b_i`
The total above is equal to 8 real operations mnk times over the entire GEMM.
2) After each of these products, all elements in the vector must be summed to create the single complex term in the output of C. This is:
`2mn * (k-1)`
Since we are summing both the real and imaginary components of each element in the vector. We do it `k-1` times to sum the entire vector.

I'm also not counting the MADDs from alpha/beta, but that would take mine down to `8mnk - mn`.

kerrmudgeon:
Thank you, Cliff. I think I'm following.

If I understand correctly, each point in _mnk_ is a complex multiplication which could be computed via 4 scalar multiplies and 2 adds (6 flops total). For k index 0, this is ‘moved’ into the accumulator registers and not counted. For k > 0, it’s added to the accumulators with 2 more flops.

Thus, 6mnk + 2*mn(k-1) flops total in an ideal scalar implementation.

That said, the planar complex kernels in CUTLASS 2.1 do not explicitly compute the complex products with six scalar flops in this manner. Rather, these kernels are implemented using Tensor Core operations. These are fused matrix multiply-add operations with _k=4_ on Volta and _k=8_ on Turing.  

To use Tensor Cores efficiently without extra temporary storage, we adopted the scheme in which the kernel accumulates the products in place using four fused multiply-add operations. It incurs `2mn` more adds over all (adding to a zero matrix for the first iteration), but these are effectively free in hardware and would be difficult to skip.

This does affect associativity. From a numerical perspective, do you have any concerns about the four multiply-adds accumulating the result without explicitly forming a complex product?

cliffburdick:
> Thank you, Cliff. I think I'm following.
> 
> If I understand correctly, each point in _mnk_ is a complex multiplication which could be computed via 4 scalar multiplies and 2 adds (6 flops total). For k index 0, this is ‘moved’ into the accumulator registers and not counted. For k > 0, it’s added to the accumulators with 2 more flops.
> 
> Thus, 6mnk + 2*mn(k-1) flops total in an ideal scalar implementation.
> 
> That said, the planar complex kernels in CUTLASS 2.1 do not explicitly compute the complex products with six scalar flops in this manner. Rather, these kernels are implemented using Tensor Core operations. These are fused matrix multiply-add operations with _k=4_ on Volta and _k=8_ on Turing.
> 
> To use Tensor Cores efficiently without extra temporary storage, we adopted the scheme in which the kernel accumulates the products in place using four fused multiply-add operations. It incurs `2mn` more adds over all (adding to a zero matrix for the first iteration), but these are effectively free in hardware and would be difficult to skip.
> 
> This does affect associativity. From a numerical perspective, do you have any concerns about the four multiply-adds accumulating the result without explicitly forming a complex product?

Hi @kerrmudgeon, thanks for clarifying! I think the makes sense for the way you're computing this, and the difference is so small that it's not really an issue. I have no worries about the associativity and loss to due that at this point. 

I will close this given that information.

PR #91, opened 2020-05-31
Honor `CMAKE_CXX_COMPILER` as CUDA host compiler.

trivialfis:
If you run `CC=gcc-8 CXX=g++-8 cmake ../`, as of cmake 3.13.4, it will ignore these environment variables for cuda host compiler.  Setting the `CMAKE_CUDA_HOST_COMPILER` builtin variable fixes the issue.

### DIFF
--- a/CUDA.cmake
+++ b/CUDA.cmake
@@ -37,6 +37,7 @@ endif()
 
 if(CUTLASS_NATIVE_CUDA)
 
+  set(CMAKE_CUDA_HOST_COMPILER ${CMAKE_CXX_COMPILER})
   enable_language(CUDA)
 
   if(NOT CUDA_VERSION)


CUDA.cmake
```diff
@@ -37,6 +37,7 @@ endif()
 
 if(CUTLASS_NATIVE_CUDA)
 
+  set(CMAKE_CUDA_HOST_COMPILER ${CMAKE_CXX_COMPILER})
```
d-k-b, line 40: I've debated adding this for a while now. It seems reasonable enough. Robert Maynard spoke a bit about why this isn't the case [here](https://gitlab.kitware.com/cmake/cmake/-/issues/17471), but since we require CXX and we know it will be active before this point in the CMake configuration, I think it should be safe to do this.

However, please test that setting CUDAHOSTCXX environment variable and CMAKE_CUDA_HOST_COMPILER from command line will override this setting. If not, we'll need to add guards to allow those overrides.

trivialfis:
@d-k-b Thanks for replying.  After looking through the discussion on CMake.

> When finally setting the CUDA host compiler we don't want to presume any relationship across C++ and CUDA.

This statement helps.  I will close this PR and let users decide what to do.

PR #151, opened 2020-11-23
Added build option to show warnings

igormp:
This would allow to easily look for some code smells to fix later (and GCC did throw a few warnings after enabling it).

Not sure if this is the best place to enable such warnings, or if that's a suitable flag name. I'm open to suggestions :)


hwu36:
Thank you very much for your pull request.

It seems we accidentally dropped  `-Wall -Werror` sometime in the past and we want to add it back for all the compilers we support in Linux, Windows, Clang without behind any flags.

So, I think we'd love to have pull request(s) fixing the warnings first and a final pull request to add `-Wall -Werror` to CMake.

igormp:
That seems like a great idea and guarantees a warning-free build from the get-go. I'll try to clean up some warnings later on then.

igormp:
Seeing how I haven't been able to keep up with this PR, I'll close it.

PR #179, opened 2021-02-16
Allow mma instructions with clang.

dxyzab:
Enable the inline PTX when using clang.

### DIFF
--- a/include/cutlass/arch/mma_sm70.h
+++ b/include/cutlass/arch/mma_sm70.h
@@ -37,13 +37,13 @@
 #include "cutlass/layout/matrix.h"
 #include "cutlass/numeric_types.h"
 
-#if (defined(__clang__) || ((__CUDACC_VER_MAJOR__ > 10) || (__CUDACC_VER_MAJOR__ == 10 && __CUDACC_VER_MINOR__ >= 1)))
+#if ((defined(__clang__) && defined(__CUDA__) && CUDA_VERSION >= 10011) || ((__CUDACC_VER_MAJOR__ > 10) || (__CUDACC_VER_MAJOR__ == 10 && __CUDACC_VER_MINOR__ >= 1)))
 #define CUTLASS_ARCH_MMA_SM70_SUPPORTED
 #endif
 
 #if (defined(__CUDA_ARCH__) && (__CUDA_ARCH__ >= 700))
 
-#if (defined(__clang__) || ((__CUDACC_VER_MAJOR__ > 10) || (__CUDACC_VER_MAJOR__ == 10 &&__CUDACC_VER_MINOR__ >= 1)))
+#if ((defined(__clang__) && defined(__CUDA__) && CUDA_VERSION >= 10011) || ((__CUDACC_VER_MAJOR__ > 10) || (__CUDACC_VER_MAJOR__ == 10 &&__CUDACC_VER_MINOR__ >= 1)))
 #define CUTLASS_ARCH_MMA_SM70_ENABLED
 #endif
 
--- a/include/cutlass/arch/mma_sm70.h
+++ b/include/cutlass/arch/mma_sm70.h
@@ -37,13 +37,13 @@
 #include "cutlass/layout/matrix.h"
 #include "cutlass/numeric_types.h"
 
-#if ((__CUDACC_VER_MAJOR__ > 10) || (__CUDACC_VER_MAJOR__ == 10 && __CUDACC_VER_MINOR__ >= 1))
+#if (defined(__clang__) || ((__CUDACC_VER_MAJOR__ > 10) || (__CUDACC_VER_MAJOR__ == 10 && __CUDACC_VER_MINOR__ >= 1)))
 #define CUTLASS_ARCH_MMA_SM70_SUPPORTED
 #endif
 
 #if (defined(__CUDA_ARCH__) && (__CUDA_ARCH__ >= 700))
 
-#if ((__CUDACC_VER_MAJOR__ > 10) || (__CUDACC_VER_MAJOR__ == 10 &&__CUDACC_VER_MINOR__ >= 1))
+#if (defined(__clang__) || ((__CUDACC_VER_MAJOR__ > 10) || (__CUDACC_VER_MAJOR__ == 10 &&__CUDACC_VER_MINOR__ >= 1)))
 #define CUTLASS_ARCH_MMA_SM70_ENABLED
 #endif
 
--- a/include/cutlass/arch/mma_sm75.h
+++ b/include/cutlass/arch/mma_sm75.h
@@ -49,7 +49,7 @@
 
 ////////////////////////////////////////////////////////////////////////////////
 
-#if (defined(__clang__) || ((__CUDACC_VER_MAJOR__ > 10) || (__CUDACC_VER_MAJOR__ == 10 && __CUDACC_VER_MINOR__ >= 2)))
+#if ((defined(__clang__) && defined(__CUDA__) && CUDA_VERSION >= 10011) || ((__CUDACC_VER_MAJOR__ > 10) || (__CUDACC_VER_MAJOR__ == 10 && __CUDACC_VER_MINOR__ >= 2)))
 
 #define CUTLASS_ARCH_MMA_SM75_SUPPORTED 1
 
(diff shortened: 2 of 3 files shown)


hwu36:
This is exciting.  Have you tested this?  Would you please let us know the version of clang and cuda you tested?

dxyzab:
Yes I have tested it on Volta using a variant of batched gemm.

Debian clang version 11.0.1-2
Target: x86_64-pc-linux-gnu
Thread model: posix

cuda 10.1

hwu36:
Thanks.  What is your clang version?  Also would you please point me to the test that you tested?

dxyzab:
That is an internal test. I can one of the basic gemm tests in the this repo, which one do you have in mind?

hwu36:
If you have things setup, would you please try to run cutlass unit tests following this https://github.com/NVIDIA/cutlass/blob/master/media/docs/quickstart.md ?

Basically `make test_unit -j` after the correct cmake.  e.g. `cmake .. -DCUTLASS_NVCC_ARCHS=70` for volta only.

dxyzab:
Unfortunately I don't have a cmake based set up.
I can run
https://github.com/NVIDIA/cutlass/blob/master/test/unit/gemm/device/gemm_f16n_f16n_f16t_volta_tensor_op_f32_sm70.cu
and put the results here if that is helpful.

hwu36:
That would be great.  Or you can run https://github.com/NVIDIA/cutlass/tree/master/examples/07_volta_tensorop_gemm if it is easier for you.

dxyzab:
https://github.com/NVIDIA/cutlass/tree/master/examples/07_volta_tensorop_gemm prints 'Passed' which is good.

I actually ran this kernel on a 2080. The PTX shows the mma.sync instructions:

image

Some kernel stats

image-3

hwu36:
Thank you.  We will do some testing and get back to you.

hwu36:
I don't have much luck.  The only combination of cuda and clang works for me is cuda10.0 + clang8.0.  volta mma requires 10.1 though.

Would you please share your command line when building the volta command line?

dxyzab:
What is your error with clang version 11.0.1-2?

hwu36:
Clang 11 has this requirement

`./include/clang/Basic/Cuda.h:35:  LATEST_SUPPORTED = CUDA_101,`


In cuda10.1, in `$CUDA/include/crt/host_config.h`
```
#if defined(__clang__) && !defined(__ibmxl_vrm__) && !defined(__ICC) && !defined(__HORIZON__) && !defined(__APPLE__)

#if (__clang_major__ >= 9) || (__clang_major__ < 3) || ((__clang_major__ == 3) &&  (__clang_minor__ < 3))
#error -- unsupported clang version! clang version must be less than 9 and greater than 3.2
#endif  /* (__clang_major__ >=  9) || (__clang_major__ < 3) || ((__clang_major__ == 3) &&  (__clang_minor__ < 3)) */
```

Then I ran clang8 with cuda10.1, I got 

`clang-8: error: cannot find libdevice for sm_70. Provide path to different CUDA installation via --cuda-path, or pass -nocudalib to build without linking with libdevice.`

clang8.0+cuda10.0 is fine.

dxyzab:
Could you please try with the "#error ..." in host_config.h removed/commented out, to see if it works with clang 11.

dxyzab:
Or try with clang 11 and cuda 11 which has

#if (__clang_major__ >= 12) || ...

Artem-B:
> clang-8: error: cannot find libdevice for sm_70. Provide path to different CUDA installation via --cuda-path, or pass -nocudalib to build without linking with libdevice.

Your problem is that you're using clang-8, which is very old. You need a more recent version in order to use recent CUDA.

hwu36:
Hi @Artem-B ,

We need your help on this. You are the expert.  We have difficulty in finding compatible clang and cuda.

ToT clang still checks `DetectedVersionIsNotSupported = Version > CudaVersion::LATEST_SUPPORTED` (https://github.com/llvm/llvm-project/blob/cd8cc00b9e2b2b2b10270f7485eb47759bc54cc1/clang/lib/Driver/ToolChains/Cuda.cpp#L232) where `LATEST_SUPPORTED = CUDA_101`.  Otherwise, I will get a warning.

However, CUDA10.1 requires the latest clang to be clang8.  Otherwise I will get an error.

From what I tested, clang8 can only work with CUDA10.0 which is not enough for volta mma.  Then, I am stuck.

So, what version of cuda and clang should I use to support volta mma without any warning and error?

Artem-B:
There are different meanings of 'support'.

10.1 is the most recent CUDA version for which clang provides fairly complete set of compiler builtins needed by CUDA headers ( and users who may have relied on them).

More recent CUDA versions have introduced new instructions. Clang does not provide them. At the same time, it's capable of compiling CUDA code which could be compiled with CUDA-10.1. Clang has limited ability to pass through PTX version and the new GPU architecture info throughout the pipeline, so you can still target new GPUs, as long as you don't touch new CUDA features that are not available in CUDA-10.1

The `don't touch new features`, unfortunately, applies to the CUDA headers themselves, too. So if you're compiling something for sm_86 that sometimes ends up including CUDA headers that rely on the features clang does not provide. E.g. bf16 or tf32 types support, or new TensorCore instruction variants. 

Hecnde, the warning, when clang allows using the new CUDA version, but does not promice that it will always work. If you do know that you code does work, you can always disable the warning. If compilation succeeds, clang does produce valid GPU binaries.

The bottom line -- ignore the warning while you're experimenting. The warning is there to indicate that you're entering the somewhat uncharted territory where clang may not be able to compile some CUDA code. That said, in practice it forks fine in most of the cases as there's not that much code that really needs the new CUDA features.

hwu36:
Thank you very much, @Artem-B .  If it is okay to Google that released CUTLASS  would generate warnings when building with Clang, we can go ahead to at least support volta mma.

Cutlass uses most new CUDA features in inline ptx, I guess Clang just treat inline box as black box.

dxyzab:
Is there anything further you need from our end? It would be good to merge the changes.

include/cutlass/arch/mma_sm70.h
```diff
@@ -37,13 +37,13 @@
 #include "cutlass/layout/matrix.h"
 #include "cutlass/numeric_types.h"
 
-#if ((__CUDACC_VER_MAJOR__ > 10) || (__CUDACC_VER_MAJOR__ == 10 && __CUDACC_VER_MINOR__ >= 1))
+#if (defined(__clang__) || ((__CUDACC_VER_MAJOR__ > 10) || (__CUDACC_VER_MAJOR__ == 10 && __CUDACC_VER_MINOR__ >= 1)))
```
Artem-B, line 40: Clang's clause should probably be a bit more detailed:
 `(defined(__clang__) && defined(__CUDA__) && CUDA_VERSION >= 10011)`

- `__CUDA__` will only be defined if clang is compiling CUDA sources. It will not be set when clang is used by NVCC during host-side compilation. It may not be necessary if mma_sm70.h is guaranteed to never be included from a C++ compilation.

- `CUDA_VERSION` is needed to make sure we enable new MMA instructions when clang uses sufficiently recent CUDA version. While nvcc version definitions and CUDA version are in sync, CUDACC_VER_* can be used to infer the CUDA version.. clang must use the `CUDA_VERSION` to do the same.

dxyzab: Updated below.

Artem-B:
clang-related conditionals look correct now, but it would be great to confirm that they do indeed work.

@hwu36 
> Cutlass uses most new CUDA features in inline ptx, I guess Clang just treat inline ptx as black box.

Correct. Inline assembly should work.

dxyzab:
The new clang-related conditionals work for me.

hwu36:
Sorry for the late response, we were busy with the 2.5 release.  Now it is out, would you please try clang again on Volta and/or Turing if possible?  We add clang support to `ldmatrix` (https://github.com/NVIDIA/cutlass/blob/master/include/cutlass/arch/memory_sm75.h), but haven't tested with `mma` yet.

hwu36:
In CUTLASS 2.6, we enhanced Clang support. Now, the combination of Clang 13 and CUDA 11.4 can build and run kernels from Pascal and Ampere.

PR #193, opened 2021-03-04
HFMA2 Convolutions for SM60 onwards 

Peter9606:
This patch made fp16 related conv testable for arch SM50.

### DIFF
--- a/test/unit/conv/device/conv2d_fprop_implicit_gemm_f16nhwc_f16nhwc_f16nhwc_simt_f16_sm60.cu
+++ b/test/unit/conv/device/conv2d_fprop_implicit_gemm_f16nhwc_f16nhwc_f16nhwc_simt_f16_sm60.cu
@@ -38,7 +38,7 @@
 
 ////////////////////////////////////////////////////////////////////////////////
 TEST(SM60_Device_Conv2d_Fprop_Analytic_ImplicitGemm_f16nhwc_f16nhwc_f16nhwc_simt_f16,
-  32x64_8x2_32x32x8) {
+  128x128_8x2_64x64x8) {
 
   /// Conv operation element types for the Gemm equivalent (ImplicitGemm)
   using ElementA           = cutlass::half_t;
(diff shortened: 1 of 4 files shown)


manishucsd:
"This patch made fp16 related conv testable" .  Can I also suggest adding the related conv unit test with this MR? Does it pass?

Peter9606:
@manishucsd  Test cases appended, and all passed on my local environment (V100 + CUDA 10.0).

hwu36:
Great! hfma conv just work out of the box.  I am curious if we can use 128x128 threadblock size and 64x64 warp size in the test.

Peter9606:
Just so you know, it's NOT running on tensor core but cuda core, these two cases spent quite long time.

manishucsd:
I am glad we are getting the first convolution unit test coming in from the community :). I see one more change and then we can merge it.

The change in `thread/mma_sm60.h` suggests that the minimum SM we should try it on is SM60. HFMA2, for all practical purposes, became usable starting SM60. Thus, I recommend changing the following:
- Rename test file to `conv2d_fprop_implicit_gemm_f16nhwc_f16nhwc_f16nhwc_simt_f16_sm60.cu`
- Rename each individual test from `SM50_*` to `SM60_*`
- Use ArchTag sm60 in the unit tests [here](https://github.com/NVIDIA/cutlass/pull/193/files#diff-bf9dadde773925c771787fa572d4e2f83b417cd47d61cf512baa6b76c38e5120R40).

hwu36:
lgtm.   We will get back to you after we run some tests.  Thanks.

Peter9606:
@manishucsd  PR is updated based on your suggestions.
@hwu36 I just tried 128x128 + 64x64 configuration, it works well. I didn't put this configuration into this PR only because of too much time consuming.

test/unit/conv/device/conv2d_fprop_implicit_gemm_f16nhwc_f16nhwc_f16nhwc_simt_f16_sm60.cu
```diff
@@ -0,0 +1,128 @@
...
+  using Conv2dFprop = cutlass::conv::device::ImplicitGemmConvolution<Conv2dFpropKernel>;
+
+  /// Run all unit test sizes with device-level Conv2d instance
+  EXPECT_TRUE(test::conv::device::TestAllConv2d<Conv2dFprop>());
+
+}
+
+////////////////////////////////////////////////////////////////////////////////
+
+TEST(SM60_Device_Conv2d_Fprop_Analytic_ImplicitGemm_f16nhwc_f16nhwc_f16nhwc_simt_f16,
```
manishucsd, line 86: This is an SM60_Device_Conv2d_Fprop_**Optimized**_ImplicitGemm_f16nhwc_f16nhwc_f16nhwc_simt_f16

I am pushing a change to your public branch to make this change. 

Additionally, I am running some full system tests on this PR internally. Once those finish, I will merge this PR.

manishucsd:
Hi Peter, 

I pushed changes to your PR's public branch:
- to fix the test name
-  only instance large 128x128 tile size in unit tests
Large tile size should speed up the unit test a bit. 

If you need other tile sizes, for this operator, the right place to instance them would be in the [generator](https://github.com/NVIDIA/cutlass/blob/master/tools/library/scripts/generator.py#L345) and try out those via cutlass_profiler.

manishucsd (approved):

PR #214, opened 2021-03-23
Bugfix: memsetAsync uses wrong default stream

Peter9606:
This error will cause workspace initialize operation run on the default stream which might differ from the stream that gemm kernel runs on. It happens when explicitly pass a non-default stream to CUTLASS gemm kernel.


hwu36:
lgtm.  Let me do some tests internally and then back to you.  It may take a while because I am behind in many tasks now.

hwu36 (approved):

PR #218, opened 2021-03-24
The kAdvance should be always 1 for column-A and Row-B

hanjiemicro:
Whether A/B is col-major or row-major, the expected layout in shared mem is A:col-major, B: row-major. Both te advanceRank should be 1.
Signed-off-by: Jie Han <15202116112@163.com>


hwu36:
It is true that in the current implementation when we pass `kAdvanceRank` to the base class, it is always 1 for both A and B operands.  However, I don't think it is a good idea to hard code it to 1 because we would like to keep the flexibility in case we want to support different layouts in the future.

mnicely:
Closing per @hwu36 comments
> However, I don't think it is a good idea to hard code it to 1 because we would like to keep the flexibility in case we want to support different layouts in the future.

Issue #211, opened 2021-03-18
Performance Gap on T4


hzfan:
I found cutlass matmul is around 30% slower than cublas on T4 with cuda 11.2. Is there anything I can do to improve its performance?

I used the following command for profiling:

```
./build/tools/profiler/cutlass_profiler \
--operation=Gemm --m=4096 --n=32 --k=25088 --beta=0 \
--A=f32:t --B=f32:n --accumulator-type=f32 --verification-enabled=false \
--split-k-slices=4
```

The best result is
```
  Problem ID: 1

        Provider: CUTLASS
   OperationKind: gemm
       Operation: cutlass_simt_sgemm_32x128_8x2_tn_align1

          Status: Success
    Verification: OFF
     Disposition: Not verified


       Arguments: --gemm_kind=universal --m=4096 --n=32 --k=25088 --A=f32:row --B=f32:column --C=f32:column --alpha=1  \
                  --beta=0 --split_k_slices=4 --batch_count=1 --op_class=simt --accum=f32 --cta_m=32 --cta_n=128 --cta_k=8  \
                  --stages=2 --warps_m=1 --warps_n=2 --warps_k=1 --inst_m=1 --inst_n=1 --inst_k=1 --min_cc=50 --max_cc=1024  \
                 

           Bytes: 414777344  bytes
           FLOPs: 6576930816  flops

         Runtime: 3.95568  ms
          Memory: 97.655 GiB/s

            Math: 1662.66 GFLOP/s
```

For cublas, I benchmarked with my own script (with around 0.1ms extra overhead):
```
time: 3.07 ms
kernel: volta_sgemm_128x32_sliced1x4_tn
```

hwu36:
Your problem size has big M, tiny N, giant K.  It needs to split k dimensions to run efficiently and there are two ways to do it.  

- One is called split-k, it launches multiple threadblocks in the k dimension.  This is what you used with cutlass profiler
- The other one is called sliced-k, it still launches 1 threadblock.  But inside each threadblock, more than 1 warp is used in the k dimension.  This is what cublas uses.  Currently, sgemm does not support sliced-k in cutlass. Igemm, all tensor core kernels support them.  (It is not hard to let sgemm to support it though.  If you need it urgently, I can show you how to do it.)

To reduce the gap
- You can try to change the `--split-k-slices` value.
[...]

PR #219, opened 2021-04-01
Fix for issue #211

mani-ananth:
- Add a sliced-K tile size kernel instance to the library for Turing 1688 TensorOp
- Fix num warps calculations in implicit gemm header to account for slice-K


hwu36:
Attn: @hzfan

hwu36 (approved):

Issue #225, opened 2021-04-05
15_ampere_sparse_tensorop_gemm fails


decioren:
running 15_ampere_sparse_tensorop_gemm fails in the correctness test in the end. It seems that all values are zeros.
changing to half data type according to https://github.com/NVIDIA/cutlass/issues/103 also fails with zeros.
I tried a few fixes to the random data generation and got the host reference implementation to work, but device code still gives zeros.

a windows 10 machine with cuda 11.2. Any ideas?

hwu36:
ACK.

Would you please show me your fix first?

hwu36:
Never mind.  This example has several problems.  I will let you know when I fix it and I will show you the diff.

For now, you can take a look at our sparse gemm unit tests in https://github.com/NVIDIA/cutlass/tree/master/test/unit/gemm/device .  Just search "sparse" there.

hwu36:
You can use `make cutlass_test_unit_gemm_device_sparse_tensorop_sm80` to just build all sparse device unit tests

decioren:
Thanks for that.
I did two things: 
I changed the data initialization (TensorFillRandomUniform) to use bits=-1 rather than '0' as I was getting just zeros in the data.
I changed the data type to half so that debugging it will be easier (4bit types are hard to look at with the debugger)

hwu36:
The fix you need is this:

diff --git a/examples/15_ampere_sparse_tensorop_gemm/ampere_sparse_tensorop_gemm.cu b/examples/15_ampere_sparse_tensorop_gemm/ampere_sparse_tensorop_gemm.cu
index c88a889..d5a2ddb 100644
--- a/examples/15_ampere_sparse_tensorop_gemm/ampere_sparse_tensorop_gemm.cu
+++ b/examples/15_ampere_sparse_tensorop_gemm/ampere_sparse_tensorop_gemm.cu
@@ -111,7 +111,8 @@ using Gemm = cutlass::gemm::device::SparseGemm<ElementInputA,
 
 // Data type and layout of meta data matrix E can be inferred from template Gemm.
 using ElementInputE = typename Gemm::ElementE;
-using LayoutInputE = typename Gemm::LayoutE;
+using LayoutInputE = cutlass::layout::RowMajor;
+using ReorderedLayoutInputE = typename Gemm::LayoutE;
 
 // Blow property is defined in include/cutlass/arch/sp_mma_sm80.h
 // 50% Sparsity on Ampere
[...]

Issue #224, opened 2021-04-05
Compilation error in 3D Convolution


rongzhao-zhang:
Hi, when I was trying to modify a 2D convolution example to a 3D one, I got a compilation error about tile iterator, saying
`cutlass/include/cutlass/conv/threadblock/implicit_gemm_multistage.h(273): error: class "cutlass::transform::threadblock::RegularTileIterator<cutlass::MatrixShape<128, 16>, ElementInputA, cutlass::layout::RowMajorTensorOpMultiplicandCrosswise<16, 16>, 0, cutlass::transform::PitchLinearWarpRakedThreadMap<cutlass::layout::PitchLinearShape<16, 128>, 128, cutlass::layout::PitchLinearShape<2, 16>, 8>, 16>" has no member "set_iteration_index" 
... ...`

The error trace seems to be:
`result.status = implicit_gemm_op.initialize(arguments, workspace.get());` 
-> 
`cudaError_t result = cudaFuncSetAttribute(cutlass::Kernel<ImplicitGemmKernel>,
                                    cudaFuncAttributeMaxDynamicSharedMemorySize,
                                    smem_size);`
-> 
`op(params, *shared_storage);` -> somewhere I cannot figure out.

The error file and the .cu file are attached here.
[err.txt](https://github.com/NVIDIA/cutlass/files/6257717/err.txt)
[turing_tensorop_conv3dfprop_i8.cu.txt](https://github.com/NVIDIA/cutlass/files/6257720/turing_tensorop_conv3dfprop_i8.cu.txt)

And the compilation command is
`nvcc -forward-unknown-to-host-compiler -DCUTLASS_ENABLE_CUBLAS=1 -DCUTLASS_ENABLE_CUDNN=1 \
	-I/home/rongzhao/github/cutlass/include -I/home/rongzhao/github/cutlass/examples/common \
	-I/home/rongzhao/github/cutlass/build/include -I/usr/local/cuda-10.2/include \
	-I/home/rongzhao/github/cutlass/tools/util/include -O3 -DNDEBUG \
	-Xcompiler=-fPIE -DCUTLASS_ENABLE_TENSOR_CORE_MMA=1 -DCUTLASS_TEST_LEVEL=0 -DCUTLASS_DEBUG_TRACE_LEVEL=0 \
	-Xcompiler=-Wconversion -Xcompiler=-fno-strict-aliasing \
	-gencode=arch=compute_75,code=[sm_75,compute_75] -std=c++11 -MD \
[...]

PR #228, opened 2021-04-07
Fix for issue#224 and issue#225

mani-ananth:
* add 2stage fprop 3d into default file (#224 )
* fix a broken sparse gemm example (#225 )


hwu36:
viz: @rongzhao-zhang , @decioren

hwu36 (approved):

Issue #242, opened 2021-04-19
Unalignment case for conv


mengchihe:
I find out that I can set AlignmentA/B in gemm to handle the case when shape is unalignment for int4.
But how can I run conv in similar case, such as channel equals to 1 or 3.
Is there any configuration  to set global load granularity for conv, thanks.

hwu36:
CONV currently do not support this feature.  However, it is pretty easy to add it.

In GEMM, this [line](https://github.com/NVIDIA/cutlass/blob/c77a52445972fd52210d718d5cc606b37b117e56/include/cutlass/gemm/threadblock/mma_multistage.h#L310) splits a 128-bit load into multiple smaller loads.  You just need to do the same to the conv mainloop in https://github.com/NVIDIA/cutlass/blob/master/include/cutlass/conv/threadblock/implicit_gemm_multistage.h .  The underlying iterators in https://github.com/NVIDIA/cutlass/blob/master/include/cutlass/conv/threadblock and several defaultxxx files need some plumbing, too.

We welcome the community to upstream this feature.

mengchihe:
okay thanks, I will try to upload a patch

hwu36:
Awesome, just first try to figure out how GEMM works, and then just do the same for CONV.  It is not hard.

PR #246, opened 2021-04-21
support unalignment input for conv2d fprop stage=2 Fix for issue #242

mengchihe:
Hi, I add some support for conv2d fprop when input shape is unalignment for issue #242.
I don't use AlignmentA/B like gemm to tell data iterator the granularity of each load, since that will change all the interface in default_conv2d_fprop.h. So I just distinguish whether input data is aligned or not, and read one element each time when not alignment. I enlarge the mask in data iterator and try not affect the original performance when shape is alignment.
I just change 2d fprop yet to see if there is any comments. If this patch is okay, I'll go on to support backward and conv3d, thanks.


hwu36:
Thank you very much!!!  You are really fast.

This is very important.  Important enough to have its own slide in a tier-1 GTC talk if it is done.  So, we need to do it perfectly which means we may have multiple iterations between you and us.

Back to the code. It needs to be done almost the same as GEMM.  If a interface needs to be changed, then change it to be the same as the GEMM counterpart.  Don't create new kernel level structs.  `Alignment` needs to be a `int` rather than a `bool`.  `Alignment = 1` means it is aligned to 1 element.  For example, if the input is fp16.  It should support alignment 1, 2, 4, 8.  The max one is 128bit aligned.

mengchihe:
Okay, I'll change it.
Is this mean that I shouldn't add more partial specialization in default_conv2d_fprop.h and add AlignmentA/B into all existing partial specializations. Since partial specializations can not have default template arguments, we needs to add these two parameters in all example/test .cu files which call DefaultConv2dFprop.
Or may be I should do like gemm which has Gemm and DefaultGemm, and add a new struct named Conv2dFprop which will call default_conv2d_fprop.h. By this way I should change DefaultConv2dFprop in all .cu files into Conv2dFprop.
Bot ways needs to change existed .cu codes, and the second way is more similar to gemm. Which way do you prefer, or is there any way that can avoid changing .cu files. Thanks.

hwu36:
Can we just change https://github.com/NVIDIA/cutlass/blob/master/include/cutlass/conv/kernel/default_conv2d_fprop.h#L50-L70
to

```
/// Defines a kernel for Conv2dFprop
template <
  typename ElementA,
  typename LayoutA,
  typename ElementB,
  typename LayoutB,
  typename ElementC,
  typename LayoutC,
  typename ElementAccumulator,
  typename OperatorClass,
  typename ArchTag,
  typename ThreadblockShape,
  typename WarpShape,
  typename InstructionShape,
  typename EpilogueOutputOp,
  typename ThreadblockSwizzle,
  int Stages,
  typename MathOperatorTag,
  int AlignmentA = 128 / cutlass::sizeof_bits<ElementA>::value,
  int AlignmentB = 128 / cutlass::sizeof_bits<ElementB>::value,
  conv::IteratorAlgorithm IteratorAlgorithm = IteratorAlgorithm::kAnalytic,
  conv::StrideSupport StrideSupport = StrideSupport::kStrided
> struct DefaultConv2dFprop;
```

Then, add `AlignmentA` and `AlignmentB` to all the partial specialization below.  I think you don't need to change any `.cu` file then unless you would like to use non-default smaller alignment.

BTW, we only need this feature for fp16/bf16/tf32 tensor core kernels so far.

mengchihe:
Oops, exposing my poor cpp level, I didn't know that partial specialization also have the same default configuration.

I'll change like that, thank you.

hwu36:
> Oops, exposing my poor cpp level, I didn't know that partial specialization also have the same default configuration.

Everyone starts from somewhere.  You start with a high profile project. :smile: 

Don't restrict your change to stage=2.  The change is almost the same for stage>2

include/cutlass/conv/kernel/default_conv2d_fprop.h
```diff
@@ -597,7 +623,7 @@ struct DefaultConv2dFprop <
     arch::CacheOperation::Always,
     IteratorB,
     SmemIteratorB,
-    arch::CacheOperation::Global,
+    arch::CacheOperation::Always,
```
hwu36, line 626: i don't think we need to change this.

include/cutlass/conv/threadblock/conv2d_fprop_filter_tile_access_iterator_optimized.h
```diff
@@ -166,11 +170,14 @@ class Conv2dFpropFilterTileAccessIteratorOptimized{
...
-      predicates_ |= (pred << s);
+      CUTLASS_PRAGMA_UNROLL
+      for (int v_idx = 0; v_idx < kAccessesPerVector; ++v_idx) {
+        predicates_[v_idx] |= (pred << s);
+      }
     }
 
-    if (filter_c_ >= problem_size.C) {
-      predicates_ = 0u;
+    for (int v_idx = 0; v_idx < kAccessesPerVector; ++v_idx) {
```
hwu36, line 179: you neeed `CUTLASS_PRAGMA_UNROLL` here

include/cutlass/conv/threadblock/conv2d_fprop_filter_tile_access_iterator_optimized.h
```diff
@@ -180,11 +187,44 @@ class Conv2dFpropFilterTileAccessIteratorOptimized{
...
+      :
+        "=r"(predicates_[index])
+      : 
+        "r"((int)clear),
+        "r"(predicates_[index])
+    );
+    #else
+      if (clear) {
+        predicates_[index] = 0;
+        predicates_[index] = 0;
```
hwu36, line 215: these two lines are the same.

include/cutlass/conv/threadblock/conv2d_fprop_filter_tile_access_iterator_optimized.h
```diff
@@ -206,9 +246,9 @@ class Conv2dFpropFilterTileAccessIteratorOptimized{
       next = params_.inc_next_c;
       filter_c_ += params_.filter_c_delta;
     }
-      
-    if (filter_c_ >= problem_size_.C) {
-      predicates_ = 0;
+
```
hwu36, line 249: pragma unroll

hwu36 (commented):
I think you need add unit tests in https://github.com/NVIDIA/cutlass/tree/master/test/unit/conv/device to test your changes.

include/cutlass/conv/kernel/default_conv2d_fprop.h
mengchihe: I change here because there is an assert in cp.async.cg says it only support 16B access size

hwu36: I see.

You need something like this for `B`: https://github.com/NVIDIA/cutlass/blob/master/include/cutlass/gemm/threadblock/default_mma.h#L427-L430

When it is 128bit use `Global`, use `Always` when it is smaller than 128bit.

mengchihe: For unit tests, do I just need to add one more TestAllConv2d in fp16_sm75/80.cu and tf32_sm80.cu with one or several unalignment case, or should I add additional tests in all relevant cu files.

hwu36: just adding align=1,2,4 tests to  fp16_sm75/80.cu and align=1,2 to `tf32_sm80.cu`

mengchihe: I add some unit tests in the new patch

include/cutlass/conv/threadblock/conv2d_fprop_filter_tile_access_iterator_optimized.h
hwu36: I still don't see `CUTLASS_PRAGMA_UNROLL` here

include/cutlass/conv/threadblock/conv2d_fprop_filter_tile_access_iterator_optimized.h
hwu36: Is this fixed?

include/cutlass/conv/threadblock/conv2d_fprop_filter_tile_access_iterator_optimized.h
hwu36: We need `CUTLASS_PRAGMA_UNROLL` before every for loop.  It is essential for the compiler to map the fragments/arrays to the registers.  Otherwise, the compiler would choose to use local memory which will slow the performance by a lot.

include/cutlass/conv/kernel/default_conv2d_fprop.h
hwu36: The newly added unit tests look good.  Have you tried running all the conv unit tests including the one you added?

hwu36 (changes requested):

include/cutlass/conv/threadblock/conv2d_fprop_filter_tile_access_iterator_optimized.h
mengchihe: Oh, I update these three position in the new patch below, and I just tried the newly added unit tests, it seems so long to run all tests, I'll try them now.

hwu36 (commented):
Thanks.  We will run more internal tests before we merge this.  As I said in the other pull request, our test system has backlogs.  It may take a while.

mengchihe:
Okay, thanks.

hwu36:
Sorry for the very late response.  I tried your change and they are great, I just need to do some minor changes and I will push to your branch directly.

Thank you very much.  It is a great new feature to everyone!

mengchihe:
> Sorry for the very late response. I tried your change and they are great, I just need to do some minor changes and I will push to your branch directly.
> 
> Thank you very much. It is a great new feature to everyone!

Okay, I'll upload backward then

hwu36 (approved):

PR #277, opened 2021-06-11
Support parallel split K mode for porfiling

Peter9606:
I'm trying to add support for parallel profiling, and this patch is what I modified. 
Unfortunately, it only works for very small portion of problem sizes whose `m` should equal `n`. Also, to make it workable, I have to hard code the number of elements computed per operation during epilogue to 1 which is obviously not correct. Hope someone can correct it.

### DIFF
--- a/tools/library/src/gemm_operation.h
+++ b/tools/library/src/gemm_operation.h
@@ -566,7 +566,9 @@ protected:
     operator_args.ldb = int(configuration->ldb);
     operator_args.ldc = int(configuration->ldc);
     operator_args.ldd = int(configuration->ldd);
-    
+
+    operator_args.batch_stride_D = int(configuration->problem_size.m()) * int(configuration->problem_size.n());
+
     return Status::kSuccess;
   }
 
--- a/tools/library/src/reduction/reduction_device.cu
+++ b/tools/library/src/reduction/reduction_device.cu
@@ -46,7 +46,7 @@ void initialize_reduce_add_linear_combination_f32_f32_f16(Manifest &manifest) {
 
   using EpilogueOutputOp = cutlass::epilogue::thread::LinearCombination<
     ElementOutput,
-    128 / cutlass::sizeof_bits<ElementOutput>::value,
+    1,
     ElementAccumulator,
     ElementCompute
   >;
@@ -81,7 +81,7 @@ void initialize_reduce_add_linear_combination_f32_f32_f32(Manifest &manifest) {
 
   using EpilogueOutputOp = cutlass::epilogue::thread::LinearCombination<
     ElementOutput,
-    128 / cutlass::sizeof_bits<ElementOutput>::value,
+    1,
     ElementAccumulator,
     ElementCompute
   >;
@@ -115,7 +115,7 @@ void initialize_reduce_add_linear_combination_cf32_cf32_cf32(Manifest &manifest)
 
   using EpilogueOutputOp = cutlass::epilogue::thread::LinearCombination<
     ElementOutput,
-    128 / cutlass::sizeof_bits<ElementOutput>::value,
+    1,
     ElementAccumulator,
     ElementCompute
   >;
--- a/tools/profiler/src/gemm_operation_profiler.cu
+++ b/tools/profiler/src/gemm_operation_profiler.cu
@@ -37,6 +37,7 @@
 #include "gemm_operation_profiler.h"
 #include "gpu_timer.h"
 
+#include "cutlass/library/singleton.h"
 #include "cutlass/library/library.h"
 #include "cutlass/library/handle.h"
 
@@ -55,6 +56,7 @@ GemmOperationProfiler::GemmOperationProfiler(Options const &options):
     library::OperationKind::kGemm,
     {
       {ArgumentTypeID::kEnumerated, {"gemm_kind"}, "Variant of GEMM (gemm, batched, array, universal, planar_complex, planar_complex_array)"},
+      {ArgumentTypeID::kEnumerated, {"split_k_mode"}, "Variant of split K mode(serial, parallel)"},
       {ArgumentTypeID::kInteger, {"m", "problem-size::m"}, "M dimension of the GEMM problem space"},
       {ArgumentTypeID::kInteger, {"n", "problem-size::n"}, "N dimension of the GEMM problem space"},
       {ArgumentTypeID::kInteger, {"k", "problem-size::k"}, "K dimension of the GEMM problem space"},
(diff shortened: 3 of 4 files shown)


hwu36:
@manishucsd and @kerrmudgeon ,  would you please help @Peter9606 here? Maybe point him to the conv parallel splitk code?

manishucsd:
Hi @Peter9606 ,  I have reviewed your code. It looks like you have followed the changes in convolutions to support parallel reductions. 

A few points to consider:
1. Can you check that everything for the reduction kernel is [initialized](https://github.com/NVIDIA/cutlass/pull/277/files#diff-885944534dbe00b4515752f80ac452d08402147060dece0ae312136938ccec4aR416) properly.

2. Also, check at the dispatch point from the profiler to the actual kernel. Use these [printfs](https://github.com/NVIDIA/cutlass/blob/3584270b6e00950d476513f85ea174b79c42007a/tools/library/src/reduction/reduction_operation.h#L253).

3. Make sure the reduction kernel you are trying to use is instantiated and compiled into the CUTLASS library. 
- Reduction kernels are manually instanced in [reduction_device.cu ](https://github.com/NVIDIA/cutlass/blob/3584270b6e00950d476513f85ea174b79c42007a/tools/library/src/reduction/reduction_device.cu)
- It currently has the largest aligned kernels. If you need smaller alignments you will need to instance them and not overwrite them. 
- Also, to use smaller alignment you will need to add something similar to [GemmPreferenceKey](https://github.com/NVIDIA/cutlass/blob/3584270b6e00950d476513f85ea174b79c42007a/tools/library/include/cutlass/library/operation_table.h#L175) which handles alignment. Note that [ReductionFunctionalKey](https://github.com/NVIDIA/cutlass/blob/3584270b6e00950d476513f85ea174b79c42007a/tools/library/include/cutlass/library/operation_table.h#L370) will give the list of all kernels that match the functional requirement. There could be many possible alignments that functionally satisfy the problem size. It will pick the largest possible alignment from a functionally equivalent kernel set. 
- I recommend making Gemm parallel split-k work for the largest alignment first. Make it work for F16 align8 kernels first.

tools/library/src/gemm_operation.h
```diff
@@ -566,7 +566,9 @@ class GemmUniversalOperation : public GemmOperationBase<Operator_> {
     operator_args.ldb = int(configuration->ldb);
     operator_args.ldc = int(configuration->ldc);
     operator_args.ldd = int(configuration->ldd);
-    
+
+    operator_args.batch_stride_D = int(configuration->problem_size.m()) * int(configuration->problem_size.n());
```
manishucsd, line 570: I don't think this change is required. Can you explain what are you trying to achieve here? Also, will this be over-written by run() -> update().

tools/library/src/reduction/reduction_device.cu
```diff
@@ -46,7 +46,7 @@ void initialize_reduce_add_linear_combination_f32_f32_f16(Manifest &manifest) {
 
   using EpilogueOutputOp = cutlass::epilogue::thread::LinearCombination<
     ElementOutput,
-    128 / cutlass::sizeof_bits<ElementOutput>::value,
+    1,
```
manishucsd, line 49: If you need smaller alignment add them as a new reduction kernel instance.

tools/profiler/src/gemm_operation_profiler.cu
```diff
@@ -373,6 +396,52 @@ void GemmOperationProfiler::initialize_result_(
...
+  // find reduction operation
+  library::ReductionFunctionalKey reduction_key(
+    library::Provider::kCUTLASS,
+    gemm_desc.tile_description.math_instruction.element_accumulator,    // element workspace
+    gemm_desc.tile_description.math_instruction.element_accumulator,    // element accumulator
+    gemm_desc.C.element,                                                // element output
+    gemm_desc.element_epilogue                                          // element coumpute
+  );
+
+  auto reduction_it = library::Singleton::get().operation_table.reduction_operations.find(reduction_key);
```
manishucsd, line 432: This should get you the largest aligned reduction kernel that satisfies your problem dimension. At the moment, you only have algin1 kernels though. master has only the largest aligned kernels. Try to make it work for the align4 (F32) reduction case first.

tools/library/src/gemm_operation.h
Peter9606: Because I tried to implement parallel split K profiling based on GemmUniversal (It's the right way, isn't?), which needs `args.batch_stride_D` to calculate device workspace size in case of parallel split K (https://github.com/NVIDIA/cutlass/blob/856d4db3fbdee6544d9aebd69b89832f569e68bb/include/cutlass/gemm/device/gemm_universal_base.h#L162).

Peter9606:
>     * For F16 output and F32 accumulation, Gemm + Parallel reduction will change the Gemm kernel you need to call. Instead of calling Gemm with F16 output, now Gemm writes output in F32, and reduction kernel writes it F16 (F32->F16).
> 

I finally get it, thank you very much!

Peter9606:
@manishucsd  Now parallel split K reduction profiling works for SMIT fp32, but only through new added reduction kernel with smaller alignment. Still not very clear why larger alignment reduction kernel cannot be selected.

Peter9606:
Seems that this version can have a successful run for fp16 with align1/2/4.

tools/profiler/src/gemm_operation_profiler.cu
hwu36: @Peter9606 , I am working on merging this PR now.  Sorry that it takes so long.  Would you please elaborate in detail a little bit here how you code can find the max aligned reduction kernel?

Peter9606: If I remembered correctly, there isn't a find "procedure" in this PR, as you can see (from this [file](https://github.com/NVIDIA/cutlass/pull/277/files#diff-88b162874290c7ac5efe445498b47a061aeb55905ee7b7ad1797346fc9b5693e) , only one candidate reduction kernel is appended for each alignment.

hwu36 (approved):

hwu36:
Now it can support 
    fp32 gemm output, fp16 reduction output
    fp32 gemm output, fp32 reduction output

It still requires the 128bit alignment in the reduction.  In this PR, I removed small aligned reduction code which requires some extra logic to find the correct reduction configuration.  To do this, we need to put alignment into the FunctionalKey or PreferenceKey of the reduction operation and use problem_size.m to decide the correct the reduction kernel to use.   We welcome the community to extend this PR to support small alignment reduction.

PR #292, opened 2021-07-08
fix annotation

103yiran:


hwu36 (commented):
Thanks for the PR, but it really means "tune" here, not a specific architecture.  The default is sm70 volta now, but I think we need to bump it up to sm80 since ampere is mainstream now.

103yiran:
thanks for your explain @hwu36

PR #295, opened 2021-07-20
Make cutlass::gemm::device::GemmArray usable

dfyz:
For my application, I want a batched GEMM with no fixed stride between matrices (in cuBLAS terms, `cublas*gemmBatched()` instead of `cublas*gemmStridedBatched()`). There's apparently a class that does exactly this (`cutlass::gemm::device::GemmArray`), however, it uses the wrong argument order for `GemmBatchedIdentityThreadblockSwizzle::get_tiled_block()` and therefore doesn't compile.

In this PR, I swap the order of the arguments and add a demo for `GemmArray` to show that it works, at least for the toy SGEMM in `examples/05_batched_gemm`.

I also tested it on my application (using FP16), and it appears to work perfectly. There might be more to it, but to me it looks like `GemmArray` is identical to `GemmBatched` in every way except actually fetching pointers to the element of the batch, so I hope my patch is not subtly broken.

### DIFF
--- a/examples/05_batched_gemm/batched_gemm.cu
+++ b/examples/05_batched_gemm/batched_gemm.cu
@@ -28,12 +28,16 @@
 
 #include "cutlass/cutlass.h"
 #include "cutlass/layout/matrix.h"
+#include "cutlass/gemm/device/gemm_array.h"
 #include "cutlass/gemm/device/gemm_batched.h"
 
 #pragma warning( disable : 4503)
 
 /*
-This example demonstrates how to use cutlass to compute a batched strided gemm.
+This example demonstrates how to use cutlass to compute a batched gemm in two different ways:
+  1. By specifying pointers to the first matrices of the batch and the stride between the consecutive
+     matrices of the batch (this is called a strided batched gemm).
+  2. By copying pointers to all matrices of the batch to the device memory (this is called an array gemm).
 In this example, both A and B matrix are non-transpose and column major matrix
 batched_C = batched_A x batched_B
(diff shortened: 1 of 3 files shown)


hwu36:
I just went over your change quickly, it looks good to me.

FYI, you can use gemm_universal to run gemm_array: https://github.com/NVIDIA/cutlass/blob/master/include/cutlass/gemm/kernel/gemm_universal.h#L378-L381

dfyz:
> you can use gemm_universal to run gemm_array

Thanks, I wasn't aware of that! Still, I think `GemmArray` provides a cleaner API in my case (e.g., it doesn't have to use [pointers to void](https://github.com/NVIDIA/cutlass/blob/c77a52445972fd52210d718d5cc606b37b117e56/include/cutlass/gemm/kernel/gemm_universal.h#L108), which can be confusing), so it might be a good idea to have `GemmArray` fixed.

I also just added a very brief description of `GemmArray` to the [CUTLASS GEMM API](https://github.com/NVIDIA/cutlass/blob/master/media/docs/gemm_api.md) document so that those who search for it can find more easily. Please let me know if you don't want it mentioned there or if it should be rephrased.

hwu36:
Fixing the broken code is always a good thing.  I will work on this PR soon.

hwu36:
Sorry for the delay.  

This is still in my todo list, but the priority is low since we normally use `GemmUniversal` to do `GemmArray`.  I hope to merge this PR this month.  You don't need to do anything.

hwu36 (approved):

PR #297, opened 2021-07-21
complex exp function bugfix

Peter9606:
exp() for real(z) was missed.


hwu36:
good catch.  we are supposed to use `fast_exp` in `fast_math.h` here.

hwu36:
good catch.

hwu36:
We fixed it in 2.6 release.  Thank you again for catching this.

PR #345, opened 2021-10-15
Add linear combination bias epilogue

Laurawly:


masahi:
@Laurawly We probably don't need this. Bias only epilogue can be implemented by `LinearCombination` with `NoBetaScaling`. Our PR https://github.com/apache/tvm/pull/9261 doesn't use `LinearCombinationBias` anymore.

PR #363, opened 2021-11-19
CUTLASS 2.8

manishucsd:
CUTLASS 2.8 is an update to CUTLASS adding:
- [TF32x3:](/examples/27_ampere_3xtf32_fast_accurate_tensorop_gemm) emulated single-precision using Tensor Cores; 45+ TFLOPs on NVIDIA A100
- [Mainloop fusion for Convolution:](/examples/25_ampere_fprop_mainloop_fusion) convolution with fused per-channel bias-add
- [Grouped GEMM:](/examples/24_gemm_grouped) similar to batched GEMM with distinct problem size per group
- [Implicit GEMM Convolution fusion](/examples/13_two_tensor_op_fusion/) supports staging 1st convolution's output accumulator in the shared memory on Turing.
- Optimal performance using [CUDA 11.5](https://developer.nvidia.com/cuda-downloads)
- CUTLASS plans to **deprecate** the following platforms in the future. Let us know if this affects your use case.
  - Maxwell and Pascal GPU architectures
  - Ubuntu 16.04
  - CUDA 10.2
- Updates and bugfixes from the community (thanks!)


kerrmudgeon (approved):
CUTLASS 2.8 looks good to me.

PR #379, opened 2021-12-13
Refactor GELU and Sigmoid epilogue to use a common template (and add SiLu, Hardswish epilogue)

masahi:
Since Sigmoid and GeLU epilogue functors are almost identical, I propose to refactor them. The motivation is to make it easy to add another functors.

Now Sigmoid and GeLU epilogue functors are defined as:
```
using LinearCombinationSigmoid = LinearCombinationGeneric<Sigmoid, ElementOutput_, Count, ElementAccumulator_,
                                                          ElementCompute_, FloatRoundStyle::round_to_nearest>;

using LinearCombinationGELU = LinearCombinationGeneric<GELU, ElementOutput_, Count, ElementAccumulator_,
                                                          ElementCompute_, FloatRoundStyle::round_to_nearest>;
```

I'm going to add two new functors for `SiLU` and `HardSwish` activation, which can be implemented in the same manner.

What do you think? @hwu36

### DIFF
--- a/include/cutlass/epilogue/thread/activation.h
+++ b/include/cutlass/epilogue/thread/activation.h
@@ -124,6 +124,25 @@ struct Sigmoid<Array<T, N> > {
   }
 };
 
+// SiLu (swish) operator
+template <typename T>
+struct SiLu {
+  CUTLASS_HOST_DEVICE
+  T operator()(T const &scalar) const {
+    return scalar * Sigmoid<T>(scalar);
+  }
+};
+
+template <typename T, int N>
+struct SiLu<Array<T, N>> {
+  CUTLASS_HOST_DEVICE
+  Array<T, N> operator()(Array<T, N> const &rhs) const {
+    Sigmoid<Array<T, N>> sigmoid_op;
+    multiplies<Array<T, N>>     mul;
+    return mul(rhs, sigmoid_op(rhs));
+  }
+};
+
 //
 // GELU function definitions implemented as described by
 //   Hendrycks, D., and Gimpel, K. in
--- a/include/cutlass/epilogue/thread/activation.h
+++ b/include/cutlass/epilogue/thread/activation.h
@@ -171,7 +171,7 @@ struct HardSwish<Array<T, N> > {
     HardSwish<T> hardswish_op;
 
     CUTLASS_PRAGMA_UNROLL
-    for (int i = 0; i < int(rhs.size()); ++i) {
+    for (int i = 0; i < N; ++i) {
       y[i] = hardswish_op(rhs[i]);
     }
 
--- a/include/cutlass/epilogue/thread/activation.h
+++ b/include/cutlass/epilogue/thread/activation.h
@@ -172,6 +172,32 @@ struct SiLu<Array<T, N>> {
   }
 };
 
+template <typename T>
+struct HardSwish {
+  CUTLASS_HOST_DEVICE
+  T operator()(T const &x) const {
+    minimum<T> mn;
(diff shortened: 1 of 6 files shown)


hwu36:
@shangz-ai, is it you that told me that you need to use fp32 to compute sigmoid for the accuracy?

masahi:
ok I've pushed two activation using this common template.

* SiLU https://pytorch.org/docs/stable/generated/torch.nn.SiLU.html
* HardSwish https://pytorch.org/docs/stable/generated/torch.nn.Hardswish.html

masahi:
@d-k-b Thanks for the feedback, the doc is updated.

hwu36:
Overall, looks good to me.  I think just change `FloatRoundStyle::round_to_nearest` to `Round` everywhere.

masahi:
Oops thanks for spotting this, fixed.

masahi:
Not sure why Ci has failed

include/cutlass/epilogue/thread/activation.h
```diff
@@ -124,6 +124,61 @@ struct Sigmoid<Array<T, N> > {
...
+
+template <typename T, int N>
+struct HardSwish<Array<T, N> > {
+  CUTLASS_HOST_DEVICE
+  Array<T, N> operator()(Array<T, N> const &rhs) const {
+    Array<T, N> y;
+    HardSwish<T> hardswish_op;
+
+    CUTLASS_PRAGMA_UNROLL
+    for (int i = 0; i < int(rhs.size()); ++i) {
```
hwu36, line 174: Shall we replace `rhs.size()` to `N` to emphasize that it is a compile time constant?

hwu36:
These 4 activations look expensive to me.  You can try to set `kIsHeavy` to `true`, check the generated code is different, and compare the performance to verify it.

If you setting them to `true`, the epilogue will not be fully unrolled which is good for the I$.  Usually complex activations such as `gelu` need to be set to `true`.

masahi:
> These 4 activations look expensive to me. You can try to set `kIsHeavy` to `true`, check the generated code is different, and compare the performance to verify it.
> 
> If you setting them to `true`, the epilogue will not be fully unrolled which is good for the I$. Usually complex activations such as `gelu` need to be set to `true`.

My patch respects the current choice for the existing epilogues: `isHeavy = true` for GELU and whatever the default for Sigmoid which I assumed was `false` . For the two new activations, my reasoning was `false` for Silu because it is just sigmoid + mul, and also `false` for hardswish since it is similar to relu.

hwu36:
SGTM.

`kIsHeavy` was newly introduced.  We benchmarked `gelu`, but not `sigmoid`.  If you find `sigmoid` is expensive in the future, you can create a PR and flip the bool.

masahi:
Making sigmoid and silu `isHeavy = true` did improve performance on efficientnetv2 slightly: 8.58 msec -> 8.26 msec where 8.58 was the number from my PR https://github.com/apache/tvm/pull/9746. So I pushed that change. The change didn't help when the vectorized variant of sigmoid is used.

hwu36:
> The change didn't help when the vectorized variant of sigmoid is used.

Vectorized one has less instruction and less burden to I$.

hwu36:
> Making sigmoid and silu isHeavy = true did improve performance on efficientnetv2 slightly: 8.58 msec -> 8.26 msec where 8.58 was the number from my PR apache/tvm#9746.

You can update your TVM PR now. :)

hwu36 (approved):

Issue #387, opened 2021-12-21
[BUG] Installing with HEADERS ONLY Set Does Not Include All Necessary Headers - v2.8.0


depaulmillz:
In v2.8.0, when installing through CPack and CMake with CUTLASS_ENABLE_HEADERS_ONLY set ON, the files in tools/util/include are not installed. The files are however referenced throughout the headers cutlass. For example: include/cutlass/epilogue/threadblock/epilogue.h references a header in util. This prevents using the library if its been installed as header only.

To reproduce the bug, run `cmake` with `-DCUTLASS_ENABLE_HEADERS_ONLY=ON`. Then in the build directory run `cpack -G ZIP`. After unzipping you will notice the include folder does not contain the util folder.

I think when `CUTLASS_ENABLE_HEADERS_ONLY=ON` that the headers in util should be included.

hwu36:
@d-k-b, would you please take a look at the PR ?

d-k-b:
Fix with https://github.com/NVIDIA/cutlass/pull/388.

PR #388, opened 2021-12-21
Fix utils include not being installed in header only

depaulmillz:
The util include files are not being installed currently when installing with header only set to on. This makes it difficult to use cutlass if it is installed as a header only library. This is a potential fix for that. This issue is described in #387.


d-k-b:
@depaulmillz, sorry for the delay, I will try to review this soon.

d-k-b (approved):

CMakeLists.txt
```diff
@@ -67,21 +67,23 @@ set(CUTLASS_ENABLE_HEADERS_ONLY OFF CACHE BOOL "Enable only the header library")
...
 set(CUTLASS_ENABLE_EXAMPLES ${CUTLASS_ENABLE_EXAMPLES_INIT} CACHE BOOL "Enable CUTLASS Examples")
 set(CUTLASS_ENABLE_TOOLS ${CUTLASS_ENABLE_TOOLS_INIT} CACHE BOOL "Enable CUTLASS Tools")
-set(CUTLASS_ENABLE_LIBRARY ${CUTLASS_ENABLE_TOOLS} CACHE BOOL "Enable CUTLASS Library")
-set(CUTLASS_ENABLE_PROFILER ${CUTLASS_ENABLE_TOOLS} CACHE BOOL "Enable CUTLASS Profiler")
+set(CUTLASS_ENABLE_LIBRARY ${CUTLASS_ENABLE_LIBRARY_INIT} CACHE BOOL "Enable CUTLASS Library")
+set(CUTLASS_ENABLE_PROFILER ${CUTLASS_ENABLE_LIBRARY} CACHE BOOL "Enable CUTLASS Profiler")
 
 if(${CMAKE_PROJECT_NAME} STREQUAL ${PROJECT_NAME})
-  set(CUTLASS_ENABLE_TESTS_INIT ${CUTLASS_ENABLE_TOOLS_INIT})
+  set(CUTLASS_ENABLE_TESTS_INIT ${CUTLASS_ENABLE_LIBRARY_INIT})
```
d-k-b, line 86: This should use `${CUTLASS_ENABLE_LIBRARY}` to capture user override.

d-k-b (changes requested):
One minor fix needed.

CMakeLists.txt
depaulmillz: Done.

d-k-b (approved):

PR #391, opened 2021-12-23
Add epilogue functor for residual block fusion

masahi:
Following the discussion in https://github.com/NVIDIA/cutlass/discussions/347, this is my proposal for enabling full fusion of residual block in deep neural network, of the form `UnaryOp(BinaryOp(ActivationOp(TensorOp(X) + bias), residual))` where `residual` has the same shape as the output tensor. 

As discussed in https://github.com/NVIDIA/cutlass/discussions/347, `LinearCombinationBiasElementwise` epilouge already supports `elementwise((alpha x conv + beta x C) + per_channel_bias)`. What we need in deep learning, however, is `elementwise((conv + per_channel_bias) + C)`. Initially I extended `LinearCombinationBiasElementwise` with another template parameter to specify in which order to do two additions, but I decided to add a new implementation of an epilogue functor from scratch for the following reasons:
* We need `ActivationOp` template parameter to express the activation function to use after broadcast addition of per-channel bias
* We don't need `T` output tensor (want to set `kStoreT = false`)
* Less template paramters, More descriptive variable names than `C`, `V` etc
* Use vectorization if possible

TODO: Currently [one of the two new tests added in `conv2d_fprop_with_broadcast_sm75.cu`](https://github.com/NVIDIA/cutlass/blob/5603a9345e6679b701f5d25d99f4282cf9aa0aec/test/unit/conv/device/conv2d_fprop_with_broadcast_sm75.cu#L154-L158),  fails on `split_k_slice > 1` cases. Any hint on how to debug? I don't understand the logic in https://github.com/NVIDIA/cutlass/blob/808c25337a3ed4c97ac21895257b1addc72d6ca8/include/cutlass/conv/kernel/implicit_gemm_convolution_with_fused_epilogue.h#L451-L455 @hwu36

### DIFF
--- a/test/unit/conv/device/conv2d_with_broadcast_testbed.h
+++ b/test/unit/conv/device/conv2d_with_broadcast_testbed.h
@@ -491,7 +491,8 @@ public:
 /////////////////////////////////////////////////////////////////////////////////////////////////////////////
 template <typename ImplicitGemm,
           typename ReferenceOp = Conv2dWithBroadcastReferenceOp<ImplicitGemm>,
-          bool AddBroadcastFirst = false>
+          bool AddBroadcastFirst = false,
+	  bool TestSplitK = true>
 bool TestAllConv2dWithBroadcast(
   const Conv2dProblemVector &conv_test_sizes = Conv2dProblemVector(),
   const Conv2dProblemVector &conv_blacklist_sizes = Conv2dProblemVector()) {
@@ -637,8 +638,8 @@ bool TestAllConv2dWithBroadcast(
     2.0
   };
 
-  // TODO(masahi): Fix tests below for kAddbroadcastfirst = true and split_k case
-  return passed;
+  if (!TestSplitK)
+    return passed;
 
   for (auto split_k_mode : split_k_modes) {
     for (auto split_k_slice : split_k_slices) {
--- a/test/unit/conv/device/conv2d_with_broadcast_testbed.h
+++ b/test/unit/conv/device/conv2d_with_broadcast_testbed.h
@@ -610,6 +610,9 @@ bool TestAllConv2dWithBroadcast(
     return passed;
   }
 
+  if (!TestSplitK)
+    return passed;
+
   // Sweep split-k-slice using serial and prallel reduction with non-unity alpha and non-zero beta for 
   // a single conv2d problem size. Convolution unit tests take a long time to run so only sweep parameters 
   // which are abolutely neccessary to catch functional bugs. The below code does provide option to sweep 
@@ -637,9 +640,6 @@ bool TestAllConv2dWithBroadcast(
   double problem_beta[] = {
     2.0
   };
-
-  if (!TestSplitK)
-    return passed;
 
   for (auto split_k_mode : split_k_modes) {
     for (auto split_k_slice : split_k_slices) {
--- a/test/unit/conv/device/conv2d_with_broadcast_testbed.h
+++ b/test/unit/conv/device/conv2d_with_broadcast_testbed.h
@@ -95,7 +95,8 @@ struct Conv2dWithBroadcastReferenceOp {
 
 template <
   typename Conv2d,
-  typename ReferenceOp = Conv2dWithBroadcastReferenceOp<Conv2d>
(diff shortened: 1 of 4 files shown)


hwu36:
To answer your splitk question, suppose you have two threadblocks, tb0 and tb1, processing the channels in the same n/p/q.  tb0 will process the first half of the channels, and write the result to the global memory, and tb1 will process the 2nd half of the channels and add the result of tb0 and write out the final result to the global memory at last.

tb0 and tb1 use a semaphore to maintain the execution order [here](https://github.com/NVIDIA/cutlass/blob/master/include/cutlass/conv/kernel/implicit_gemm_convolution_with_fused_epilogue.h#L440).  tb0 is first, tb1 is the 2nd.  tb0 and tb1 also share the exactly same output tensor space.  tb0 writes its partial result to this space first, tb1 then read this partial result and write the final result to it eventually.

In the case of the original broadcast fused conv, when tb0 reaches this [line](https://github.com/NVIDIA/cutlass/blob/master/include/cutlass/conv/kernel/implicit_gemm_convolution_with_fused_epilogue.h#L375), `skip_elementwise_` will be set to `true` [here](https://github.com/NVIDIA/cutlass/blob/master/include/cutlass/epilogue/thread/linear_combination_bias_elementwise.h#L180).  tb0's bias vector is set to be `nullptr` [here](https://github.com/NVIDIA/cutlass/blob/master/include/cutlass/conv/kernel/implicit_gemm_convolution_with_fused_epilogue.h#L454) which means it will skip bias vector.  Then tb0 just compute `alpha x first_half_of_conv + beta x C` [here](https://github.com/NVIDIA/cutlass/blob/master/include/cutlass/epilogue/thread/linear_combination_bias_elementwise.h#L205) and eventually output it to the global memory.

As to tb1, its `C` tensor is hard coded to be the output of the tb0's input [here](https://github.com/NVIDIA/cutlass/blob/master/include/cutlass/conv/kernel/implicit_gemm_convolution_with_fused_epilogue.h#L437) and tb1's `beta` is hard coded to be `1` [here](https://github.com/NVIDIA/cutlass/blob/master/include/cutlass/epilogue/thread/linear_combination_bias_elementwise.h#L176).  So `z = elementwise(plus(alpha_ x tmp_Accum[i] + beta_ x tmp_C[i], V[i])) = elementwise(plus(alpha x second_half_of_conv + tb0_parital_result, V[i]))` which is what we want eventually.

masahi:
@hwu36 Thank you very much, it is very interesting! The way split-k is done in conv2d was mysterious to me, since I always thought split-k requires two kernel launches but I didn't find the second one.

 Now I understand where the bug is in my code: I have to skip the activation in `UnaryOp(BinaryOp(ActivationOp(TensorOp(X) + bias), residual))` as well when `skip_elemwise = true`.  One of two new tests is passing because it's `ActivationOp` is identity. 

I'll update the PR next week.

masahi:
@hwu36 I realized that, the need to apply an activation op before `BinaryOp` in `BinaryOp(ActivationOp(TensorOp(X) + bias), residual)`, makes it impossible to support split-k unless `ActivationOp` is `Identity`. `tb0` computes `alpha x first_half_of_conv + beta x C` but the activation should not be applied to the `beta x C` part.

So I want to drop support for split-k when using this new epilogue functor, unless the activation op is `Identity`. Otherwise I had to change a lot of interfaces to plumb through all required arguments to the last block, where tthe activation and elemwise op are applied. Let me know if that's acceptable.

hwu36:
> Let me know if that's acceptable.

It is okay.  Please add a comment in the beginning of the testbed to explain the reason.

I will take a look of your code tomorrow.

test/unit/conv/device/conv2d_with_broadcast_testbed.h
```diff
@@ -625,6 +638,9 @@ bool TestAllConv2dWithBroadcast(
     2.0
   };
 
+  if (!TestSplitK)
```
hwu36, line 641: these 2 lines can be moved up to line 612.

hwu36 (approved):

PR #402, opened 2022-01-30
Enable convolution with fused epilogue for Volta Tensor Cores

kerrmudgeon:
Enable convolution with fused epilogue for Volta Tensor Cores


hwu36 (approved):

antinucleon:
Thanks @hwu36 @kerrmudgeon! I can confirm this PR fixed the bug on V100.

PR #403, opened 2022-02-02
[hardswish] correct implmentation

antinucleon:
14% speed up on hardswish fusion

### DIFF
--- a/include/cutlass/epilogue/thread/activation.h
+++ b/include/cutlass/epilogue/thread/activation.h
@@ -1,5 +1,5 @@
 /***************************************************************************************************
- * Copyright (c) 2017-2021, NVIDIA CORPORATION.  All rights reserved.
+ * Copyright (c) 2017-2022, NVIDIA CORPORATION.  All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without modification, are permitted
  * provided that the following conditions are met:
@@ -105,7 +105,6 @@ struct ReLu<Array<T, N>> {
     }
     return result;
   }
-
 };
 
 // Sigmoid operator
@@ -125,7 +124,7 @@ struct Sigmoid<Array<T, N> > {
     Sigmoid<T> sigmoid_op;
 
     CUTLASS_PRAGMA_UNROLL
-    for (int i = 0; i < int(rhs.size()); ++i) {
+    for (int i = 0; i < N; ++i) {
       y[i] = sigmoid_op(rhs[i]);
     }
 
@@ -190,7 +189,23 @@ template <typename T>
 struct HardSwish {
   CUTLASS_HOST_DEVICE
   T operator()(T const &x) const {
-    return ((x + 3.0f) > 0) * ((x + 3.0f) > 6.0f ? 6.0f: (x + 3.0f)) * x * 0.16666667f;
+    minimum<T> mn;
+    maximum<T> mx;
+    T relu6 = mn(mx(x + T(3), T(0)), T(6));
+    return x * relu6 / T(6);
+  }
+};
+
+template <>
+struct HardSwish<float> {
+  using T = float;
+
+  CUTLASS_HOST_DEVICE
+  T operator()(T const &x) const {
+    minimum<T> mn;
+    maximum<T> mx;
+    T relu6 = mn(mx(x + T(3), T(0)), T(6));
+    return x * relu6 * 0.16666667f;
   }
 };
 
@@ -207,6 +222,21 @@ struct HardSwish<Array<T, N> > {
     }
 
     return y;
+  }
+};
+
+template <int N>
+struct HardSwish<Array<half_t, N> > {
+  using T = half_t;
+
+  CUTLASS_HOST_DEVICE
+  Array<T, N> operator()(Array<T, N> const &rhs) const {
+    minimum<Array<T, N> > mn;
+    maximum<Array<T, N> > mx;
+    multiplies<Array<T, N> > mul;
+    plus<Array<T, N> > add;
+ 
+    return mul(mul(mn(mx(add(rhs, T(3)), T(0)), T(6)), rhs), T(0.16666667f));
   }
 };
 
--- a/include/cutlass/epilogue/thread/activation.h
+++ b/include/cutlass/epilogue/thread/activation.h
@@ -59,35 +59,35 @@ struct Identity {
 template <typename T>
 struct ReLu {
   static const bool kIsHeavy=false;
-
   CUTLASS_HOST_DEVICE
   T operator()(T const & threshold, T value) const {
     maximum<T> mx;
+
     return mx(value, threshold);
   }
 
   CUTLASS_HOST_DEVICE
-  T operator()(T const &x) const {
+  T operator()(T value) const {
     maximum<T> mx;
-    return mx(x, T(0));
+
+    return mx(value, T(0));
   }
 };
 
 template <typename T, int N>
-struct ReLu<Array<T, N> > {
+struct ReLu<Array<T, N>> {
   static const bool kIsHeavy=false;
-
   CUTLASS_HOST_DEVICE
   Array<T, N> operator()(T const & threshold, Array<T, N> const &frag) const {
-    Array<T, N> result;
     maximum<Array<T, N> > mx;
-    return mx(frag, threshold);
+
+    return mx(threshold, frag);
   }
 
   CUTLASS_HOST_DEVICE
-  Array<T, N> operator()(Array<T, N> const &rhs) const {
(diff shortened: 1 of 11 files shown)


hwu36:
Attn @masahi

masahi:
So what was the bug?

hwu36:
Scale was not supported in Generic.

I also added some optimizations.

masahi:
I don't think we need scale for hardswish. For now generic is used for sigmoid, hardswish, and silu, all of them just use scale = 1. So in what sense my implementation was "not correct"?

hwu36:
I mean this [one](https://github.com/NVIDIA/cutlass/blob/master/include/cutlass/epilogue/thread/scale_type.h#L42-L50) is not supported.

masahi:
Yeah I know that, I didn't add it because for activation I tested, it is always assumed to be `NoBetaScaling`. I'm asking about the PR title. Are you saying, there is a case hardswish can be used with beta scaling != 1?

hwu36:
The original code has this: `intermediate = mul_add_source(beta_, converted_source);`  beta is not ignored.

masahi:
https://github.com/NVIDIA/cutlass/blob/0dc3ba60b310a7e8915d235613e88bdbf5bf21b5/include/cutlass/epilogue/thread/linear_combination_generic.h#L166

I think it is taken into account? Note that what I did in that PR was just refactoring, I simply copy-pasted existing code.

hwu36:
I understand. I am not saying it is your fault. Just want to bring it to your attention that you may want to set scaletype explicitly after this pr gets merged.  Also i think i significantly improve the f32/f16 hardswish performance.

masahi:
I'm still confused with the PR title - Is computing `x * (relu6 / T(6))` for hardswish incorrect?

antinucleon:
> I'm still confused with the PR title - Is computing `x * (relu6 / T(6))` for hardswish incorrect?

Sorry for the title, but in general we don't use ```/``` in both CPU and GPU if possible. As we can see simple replace ```/``` will increase perf by 14%

masahi:
Yeah, I thought the compiler would take care of that.

masahi:
I suggest splitting this PR into perf improvement (vectorized activation) and functionality addition (scale support) to make it less confusing.

masahi:
> Sorry for the title, but in general we don't use `/` in both CPU and GPU if possible. As we can see simple replace `/` will increase perf by 14%

@antinucleon Isn't the speedup due to vectorization?

antinucleon:
> Yeah, I thought the compiler would take care of that.

Also the original implementation can' pass numerical test. So it is "incorrect".

masahi:
Which numerical test? I also remember there is a slight difference in output depending on the associativity. But it depends on what implementation to use as a reference.

antinucleon:
> 



> > Sorry for the title, but in general we don't use `/` in both CPU and GPU if possible. As we can see simple replace `/` will increase perf by 14%
> 
> @antinucleon Isn't the speedup due to vectorization?

vectorization bings a little extra. replace ```/``` has raised the bar to very high level.

antinucleon:
> Which numerical test? I also remember there is a slight difference in output depending on the associativity. But it depends on what implementation to use as a reference.
> 
> I think I picked the implementation that matched whatever TVM reference I had at that time https://github.com/apache/tvm/blob/a1f51aa230d33bed831b65eec6e209733dbfec57/tests/python/contrib/test_cutlass.py#L150-L154

I have some Pytorch test and unfortunately old one was not able to pass. I first located the bug around beta scaling and thanks to @hwu36 it is fixed in general also bring extra speed up by vectorization epilogue.

masahi:
> I first located the bug around beta scaling

You mean your beta is not 1?

antinucleon:
> > I first located the bug around beta scaling
> 
> You mean your beta is not 1?

Yes ```NoBetaScaling``` is able to pass the test.

masahi:
ah may be you don't have a bias. If that's the case, you can also pass `beta = 0` I believe.

antinucleon:
> ah may be you don't have a bias. If that's the case, you can also pass `beta = 0` I believe.

It is not this case. I have bias. In most neural networks, we don't need beta scaling for bias add. So ```NoBetaScaling``` is:
```
act(alpha x AB[n,h,w,c] + bias[n,h,w,c])
```

masahi:
Ok I understand the situation better. All benchmark and tests I used in TVM are of that form, but I used `beta = 1` explicitly because I got compile error if I try to use `NoBetaScaling` for many epilogues. May be that was fixed in the PR.

hwu36:
NVIDIA GPU lowers float pointer division into a function call which has non trivial control flows and many instructions.  Try to avoid it if possible.

In this PR, I changed fp32 hardswish to use 5 instructions per element and fp16 hardswish to use 2.5 instructions per elements (on ampere only, early ARCH needs a little more instructions) with no control flow and no comparison.

include/cutlass/epilogue/thread/activation.h
```diff
@@ -193,7 +192,20 @@ struct HardSwish {
     minimum<T> mn;
     maximum<T> mx;
     T relu6 = mn(mx(x + T(3), T(0)), T(6));
-    return x * (relu6 / T(6));
+    return x * relu6 / T(6);
```
antinucleon, line 195: return x * relu6 * T(0.16666667);

include/cutlass/functional.h
```diff
@@ -1536,6 +1535,224 @@ struct multiply_add<Array<half_t, N>, Array<half_t, N>, Array<half_t, N>> {
...
+    Array<half_t, N> result;
+    #if defined(__CUDA_ARCH__) && (__CUDA_ARCH__ >= 530)
+
+    __half2 *result_ptr = reinterpret_cast<__half2 *>(&result);
+    __half2 const *lhs_ptr = reinterpret_cast<__half2 const *>(&lhs);
+    __half2 const *rhs_ptr = reinterpret_cast<__half2 const *>(&rhs);
+
+    CUTLASS_PRAGMA_UNROLL
+    for (int i = 0; i < N / 2; ++i) {
+      result_ptr[i] = __hmax2(lhs_ptr[i], rhs_ptr[i]);
```
antinucleon, line 1660: ```__hmax2``` only works with ```__CUDA_ARCH__ >= 800```
should be like:
```
#if __CUDA_ARCH__ >= 800
result_ptr[i] = __hmax2(lhs_ptr[i], rhs_ptr[i]);
#else
result_ptr[i].x = lhs_ptr[i].x > rhs_ptr[i].x ? lhs_ptr[i].x :  rhs_ptr[i].x;
result_ptr[i].y = lhs_ptr[i].y > rhs_ptr[i].y ? lhs_ptr[i].y :  rhs_ptr[i].y;
#end
```

include/cutlass/functional.h
```diff
@@ -1536,6 +1535,224 @@ struct multiply_add<Array<half_t, N>, Array<half_t, N>, Array<half_t, N>> {
...
+    }
+
+    if (N % 2) {
+      __half const *a_residual_ptr = reinterpret_cast<__half const *>(&lhs);
+      __half const *b_residual_ptr = reinterpret_cast<__half const *>(&rhs);
+
+      __half d_residual = __hmax(
```
antinucleon, line 1667: Same ```__hmax``` only works with ```__CUDA_ARCH__ >= 800```

antinucleon:
@hwu36 I have updated ```__hmax``` but I think ```__hmin``` will have similar problem, which I don't fix yet.

include/cutlass/epilogue/thread/activation.h
hwu36: Base class still needs to use `/ 6`.  Not all data types supports `* 0.167`.

antinucleon:
@hwu36 I tested e2d with relu0. The mix-precision has 6% perf regression. For pure fp16 it is 0.1% faster. So I don't think it is important to keep relu0 in the codebase.

hwu36:
@antinucleon , have you tried int8 input, int32 accumulate, fp32 epilogue, int8 output tensor core gemm/conv kernels on sm86?  relu0 should be helpful, too.

@masahi , what is your conclusion of relu0?

antinucleon:
> 

I haven't written any code for int8 yet so I can't test. From fp16 data point I don't think it is super helpful.

masahi:
I haven't looked at relu0 in detail yet, but relu is always used with 0 as the threshold and I never use serial split-k. So it is fine from my end.

hwu36:
I will merge it as it is.  I understand Relu0 is not very useful, but I will keep it in the case it is useful for integer cases.  I will just not advocate Relu0.

hwu36 (approved):

PR #407, opened 2022-02-14
Actually use float accumulation in gemm_f16t_f16t_f16t_wmma_tensor_op…

jspark1105:
…_f32_sm70.cu

gemm_f16t_f16n_f16t_wmma_tensor_op_f32_sm70.cu also needs to be changed similarly


hwu36:
Thanks for catching this.

I am testing internally now.

hwu36 (approved):

PR #419, opened 2022-03-07
numeric_limit for uint1b_t

yanghaojin:
The type of max() of uint1b_t is assigned to uint4b_t, is that a bug?


hwu36:
looks like a bug.  i will test it internally first.  thank you.

hwu36 (approved):

PR #441, opened 2022-03-24
[BUGFIX]: falsely unroll loop

Peter9606:
Force unroll a loop that doesn't have compilation constant loop times is dangerous.

I cannot reproduce the failure using NVCC, but it did 100% crash using our clang tools while running examples/02_dump_reg_shmem.


d-k-b:
Thanks @Peter9606. This fix seems reasonable, but it also sounds like a bug on clang's part. The compiler should either error out or handle that gracefully with a partial unroll and predication.

Peter9606:
Thanks for your kindly reminding @d-k-b.  I agree with you that the compiler should be able handle such case.

hwu36 (approved):

PR #442, opened 2022-03-24
Added missing public qualifier to enable GemmUniversal with row-major output

kerrmudgeon:
Added missing public qualifier to enable GemmUniversal with row-major output.


hwu36:
This one is already fixed in 2.9

PR #452, opened 2022-04-05
Use platform:: instead of std::abs and std::conditional

kroburg:
There are two types of errors during threadblock gemm compilation using nvrtc:

```
cutlass/include/cutlass/conv/conv2d_problem_size.h(502): error: namespace "std" has no member "abs"

cutlass/include/cutlass/epilogue/threadblock/default_epilogue_tensor_op.h(324): error: namespace "std" has no member "conditional"
```


hwu36:
yeah, we should always use `platform::`.  Want to file a PR?

hwu36:
Hi, @kroburg ,

Thank you for your PR and sorry for the late response.

We fixed most of the `std` problem you mentioned in 2.9 release, but we'd still like to merge your change in `platform.h`.

I tried to do it myself, but somehow I am not allowed to change your branch.  Would you please merge with 2.9 and push again.  Thank you very much.

hwu36:
Hi, @kroburg 

We still want to keep `absolute_value` in `conv2d_problem_size.h` and `fast_math.h` because `absolute_value` can support  data types that are not `int32_t` and `int64_t`.  std or platform `abs` should only support integers.

We want to keep your change in `platform.h` though.

hwu36 (approved):

Issue #310, opened 2021-08-16
automatically add int128 flag when NVC++ is used


jeffhammond:
Support for 128-bit integers is a recent addition to the NVIDIA C++ compiler (https://docs.nvidia.com/hpc-sdk/compilers/hpc-compilers-ref-guide/#data-typesc-cpp-type-keywords), which means that it is now possible to build CUTLASS files that use these.

I wonder if CUTLASS could add the CMake convenience that automatically employs the `-Mint128` flag when the NVC++ compiler is used.  It's not hard to do it manually, but only if one is sufficiently experienced with NVC++ to know that this feature has been added and that there is a special flag for it.

This works today:
```
cmake .. -DCMAKE_CXX_COMPILER=nvc++ -DCMAKE_CXX_FLAGS="-Mint128"
```

My hope is that the second argument is not necessarily in the future.

hwu36:
@kerrmudgeon and @d-k-b , thoughts about how to support `nvc++` ?

mnicely:
@jeffhammond are you building CUTLASS with nvc++?

jeffhammond:
I build everything I can with NVHPC compilers, for obvious reasons 😃

jeffhammond:
This is trivial to fix. Please do it.

PR #473, opened 2022-04-26
Update CMakeLists.txt

hwu36:
Add 128bit int support if using nvc++ to solve #310 

@jeffhammond, would you please give it a try?


jeffhammond:
Looks good to me

PR #480, opened 2022-04-28
2.9 fixes for nvrtc

kroburg:
Hello.

There are two problems related to nvrtc in recent release:
I Using `std::is_same` instead of `platform::is_same`
II `#include <cuComplex.h>` was [removed](https://github.com/NVIDIA/cutlass/commit/12f4108ac2233022b92f0e3533c23bed399fcf45#diff-e61f0e1523fa79466124610cca6dbdb83da572fed0944c5e48af8814898d9059L27) from `cutlass/complex.h` includes for nvrtc.

I For me it is not clear what is the cutlass strategy about nvrtc.
1. Always use `platform::`.

2. Use cuda/std/some_std headers for nvrtc.
Looks ugly:
```
#if defined(__CUDACC_RTC__)
#include <cuda/std/cstdint>
#else
#include <cstdint>
#endif
```

Also [CUB](https://github.com/NVIDIA/cub) will be broken anyway, so one should use "safe headers" from something like [jitify](https://github.com/NVIDIA/jitify).
But safe headers from jitify are not compatible with cuda/std from CUDA distribution.
So one should put jitify header `some_std` instead of `cuda/std/some_std` and cross fingers.

3. Generate stubs for `#include <some_std>`:
```
#pragma once
#include <cuda/std/some_std>
```
This way still does not fix CUB because it requires some stuff like `stdlib.h` which is not a part of cuda/std.

Any other?

### DIFF
--- a/include/cutlass/platform/platform.h
+++ b/include/cutlass/platform/platform.h
@@ -416,11 +416,11 @@ using std::remove_cv;
 
 #if defined(__CUDACC_RTC__) || (!defined(_MSC_VER) && (__cplusplus < 201103L)) || (defined(_MSC_VER) && (_MSC_VER < 1500))
 
-/// platform::is_same (false specialization)
+/// std::is_same (false specialization)
 template <typename A, typename B>
 struct is_same : false_type {};
 
-/// platform::is_same (true specialization)
+/// std::is_same (true specialization)
 template <typename A>
 struct is_same<A, A> : true_type {};
 
--- a/include/cutlass/platform/platform.h
+++ b/include/cutlass/platform/platform.h
@@ -454,7 +454,7 @@ struct is_base_of
                                      typename remove_cv<DerivedT>::type>::value)> {};
 #else
 
-using platform::is_same;
+using std::is_same;
 using std::is_base_of;
 
 #endif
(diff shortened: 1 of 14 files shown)


hwu36:
We should always use `platform::` instead of `std::`.  Sorry about that.

include/cutlass/platform/platform.h
```diff
@@ -416,11 +416,11 @@ using std::remove_cv;
 
 #if defined(__CUDACC_RTC__) || (!defined(_MSC_VER) && (__cplusplus < 201103L)) || (defined(_MSC_VER) && (_MSC_VER < 1500))
 
-/// std::is_same (false specialization)
+/// platform::is_same (false specialization)
```
hwu36, line 419: we don't need to change this comment.

include/cutlass/platform/platform.h
```diff
@@ -416,11 +416,11 @@ using std::remove_cv;
...
 template <typename A, typename B>
 struct is_same : false_type {};
 
-/// std::is_same (true specialization)
+/// platform::is_same (true specialization)
```
hwu36, line 423: ditto

include/cutlass/platform/platform.h
kroburg: Ouch. Too much sed.

hwu36 (approved):

PR #562, opened 2022-07-09
fix out of bound write bug in stride dgrad

hwu36:
This commit fixes a functional bug.  It let threadblock early exit if start_h >= H or start_w >= W since they are out of bound.  It only happens when stride_h > H or stride_w > W

@manishucsd @masahi


manishucsd:
Thanks for fixing this. LGTM.

hwu36:
Please approve it.

manishucsd (approved):

manishucsd:
We should also consider adding the failing problem sizes in the [unit tests](https://github.com/NVIDIA/cutlass/blob/master/test/unit/conv/device/conv2d_problems.h#L236)

hwu36:
Please help me merge. @kerrmudgeon 

Unit test will come with 2.10

kerrmudgeon (approved):

PR #563, opened 2022-07-11
Remove redundant <fstream> includes

dfyz:
In two different code fragments, CUTLASS includes `<fstream>` but doesn't actually use any identifiers from the header. This PR removes these redundant includes.

This seems like a gratuitous change, but it does have a nice side-effect of making CUTLASS compile in one of our legacy configurations (with broken `<fstream>` that doesn't compile in device code).


hwu36:
Hi @dfyz , we tested your change.  It turns out that we need to add more `#include <fstream>` to other example and unit test files.  Would you please call `make all` and add the missing ones?

dfyz:
Whoops, sorry, my bad. A lot of tests/examples apparently needed the implicit `<fstream>` include from `tools/util/include/cutlass/util/distribution.h`. We only build a small subset of examples relevant to us internally, and none of them has an option to save workspace (for which `<fstream>` is needed), so I didn't notice the (quite substantial) build breakage.

I went through the code in `{test,examples}` and hopefully fixed everything. Now:
  1. `make all` succeeds on my machine.
  2. `<fstream>` is now present in a file if and only if the file directly uses file operations (i.e., `std::{i,o}fstream`), so that there are no implicit dependencies.

Could you please give it another go?

test/unit/epilogue/warp/fragment_iterator_volta_tensor_op.cu
```diff
@@ -32,6 +32,8 @@
     \brief Unit tests for thread-level GEMM
 */
 
+#include <fstream>
```
jackkosaian, line 35: This include is not necessary, as `ofstream` is commented out below

dfyz: Indeed, thanks.

hwu36 (approved):

PR #590, opened 2022-08-02
b2b gemm residual

hwu36:
Add residual support for shmem staging iterator used in back-to-back GEMM fusion. This allows support of problem_size_0_n that is not multiple of 32.

@danthe3rd , would you please give it a try?

### DIFF
--- a/include/cutlass/gemm/warp/mma_tensor_op_tile_access_iterator.h
+++ b/include/cutlass/gemm/warp/mma_tensor_op_tile_access_iterator.h
@@ -1,0 +1,362 @@
+/***************************************************************************************************
+ * Copyright (c) 2017 - 2022 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: BSD-3-Clause
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ * list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ *
+ * 3. Neither the name of the copyright holder nor the names of its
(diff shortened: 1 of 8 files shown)


danthe3rd:
Looking quickly through the code it looks great! Thanks a lot for putting that up so fast :)
Unfortunately, I'll be away for the entire month of August. I can give it a try in September, I'll also try to find someone else to do it in my absence - but in the meantime feel free to merge. If it passes the tests for values of `problemSize1.K` like `[16, 32, 34, 36, 48, 64, 80]`, it should work for my usecase.
Once again thanks a lot :)

danthe3rd:
So I gave it a shot and it's working great! Thanks for putting this together :)
There is just one issue when `problem_size_0_n > 32 && problem_size_0_n % 2 == 1` (for instance `problem_size_0_n = 33`): it triggers an "CUDA Exception: Warp Misaligned Address" upon reading the shared-memory because it's not aligned on 32bits after the initial residual tile. I'm not sure if this is a specific case you want to support tho.

include/cutlass/gemm/warp/mma_tensor_op_tile_access_iterator.h
```diff
@@ -0,0 +1,362 @@
...
+            MatrixCoord offset(
+              access_m_idx * 8 + inst_m_idx * InstructionShape::kRow, 
+              inner_idx * 4 * kElementsPerAccess + iterations_ * InstructionShape::kColumn);
+
+            MatrixCoord access_coord = origin_ + offset;
+
+//            if(access_coord.row() < extent_.row() && access_coord.column() < extent_.column()) {
+
+              access_ptr[access_idx] = *reinterpret_cast<AccessType const *>(
+                ref_.data() + ref_.offset(offset));
```
danthe3rd, line 312: support `problem_size_0_n = 33` would require to change this to 2 accesses, but will also be slower (maybe via template parameter?)

hwu36:
We have no itention to support small alignment in b2b gemm.  It is not a good idea to apply b2b on not well aligned. Inputs.  Padding matices is easy and can immediately bring good performance.

kerrmudgeon (approved):

PR #631, opened 2022-09-19
[DRAFT, do not merge] gemm_universal_with_broadcast, +2 sources.

ipiszy:
This is the original gemm_universal_with_broadcast PR written at April. The added unittest test/unit/gemm/device/gemm_broadcast_test.cu passed at that time. But now it cannot pass any more.


hwu36:
@jackkosaian consolidated v1 and v2 in 2.11. closing

PR #632, opened 2022-09-19
Gemm broadcast 

ipiszy:
* Add a gemm_universal_with_broadcast gemm device level API;
* Have gemm_with_fused_epilogue, epilogue_with_broadcast and linear_combination_residual_block support 2 sources.

Note that the original PR (https://github.com/NVIDIA/cutlass/pull/631) was written at April. In this PR, there are multiple V2 classes as the original PR had several merge conflicts with cutlass 2.8. Unittest in the original PR was also removed in this update.


hwu36:
i will do some sanity check first, and merge it as is. and then merge v1 and v2.

PR #639, opened 2022-09-22
Correct typos in comments

alexfreudenberg:
Correct comments in code on type of generated distribution. Improve Gaussian RNG to take advantage of Box Muller method


tools/util/include/cutlass/util/reference/host/tensor_fill.h
```diff
@@ -205,23 +205,29 @@ struct RandomGaussianFunc<complex<Element> > {
...
+    double u2 = double(std::rand()) / double(RAND_MAX);
 
-      if (int_scale >= 0) {
-        rnd = double(int(rnd * double(1 << int_scale)));
-        reals[i] = from_real<Element>(rnd / double(1 << int_scale));
+    // Compute Gaussian random values
+    double rnd1 = std::sqrt(-2 * std::log(u1)) * std::cos(2 * pi * u2);
+    double rnd2 = std::sqrt(-2 * std::log(u1)) * std::sin(2 * pi * u2);
+    rnd1 = mean + stddev * rnd1;
+    rnd2 = mean + stddev * rnd2;
```
hwu36, line 216: small nit, can we put box-muller algorithm into an inline helper function with documentation. it can be used in both complex and quaternion then.

tools/util/include/cutlass/util/reference/host/tensor_fill.h
```diff
@@ -404,7 +429,7 @@ void TensorFillRandomGaussian(
 }
 
 ///////////////////////////////////////////////////////////////////////////////////////////////////
-/// Fills a tensor with random values with a Gaussian distribution.
+/// Fills a rank-2 tensor symmetrically with random values of a Gaussian distribution.
```
hwu36, line 432: This is filling up a "symmetric rank-2 tensor", not filling up a "rank-2 tensor symmetrically" :). Bcoz for a symmetric matrix we don't fill up the invalid region.

alexfreudenberg:
I tried using better semantics and added an inline function for the Box Muller algorithm

tools/util/include/cutlass/util/reference/host/tensor_fill.h
```diff
@@ -205,23 +211,22 @@ struct RandomGaussianFunc<complex<Element> > {
...
-      double u2 = double(std::rand()) / double(RAND_MAX);
-
-      // Compute Gaussian random value
-      double rnd = std::sqrt(-2 * std::log(u1)) * std::cos(2 * pi * u2);
-      rnd = mean + stddev * rnd;
-
-      if (int_scale >= 0) {
-        rnd = double(int(rnd * double(1 << int_scale)));
-        reals[i] = from_real<Element>(rnd / double(1 << int_scale));
+    double rnd1 = detail::BoxMullerFunc(mean, stddev);
```
hwu36, line 214: they are both using `cos` now.  do we want to call `cos` and `sin` by turns?

tools/util/include/cutlass/util/reference/host/tensor_fill.h
```diff
@@ -64,6 +64,18 @@ namespace host {
 
 namespace detail {
 
+/// Returns a value of the Gaussian distribution generated by the Box Muller method
+__forceinline__
```
hwu36, line 68: we don't need `__forceinline__` here. compiler will do it.  we dont use it explicitly in this file.

tools/util/include/cutlass/util/reference/host/tensor_fill.h
alexfreudenberg: Yes, we're wasting some efficiency here, but I wasn't able to come up with an easy and comprehensible inline function (without a loop) which returns a dynamic number of random values. A fully efficient function would reuse the generated uniform values `u1, u2`, but this would make the function much less readable...

tools/util/include/cutlass/util/reference/host/tensor_fill.h
alexfreudenberg: I'll remove it

tools/util/include/cutlass/util/reference/host/tensor_fill.h
hwu36: i think we only need to change the complex and quantitation ones.  we can leave the real one untouched.  the helper function just need to return a pair of Gaussian random values by calling both cos and sin.

hwu36:
I made some changes to pass our ci/cd.  Would you please test it in your end too?

alexfreudenberg:
Thank you! I wasn't aware that the struct-operator structure is a requirement. The new version also works for me.

hwu36 (approved):

PR #641, opened 2022-09-25
Auxiliary debugging block

alexfreudenberg:
Add an extra block of code which tests if the required size of dynamically allocated shared memory can be implemented. This is helpful for identifying invalid shapes

### DIFF
--- a/include/cutlass/gemm/device/gemm.h
+++ b/include/cutlass/gemm/device/gemm.h
@@ -372,7 +372,7 @@ public:
       return Status::kErrorMemoryAllocation;
     }
 
-    int smem_capacity = static_cast<int>(properties.sharedMemPerMultiprocessor);
+    int smem_capacity = static_cast<int>(properties.sharedMemPerBlockOptin);
     if (smem_size > smem_capacity) {
       return Status::kErrorMemoryAllocation;
     }
--- a/include/cutlass/gemm/device/gemm.h
+++ b/include/cutlass/gemm/device/gemm.h
@@ -480,6 +480,15 @@ public:
     cudaError_t result;
 
     int smem_size = int(sizeof(typename GemmKernel::SharedStorage));
+    
+    // MaxDynamicSharedMemorySize must be smaller than the device attribute cudaDevAttrMaxSharedMemoryPerBlockOptin minus the function attribute sharedSizeBytes  
+    cudaFuncAttributes attr;
+    cudaFuncGetAttributes(&attr, Kernel<GemmKernel>);
+    int max_mem = 0;
+    cudaDeviceGetAttribute(&max_mem, cudaDevAttrMaxSharedMemoryPerBlockOptin, 0);
+    if (smem_size > (int) (max_mem - attr.sharedSizeBytes)) {
+      return Status::kErrorMemoryAllocation;
+    }
 
     if (smem_size >= (48 << 10)) {
       result = cudaFuncSetAttribute(Kernel<GemmKernel>,
--- a/include/cutlass/gemm/device/gemm.h
+++ b/include/cutlass/gemm/device/gemm.h
@@ -365,11 +365,15 @@ public:
     
     // MaxDynamicSharedMemorySize must be smaller than the device attribute cudaDevAttrMaxSharedMemoryPerBlockOptin minus the function attribute sharedSizeBytes  
     int smem_size = int(sizeof(typename GemmKernel::SharedStorage));
-    cudaFuncAttributes attr;
-    cudaFuncGetAttributes(&attr, Kernel<GemmKernel>);
-    int max_mem = 0;
-    cudaDeviceGetAttribute(&max_mem, cudaDevAttrMaxSharedMemoryPerBlockOptin, 0);
-    if (smem_size > (int) (max_mem - attr.sharedSizeBytes)) {
+    cudaDeviceProp properties;
+    result = cudaGetDeviceProperties(&properties, device_idx);
+
+    if (result != cudaSuccess) {
+      return Status::kErrorMemoryAllocation;
+    }
+
+    smem_capacity = static_cast<int>(properties.sharedMemPerMultiprocessor);
+    if (smem_size > smem_capacity) {
       return Status::kErrorMemoryAllocation;
     }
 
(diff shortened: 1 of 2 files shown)


include/cutlass/gemm/device/gemm.h
```diff
@@ -480,6 +480,15 @@ class Gemm {
     cudaError_t result;
 
     int smem_size = int(sizeof(typename GemmKernel::SharedStorage));
+
```
hwu36, line 483: the right place to do this check is in `can_implement()` above.

Also, `device::gemm_universal` is more useful than `device::gemm` now.  the former can do batch gemm, splitk in the same kernel.  `gemm_universal` can have the same smem size check.

alexfreudenberg: Thanks, I moved the block to the `can_implement` function. I wasn't aware of the `gemm_universal` template, but it seems to already have some memory checks in the `maximum_active_blocks` function. Do you think the additional checks should be added anyway?

include/cutlass/gemm/device/gemm.h
```diff
@@ -362,7 +362,18 @@ class Gemm {
     if (!kSplitKSerial && args.split_k_slices > 1) {
       return Status::kErrorInvalidProblem;
     }
+    
+    // MaxDynamicSharedMemorySize must be smaller than the device attribute cudaDevAttrMaxSharedMemoryPerBlockOptin minus the function attribute sharedSizeBytes  
+    int smem_size = int(sizeof(typename GemmKernel::SharedStorage));
+    cudaFuncAttributes attr;
+    cudaFuncGetAttributes(&attr, Kernel<GemmKernel>);
+    int max_mem = 0;
+    cudaDeviceGetAttribute(&max_mem, cudaDevAttrMaxSharedMemoryPerBlockOptin, 0);
```
hwu36, line 371: we can change the logic here that only checking when `smem_size` is larger than 48KB.  then we dont need to get `attr`.

hwu36: Instead, we can also use 

```
        cudaDeviceProp properties;
        result = cudaGetDeviceProperties(&properties, device_idx);

        if (result != cudaSuccess) {
          return -1;
        }

        smem_capacity = static_cast<int>(properties.sharedMemPerMultiprocessor);
```

include/cutlass/gemm/device/gemm.h
hwu36: maximum_active_blocks is used to check the occupancy.  it does not decide if something can run or not.  if the allocated shared mem is too big, it just returns 0.

include/cutlass/gemm/device/gemm.h
alexfreudenberg: > we can change the logic here that only checking when `smem_size` is larger than 48KB. then we dont need to get `attr`.
Wouldn't this limit the capacity of newer architectures with more shared memory available?

alexfreudenberg: Is the `sharedMemPerMultiprocessor` property equivalent to the original query? I can't infer this from the documentation

hwu36: > Wouldn't this limit the capacity of newer architectures with more shared memory available?

48KB is the limit of static allocated memory.  No plan to change.

> Is the sharedMemPerMultiprocessor property equivalent to the original query?

Feel free to try it first.

mnicely:
@alexfreudenberg I just wanted to follow-up on the status of this PR. Do you need anything from us to get unblocked?

alexfreudenberg:
@mnicely I'm very sorry, I think this got mixed up with the other PR in my inbox. I will try to implement the solution proposed by Haicheng.

mnicely:
No worries, thanks for following up

include/cutlass/gemm/device/gemm.h
```diff
@@ -362,7 +362,22 @@ class Gemm {
...
+    // MaxDynamicSharedMemorySize must be smaller than the device attribute cudaDevAttrMaxSharedMemoryPerBlockOptin minus the function attribute sharedSizeBytes  
+    int smem_size = static_cast<int>(sizeof(typename GemmKernel::SharedStorage));
+    cudaDeviceProp properties;
+    result = cudaGetDeviceProperties(&properties, device_idx);
+
+    if (result != cudaSuccess) {
+      return Status::kErrorMemoryAllocation;
+    }
+
+    int smem_capacity = static_cast<int>(properties.sharedMemPerMultiprocessor);
```
hwu36, line 375: we need to use `sharedMemPerBlockOptin` to get the max usable shared memory.

hwu36 (changes requested):

include/cutlass/gemm/device/gemm.h
alexfreudenberg: @hwu36 Thanks for reviewing this. I can't find the `sharedMemPerBlockOptin` attribute in the device properties, they only offer the more general `sharedMemPerBlock`. 
Would it make sense to go back to my original suggestion of getting the attributes through `cudaDeviceGetAttribute` and `cudaFuncGetAttributes`?  At least this is what I understood from the documentation of the `cudaFuncSetAttribute` function.

mnicely: @alexfreudenberg https://docs.nvidia.com/cuda/cuda-runtime-api/structcudaDeviceProp.html#structcudaDeviceProp_1c23eacaeeb59e1798d6d442be70f1c3e

include/cutlass/gemm/device/gemm_universal_base.h
```diff
@@ -262,7 +276,7 @@ class GemmUniversalBase {
           return -1;
         }
 
-        smem_capacity = static_cast<int>(properties.sharedMemPerMultiprocessor);
+        smem_capacity = static_cast<int>(properties.sharedMemPerBlockOptin);
```
hwu36, line 279: thank you for the change.  I think you would need a rebase with the latest 2.11.  this line is already changed in tot.

alexfreudenberg: Thanks! I rebased both versions and also updated the device identifier, which was undetermined in the last commits

hwu36:
thanks for the rebasing, but i noticed even more changes that do not caused by you.  Maybe, just close this one and create a new PR?

alexfreudenberg:
Alright, will do. Thanks!!

Issue #644, opened 2022-09-26
[QST] Why keeping `col_permute_` and `row_permute_` as class member in `permute.h`


Enter-tainer:
**What is your question?**

`col_permute_` and `row_permute_` can be uniquely determined by the input coordinate. They are private, and not being read by any other member functions. It seems that there is no need to store them in the layout class. I think they should be local variables.

https://github.com/NVIDIA/cutlass/blob/master/include/cutlass/layout/permute.h#L75-L76

Is there any special concern for keeping them?

hwu36:
You are correct.  They can be local variables.  PR is very welcome.

PR #646, opened 2022-09-27
refactor(permute): change unused class member to local var

Enter-tainer:
`col_permute_` and `row_permute_` can be uniquely determined by the input coordinate. They are private, and not being read by any other member functions. So they can be local variables.

See more at #644

### DIFF
--- a/include/cutlass/layout/permute.h
+++ b/include/cutlass/layout/permute.h
@@ -119,9 +119,6 @@ private:
 
   Index stride_permute_;
   
-  Index col_permute_;
-  Index row_permute_;
-
 public:
   //
   // Methods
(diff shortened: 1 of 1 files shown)


include/cutlass/layout/permute.h
```diff
@@ -159,10 +156,10 @@ class Tensor4DPermute0213 {
     int i = row_init / D1;
 
     // After the Permute Op
-    col_permute_ = l + j * D3;
-    row_permute_ = k + i * D2;
+    Index col_permute = l + j * D3;
+    Index row_permute = k + i * D2;
 
-    return LongIndex(row_permute_) * LongIndex(stride_permute_) + LongIndex(col_permute_);
+    return LongIndex(row_permute) * LongIndex(stride_permute_) + LongIndex(col_permute);
```
Enter-tainer, line 162: Question here: given that we know `[D0, D1, D2, D3]`, and the corresponding coordinate `[i, j, k, l]`, can we compute the final index using something like $l + D_3k + jD_2D_3j + iD_1D_2D_3$ ?

hwu36 (approved):

PR #658, opened 2022-10-10
ensure singleton::get thread safe construct instance

Xseventh:


hwu36:
Sorry for the late response.  Would you please explain in detail what goes wrong and how you fixed it?  Thank you.

Xseventh:
> Sorry for the late response. Would you please explain in detail what goes wrong and how you fixed it? Thank you.

I want to use cutlass library for write custom tensorflow op, when I learning cutlass library code, I see this singleton is not thread safe, when i call singleton::get() in custom op, it may cause data race and call multiple manifest.initialize() in theory. 

I think use static variable is the best practice for c++ singleton implementation, we are need not be care about multithread data race.

kerrmudgeon (approved):

tools/library/src/singleton.cu
```diff
@@ -54,10 +50,8 @@ Singleton::Singleton() {
 }
 
 Singleton const & Singleton::get() {
-  if (!instance.get()) {
-    instance.reset(new Singleton);
-  }
-  return *instance.get();
+  static Singleton instance;
+  return &instance;
```
hwu36, line 54: `error: no suitable constructor exists to convert from "cutlass::library::Singleton *" to "cutlass::library::Singleton"`

Xseventh: sorry this is a mistake, I will fix it

PR #670, opened 2022-10-21
Example 43 - DualGemm

danthe3rd:
Implement dual-GEMM, which executes this:
```
D0 = epilogue0(X @ B0, C0) # GEMM0
D1 = epilogue1(X @ B1, C1) # GEMM1
D2 = epilogue2(D0, D1)
```

Benefits:
- The GEMM part (GEMM0 & GEMM1) is faster because we load X only once
- The following operation no longer requires to re-load D0/D1 from gmem

Options supported:
- SplitKSerial
- Bias or no bias
- Storing (or not) D0/D1 in global memory (NOTE: storing D0/D1 in gmem is required for `SplitKSerial=1`)

Limitations:
- Only supports MmaMultistage (so only A100+)
- Only efficient for f16/bf16 operands (accumulation can be either f16 or f32)
- Only supports sharing the same operand A (eg does not work for `D0 = A0 @ B; D1 = A1 @ B`)

Implementation details:
- This adds a new `threadblock::DualMmaMultistage` which is basically the same as `MmaMultistage`, except that we load 3 matrices
- I also needed to add a `threadblock::DualEpilogue` to process 2 sources/3 outputs at the same time.

Performance:
- ~5% faster
- Can be even faster in some cases (eg SplitKSerial when gmem is used even more)


examples/43_dual_gemm/device/dual_gemm.h
```diff
@@ -0,0 +1,457 @@
...
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ **************************************************************************************************/
+/*! \file
+    \brief Performs a dual gemm:
+```
+D0 = epilogue0(X @ B0, C0)
+D1 = epilogue0(X @ B1, C1)
```
hwu36, line 35: epilogue1 here?

the comment is also a ilttle different from dual_gemm.cu which is

```
D0 = epilogue0(X @ B0, C0)
D1 = epilogue1(X @ B1, C1)
D2 = epilogue2(D0, D1)
```

danthe3rd: You are correct, this should be `D1 = epilogue1(X @ B1, C1)`

hwu36 (approved):

PR #662, opened 2022-10-12
ex42: Fused MHA imported from xFormers

danthe3rd:
Forward only for now.
Currently supports (and tested) on: A100 (f16/f32/bf16), V100 (f16/f32), P100 (f16/f32).

<details>
<summary>Main modifications and customizations to cutlass</summary>

We found ourselves extending cutlass's classes on several occasions:
(1) We subclassed the `threadblock::Mma` [classes](https://github.com/facebookresearch/xformers/tree/main/xformers/components/attention/csrc/cuda/mem_eff_attention/gemm) - mainly to support [doing the prologue early](https://github.com/facebookresearch/xformers/blob/fdbf3eb663c68ea95c965e6e2779e776b84e2db5/xformers/components/attention/csrc/cuda/mem_eff_attention/gemm/custom_mma_multistage.h#L256) (so when we start our second matmul, data is already loaded in memory and we don't have to wait for gmem)
(2) We took the classes from the fused gemm example, and [adapted them to work for V100/SIMT](https://github.com/facebookresearch/xformers/blob/main/xformers/components/attention/csrc/cuda/mem_eff_attention/mma_from_smem.h)
(3) We also had to [change how iterators handle residuals](https://github.com/facebookresearch/xformers/pull/393), to handle arbitrary sequence lengths for Q and K - 
(4) We added a [`prefetch` method](https://github.com/facebookresearch/xformers/blob/fdbf3eb663c68ea95c965e6e2779e776b84e2db5/xformers/components/attention/csrc/cuda/mem_eff_attention/iterators/epilogue_predicated_tile_iterator.h#L282) to the epilogue's tile iterator
(5) Finally, we also modified [the `threadblock::Epilogue` class](https://github.com/facebookresearch/xformers/blob/fdbf3eb663c68ea95c965e6e2779e776b84e2db5/xformers/components/attention/csrc/cuda/mem_eff_attention/epilogue_pipelined.h#L33-L42) to pipeline memory loads from source (was important for performance in our case as we did many matmul on matrices with small K), support source having a different dtype than the target (we want to accumulate in f32, but the output can be f16), and passing the row id to our custom epilogue
(6) For proper performance in the future for `K <= 32`, we would need to implement some more code, as @tridao [pointed out](https://github.com/NVIDIA/cutlass/issues/599#issuecomment-1250117612)
</details>

<details>
<summary>Image explaining how things work together</summary>

(the A100 version uses `MmaMultistage*` instead of `MmaPipelined*`)

image

</details>

<details>
[...]

PR #691, opened 2022-11-08
Add variable-sequence-length FMHA example

jackkosaian:
This pull request adds a variable-sequence-length FMHA kernel based on the fixed-sequence-length FMHA kernel in example 42 (implemented in #662).

The implementation wraps the FMHA kernels from example 42 with CUTLASS's grouped kernels infrastructure. Additional examples of using grouped kernels can be found in [24_gemm_grouped](https://github.com/NVIDIA/cutlass/blob/master/examples/24_gemm_grouped/gemm_grouped.cu), and [38_syr2k_grouped](https://github.com/NVIDIA/cutlass/blob/master/examples/38_syr2k_grouped/syr2k_grouped.cu), along with additional [documentation](https://github.com/NVIDIA/cutlass/blob/master/media/docs/grouped_scheduler.md).

This PR does not remove the existing fixed-sequence-length FMHA kernel, which remains available as `examples/42_fused_multi_head_attention/fused_multihead_attention_fixed_seqlen.cu`

CC @danthe3rd

### DIFF
--- a/examples/42_fused_multi_head_attention/fused_multihead_attention_variable_seqlen.cu
+++ b/examples/42_fused_multi_head_attention/fused_multihead_attention_variable_seqlen.cu
@@ -1,0 +1,1193 @@
+/***************************************************************************************************
+ * Copyright (c) 2017 - 2022 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: BSD-3-Clause
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ * list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ *
+ * 3. Neither the name of the copyright holdvr nor the names of its
+ * contributors may be used to endorse or promote products derived from
+ * this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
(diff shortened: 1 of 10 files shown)


examples/42_fused_multi_head_attention/fused_multihead_attention_variable_seqlen.cu
```diff
@@ -0,0 +1,1193 @@
...
+    std::cerr << "Aborting execution." << std::endl;
+    return -1;
+  }
+
+  if (options.use_mask) {
+    std::cerr << "--use_mask is not supported at the moment\n";
+    return -2;
+  }
+  if (options.alignment != 1) {
+    std::cerr << "--alignment=1 is the only supported value\n";
```
Laurawly, line 1169: Why is alignment required to be 1 here? Will it effect performance?

jackkosaian: Hi, @Laurawly. 

This restriction was originally in place in the fixed-sequence-length FMHA example that this MR builds atop (see [here](https://github.com/NVIDIA/cutlass/blob/168ea8b0e1f0de5f508fba9489c15a8c1958334c/examples/42_fused_multi_head_attention/fused_multihead_attention.cu#L1040-L1043)). As far as I can tell, this restriction can be eliminated, and I have done so in a recent commit.

Perhaps @danthe3rd can comment as to the need for this restriction in #662, in case I am missing something.

danthe3rd: This is used to control for padding in the sequences, as far as I understand from the code.
This MHA implementation does not support padding, so we only support alignment=1

jackkosaian: Thank you for explaining.

I changed the logic a bit in [this commit](https://github.com/NVIDIA/cutlass/pull/691/commits/241328e2a4ce39ed1ce95ed4fa445fc69460a916) so that sequence length and head size must be multiples of `alignment`. This allows one to support higher alignment without padding.

jackkosaian:
Closing this PR, as the variable-sequence-length implementation has been added to example 41 [here](https://github.com/NVIDIA/cutlass/blob/master/examples/41_fused_multi_head_attention/fused_multihead_attention_variable_seqlen.cu).

PR #698, opened 2022-11-15
add `#pragma once` for header file in example 42

Enter-tainer:
As title


hwu36:
@danthe3rd

danthe3rd (approved):
Thanks a lot! good catch

hwu36 (approved):

PR #701, opened 2022-11-17
clean up layout K

hwu36:
@danthe3rd @antinucleon @terrychenism

please take a look.


hwu36:
add @Laurawly for viz

hwu36:
close this since it is part of 2.11 now.  if it causes problem, let me know.

PR #720, opened 2022-12-04
Fix typos in conv problem sizes

alihassanijr:
## What's the problem?

Noticed a few typos on the equality overloaders in Conv2dProblemSize and Conv3dProblemSize.
The former compares height to the target width, and compares dilation_h twice instead of dilation_h and dilation_w.
The latter's just getting dilations wrong.

Haven't really seen it repeated elsewhere but will keep any eye out and commit if I find any.

### DIFF
--- a/include/cutlass/conv/conv2d_problem_size.h
+++ b/include/cutlass/conv/conv2d_problem_size.h
@@ -221,12 +221,12 @@ public:
   CUTLASS_HOST_DEVICE
   bool operator==(Conv2dProblemSize const &conv) const {
     return (
-      (N == conv.N) && (W == conv.H) && (W == conv.W) && (C == conv.C) &&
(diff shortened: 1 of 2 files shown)


jackkosaian:
Thanks for catching this, @alihassanijr . Could you please also change [this line](https://github.com/NVIDIA/cutlass/blob/master/include/cutlass/conv/conv3d_problem_size.h#L208) from:
```
(stride_d == conv.stride_d) && (stride_h == conv.stride_h) && (stride_w == conv.stride_h) &&
```

to

```
(stride_d == conv.stride_d) && (stride_h == conv.stride_h) && (stride_w == conv.stride_w) &&
```
(i.e., comparing `stride_w` to `conv.stride_w`, rather than to `conv.stride_h`)?

alihassanijr:
Thanks for bringing that up @jackkosaian . Done.

hwu36 (approved):

PR #754, opened 2023-01-07
Enable shared memory intrinsics and ldmatrix PTX on Clang.

Gregory-Meyer:
This PR adds preprocessor checks to enable the shared memory intrinsics `__cvta_generic_to_shared` and `__nvvm_get_smem_pointer`, as well as the `ldmatrix` PTX instructions, on Clang. Preventing these intrinsics from being used is a significant latency regression on Clang.

### DIFF
--- a/include/cute/arch/copy_sm75.hpp
+++ b/include/cute/arch/copy_sm75.hpp
@@ -35,8 +35,23 @@
 #include <cute/arch/copy.hpp>
 
 // Config
-#if (defined(__CUDA_ARCH__) && (__CUDA_ARCH__ >= 750))
-#  define CUTE_ARCH_LDSM_SM75_ENABLED
+#if defined(__clang__) && defined(__CUDA__)
+  // ldmatrix PTX instructions added in Clang 14: https://reviews.llvm.org/D107046
+  // ... but broken until Clang 15:
+  //   * https://reviews.llvm.org/D121666
+  //   * https://reviews.llvm.org/D126846
+  #define CUTE_ARCH_CLANG_SUPPORTS_LDSM_SM75 (__clang_major__ >= 15)
+#endif
+
+#ifdef __CUDACC__
+  // ldmatrix PTX instruction added in CUDA 10.2+
+  #define CUTE_ARCH_NVCC_SUPPORTS_LDSM_SM75 ((__CUDACC_VER_MAJOR__  == 10 && __CUDACC_VER_MINOR__ >= 2) || __CUDACC_VER_MAJOR__ >= 11)
+#endif
+
+#define CUTE_ARCH_LDSM_SM75_SUPPORTED (CUTE_ARCH_NVCC_SUPPORTS_LDSM_SM75 || CUTE_ARCH_CLANG_SUPPORTS_LDSM_SM75)
+
+#if CUTE_ARCH_LDSM_SM75_SUPPORTED && defined(__CUDA_ARCH__) && __CUDA_ARCH__ >= 750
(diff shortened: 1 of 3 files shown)


include/cutlass/arch/memory_sm75.h
```diff
@@ -56,30 +56,57 @@ inline __device__ void ldsm(Array<unsigned, MatrixCount> & D, void const* ptr);
...
-#define CUDA_LDMATRIX_ACTIVATED 1
+  // ldmatrix PTX instructions added in Clang 14: https://reviews.llvm.org/D107046
+  // ... but broken until Clang 15:
+  //   * https://reviews.llvm.org/D121666
+  //   * https://reviews.llvm.org/D126846
+  #define CUTLASS_CLANG_SUPPORTS_LDMATRIX (__clang_major__ >= 15)
 #endif
 
-#define CUDA_LDMATRIX_SUPPORTED 1
+#ifdef __NVCC__
```
hwu36, line 78: nvrtc is not included.

Gregory-Meyer: Replaced with `__CUDACC__` as per [NVCC](https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/#nvcc-identification-macro) and [NVRTC](https://docs.nvidia.com/cuda/nvrtc/index.html#predefined-macros) documentation.

hwu36:
@d-k-b , would you please take a look?

Gregory-Meyer:
This should be compatible with the changes in CUTLASS 3.0.0 to migrate the contents of `cutlass_get_smem_pointer` into `cute::cast_smem_ptr_to_uint` now

thakkarV:
Thanks for your contributions. Are your changed compatible with the ldmatrix PTX in `cute/arch/copy_sm75.hpp` as well?

thakkarV:
Yes. I think that might be fine but let me run this by @hwu36

Gregory-Meyer:
d1b70dc9167cbc2238abb60e8f909bc79c6e223c makes [`cute/arch/copy_sm75.hpp`](../blob/master/include/cute/arch/copy_sm75.hpp) compatible with Clang, as requested

include/cute/arch/copy_sm75.hpp
```diff
@@ -51,7 +66,7 @@ struct SM75_U32x1_LDSM_N
   copy(uint128_t const& smem_src,
        uint32_t& dst)
   {
-#if defined(CUTE_ARCH_LDSM_SM75_ENABLED)
```
thakkarV, line 54: nitpick: this strays away from the CUTLASS 3.0/ cute convention of using `if defined(MACRO_NAME)` -- any reason for going with this approach

Gregory-Meyer: I chose to use `#if MACRO_NAME` instead of `#if defined(MACRO_NAME)` so that users could specify `-DMACRO_NAME=1` or `-DMACRO_NAME=0` to enable and disable these features without making any source code changes. However, as this style goes against the existing idioms of CUTLASS and CuTE, I'll revert these changes.

Gregory-Meyer: Done.

hwu36 (approved):

hwu36:
@Gregory-Meyer , i made some changes to the macro.  do you want to test it again with clang before i merge it?

hwu36:
i will merge this since it will show up in 3.1.  let me if it does not work with clang.

Issue #767, opened 2023-01-11
[QST] INT8 GEMM with FP16 output


gsujankumar:
Hey I have noticed INT8 gemm with the following settings where we have `float` bias and `float` output works okay:

```
using ElementAccumulator = int32_t;                   // <- data type of accumulator
using ElementComputeEpilogue = float;  // <- data type of epilogue operations
using ElementInputA = int8_t;              // <- data type of elements in input matrix A
using ElementInputB = int8_t;              // <- data type of elements in input matrix B
using ElementOutput = float;                        // <- data type of elements in output matrix D
```

But setting the outputs in `cutlass::half` does not? 
```
using ElementAccumulator = int32_t;                   // <- data type of accumulator
using ElementComputeEpilogue = float;  // <- data type of epilogue operations
using ElementInputA = int8_t;              // <- data type of elements in input matrix A
using ElementInputB = int8_t;              // <- data type of elements in input matrix B
using ElementOutput = cutlass::half;                        // <- data type of elements in output matrix D
```

Is there any specific restriction preventing it? How can this be enabled?

hwu36:
please try copy paste https://github.com/NVIDIA/cutlass/blob/master/include/cutlass/epilogue/threadblock/default_epilogue_tensor_op.h#L187-L223 and change float to int.  we need a new specialization for your case.  if it is working, feel free to open a PR.

gsujankumar:
That resolves some errors:

Now I am running into this error:

```
../../include/cutlass/epilogue/threadblock/epilogue.h(179): error: more than one partial specialization matches the template argument list of class "cutlass::epilogue::threadblock::SharedLoadIteratorMixed<cutlass::epilogue::threadblock::OutputTileOptimalThreadMap<cutlass::epilogue::threadblock::OutputTileShape<128, 8, 2, 1, 1>, cutlass::epilogue::threadblock::OutputTileShape<1, 8, 1, 1, 8>, 128, 8, 16>::CompactedThreadMap, int32_t, 32, 16, 8, 8>"
            "cutlass::epilogue::threadblock::SharedLoadIteratorMixed<ThreadMap_, Element_, 32, 16, 8, 8>"
            "cutlass::epilogue::threadblock::SharedLoadIteratorMixed<ThreadMap_, int32_t, 32, OutputSizeBits_, 8, 8>"
```

hwu36:
[...]

PR #768, opened 2023-01-12
update: add partial specialization of int32 to fp16/bf16 IteratorsTen…

PerkzZheng:
…sorOp


mnicely:
@PerkzZheng can you please add context and motivation of this PR?

PerkzZheng:
this is a quick fix for the https://github.com/NVIDIA/cutlass/issues/767. We had similar requests internally, and that might be a reference. CC @hwu36 for vis. 
we can close it if it is no longer needed.

mnicely:
closes #767

hwu36:
this is committed in #812

Issue #771, opened 2023-01-13
[QST] fMHA example o_strideM


tianleiwu:
**What is your question?**
I saw o_strideM of the fMHA example is patched in a few applications like:

https://github.com/AITemplate/cutlass/commit/f434be22a6270f9f000712286f92545ccca045b7

https://github.com/Oneflow-Inc/oneflow/blob/e79298104b56b00f23394523505d09469bb89063/oneflow/user/kernels/fused_multi_head_attention_inference_kernel.cu#L109

and in the original code from xFormers:
https://github.com/facebookresearch/xformers/blob/1191f035e8aa06241850dedf2c20d66f5b3ba575/xformers/csrc/attention/cuda/fmha/kernel_forward.h#L127

Could you check whether the change of o_strideM shall be applied?

hwu36:
@terrychenism and @danthe3rd 

I think ait changed it to save a permutation. Ait and xformers developers can explain it better than I do.  If they think I should make the change, I am happy to do it.

danthe3rd:
Hi,

In CUTLASS, I used the shape `[B, H, M, K]` for the attention inputs, to match the [previous multi_head_attention](https://github.com/NVIDIA/cutlass/tree/v2.10.0/examples/41_multi_head_attention) example.
In xFormers, the attention inputs have shape [B, M, H, K]. This is commonly the case in practice, because it's coming from the output of a linear layer as a `[B, M, H * K]` tensor, so it's more intuitive and might avoid data copies, as the output will likely have to be reshaped to `[B, M, H*K]`.

In both cases, we want the output to match the input shape. Because the output is contiguous, we can deduce the strides for dimension M from the shape, so it's either "K" for the shape `[B, H, M, K]` in CUTLASS, or "K * H" for the shape `[B, M, H, K]` in xformers/AITemplate.

tianleiwu:
@danthe3rd, thanks for the effort. We prefer [B, M, H, K] as well since it could avoid extra transpose for inputs/outputs.

PR #773, opened 2023-01-16
xFormer updates to fMHA FW

danthe3rd:
Note: this changes the input/output formats to BMHK from BHMK (see https://github.com/NVIDIA/cutlass/issues/771). The output will be contiguous (no longer possible to specify custom strides for the outputs)

I plan to open another PR after this one to bring the changes from https://github.com/facebookresearch/xformers/pull/587 once it's merged

(A lot of the changes touch files also used for the BW pass, which will be upstreamed later, but I prefer to keep everything in sync as it's easier)


hwu36:
@terrychenism, i assume the change of `input/output formats to BMHK from BHMK` is okay to you since AIT is already doing it.

@mnicely , who else in addtion to ait uses xformers mha through cutlass?  we need to notice them, too.

This api change is needed for the issue mentioned in #771

terrychenism:
Yeah, thanks for the updates! It's aligned with AIT.

danthe3rd:
@terrychenism as it is right now (as in xFormers), the output will always be contiguous with shape BMHK. Is there a use-case where you need to have custom strides for the output as well?

terrychenism (approved):

examples/41_fused_multi_head_attention/mma_from_smem.h
```diff
@@ -57,8 +57,8 @@
 #include "epilogue_thread_apply_logsumexp.h"
 #include "gemm_kernel_utils.h"
 #include "iterators/make_residual_last.h"
-
-/////////////////////////////////////////////////////////////////////////////////////////////////
+#include "iterators/transpose_warp_iterator.h"
+#include "iterators/warp_iterator_from_smem.h"
```
tianleiwu, line 61: Need add these files to avoid build error.

danthe3rd: Oh my bad! Adding them right now

tianleiwu:
@danthe3rd, there is a line in kernel_forward.h is nice to change.

When an application is compiled with sm from 37 to 80 ( to support as many device as possible), this line will cause build error:
https://github.com/NVIDIA/cutlass/blob/0cc33af50c9667c57d842f92512f4cfc3dcfd6d4/examples/41_fused_multi_head_attention/kernel_forward.h#L924
It prevents us to directly use kernel_forward.h. Users like us have to patch the line.

Proposed change is to comment the line like in pytorch:
https://github.com/pytorch/pytorch/blob/95e3e339a8f896799872fad910fc60ea13d96ead/aten/src/ATen/native/transformers/cuda/mem_eff_attention/kernel_forward.h#L878

danthe3rd:
> @danthe3rd, there is a line in kernel_forward.h is nice to change.

Good catch! I removed that

examples/41_fused_multi_head_attention/fused_multihead_attention_fixed_seqlen.cu
```diff
@@ -839,18 +857,17 @@ public:
...
-      p.v_strideH = p.v_strideM * options.seq_length_kv;
-      p.o_strideH = options.head_size_v * options.seq_length;
-      p.q_strideB = p.q_strideH * options.head_number;
-      p.k_strideB = p.k_strideH * options.head_number;
-      p.v_strideB = p.v_strideH * options.head_number;
-      p.o_strideB = options.head_size_v * options.seq_length * options.head_number;
+      p.q_strideB = p.q_strideM * options.seq_length;
+      p.k_strideB = p.k_strideM * options.seq_length_kv;
+      p.v_strideB = p.v_strideM * options.seq_length_kv;
+      std::cout << "stridesQ: " << p.q_strideB << ", " << p.q_strideM << ", " << p.q_strideH << std::endl;
```
MARD1NO, line 870: Maybe you should delete this debug code?

MARD1NO:
Hey guys, Now we change the format from BHMK to BMHK，and in `check_supported`, we still check `stride_M % kAlignment`，shall we change to check stride_H？

examples/41_fused_multi_head_attention/fused_multihead_attention_fixed_seqlen.cu
danthe3rd: Good point - removed thanks :)

danthe3rd:
> Hey guys, Now we change the format from BHMK to BMHK，and in `check_supported`, we still check `stride_M % kAlignment`，shall we change to check stride_H？ @danthe3rd

We check both `strideM` and `strideH` which I believe is correct

MARD1NO:
> > Hey guys, Now we change the format from BHMK to BMHK，and in `check_supported`, we still check `stride_M % kAlignment`，shall we change to check stride_H？ @danthe3rd
> 
> We check both `strideM` and `strideH` which I believe is correct

sry, I missed these code... :)

hwu36 (approved):

PR #779, opened 2023-01-18
Enable aarch64 support

psaab:


kerrmudgeon:
Thank you for submitting this. We are checking it on our side and will then merge upon passing.

kerrmudgeon (approved):

PR #790, opened 2023-01-27
Extend DualGemm: support batched mode + decouple B0/B1 layouts

aakhundov:
The two extensions are suggested for the `DualGemm` in the `examples/45_dual_gemm` folder:

1. Following the `GemmUniversalMode::kBatched` [implementation](https://github.com/NVIDIA/cutlass/blob/add4ba622f1cdebc145d1df0e9620c3c84c00a52/include/cutlass/gemm/kernel/gemm_universal.h#L481), batched mode is added to the `DualGemm`. `DualGemmMode::kBatched` and `SplitKSerial` are incompatible: `Status::kErrorInvalidProblem` is returned if both are set.

2. The `DualGemm` template currently assumes the same layout, `LayoutB`, for both right operand matrices `B0` and `B1`. This is problematic if the layouts of the two matrices are different. In particular, this may be the case when one of the matrices is row-major, while the other is a (column) vector that has to be broadcasted in column-major with zero stride (e.g., as `{B1.device_data(), 0}`) for the `DualGemm` implementation to be able to process `B0` and `B1` simultaneously. In this PR, `LayoutB0` and `LayoutB1` are decoupled throughout the `DualGemm` code (device, kernel, and mma). The batch strides of `B0` and `B1` are also decoupled to accommodate the column vector `B1` case described above.

The unit tests in `dual_gemm.cu` are extended accordingly.

Plus a bug fix in the MHA kernel.

### DIFF
--- a/examples/41_fused_multi_head_attention/kernel_forward.h
+++ b/examples/41_fused_multi_head_attention/kernel_forward.h
@@ -163,8 +163,6 @@ struct AttentionKernel {
     int32_t num_heads;
 
     CUTLASS_HOST_DEVICE int32_t o_strideM() const {
-      // Note: Note in sync with cutlass' main branch!! Make sure to apply
-      // when updating cutlass.
       return head_dim_value * num_heads;
     }
 
--- a/examples/41_fused_multi_head_attention/kernel_forward.h
+++ b/examples/41_fused_multi_head_attention/kernel_forward.h
@@ -163,7 +163,9 @@ struct AttentionKernel {
     int32_t num_heads;
 
     CUTLASS_HOST_DEVICE int32_t o_strideM() const {
-      return head_dim_value;
+      // Note: Note in sync with cutlass' main branch!! Make sure to apply
+      // when updating cutlass.
+      return head_dim_value * num_heads;
     }
 
     // Moves pointers to what we should process
--- a/examples/41_fused_multi_head_attention/kernel_forward.h
+++ b/examples/41_fused_multi_head_attention/kernel_forward.h
@@ -163,7 +163,7 @@ struct AttentionKernel {
     int32_t num_heads;
 
     CUTLASS_HOST_DEVICE int32_t o_strideM() const {
-      return head_dim_value * num_heads;
+      return head_dim_value;
     }
 
     // Moves pointers to what we should process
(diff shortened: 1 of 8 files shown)


ipiszy:
cc @danthe3rd

examples/41_fused_multi_head_attention/kernel_forward.h
```diff
@@ -163,7 +163,7 @@ struct AttentionKernel {
     int32_t num_heads;
 
     CUTLASS_HOST_DEVICE int32_t o_strideM() const {
-      return head_dim_value;
+      return head_dim_value * num_heads;
```
danthe3rd, line 166: I don't think this is correct, as we produce outputs in BHMK format in the cutlass implementation (vs BMHK in AITemplate - that's why there is a difference).
This will change to BMHK in https://github.com/NVIDIA/cutlass/pull/773 anyway :)

aakhundov:
@hwu36 The unit tests in `dual_gemm.cu` are extended to cover the added `DualGemm` functionality (batched mode, decoupled `B0`/`B1` layout, `B1` broadcasting).

The runtimes of the full `B1` case are comparable to the previous version, so there doesn't seem to be a perf regression. However, we observe worse perf (up to 2x runtime) with broadcasted `B1` in column-major. The perf issue seems to be unrelated to the `DualGemm` and is also observed with the `cutlass::gemm::device::Gemm`.

Just replacing the `Gemm` argument `tensor_B1.device_ref()` at [this line](https://github.com/NVIDIA/cutlass/blob/master/examples/45_dual_gemm/dual_gemm_run.h#L309) with `{tensor_B1.device_data(), 0}` leads to 1.8x--2x longer "gemm 1 time" further in the code. Basically, this replacement corresponds to broadcasting the first column of the matrix `B1` (`B1` is in column-major) instead of using the full matrix `B1`.

We've compared the `ncu` traces of both versions: green is with the full `B1` (`tensor_B1.device_ref()`), blue is with the broadcasted column of `B1` (`{tensor_B1.device_data(), 0}`). The version with the broadcasted `B1` seems to get stuck in the "Stall Long Scoreboard" state, which may suggest long waiting for GMEM load.

image

danthe3rd (approved):
The code looks good to me as long as the tests pass - no performance regression when using `LayoutB0 = LayoutB1`

hwu36 (approved):

PR #796, opened 2023-01-31
Fix type bug in conv2d/gemm with broadcast

ShuaiShao93:
When ElementC is INT8 and EpilogueOutputOp::ElementCompute is float, we should use float for compute


hwu36 (approved):

hwu36:
i added elementvector for the bias vector so that it can be different from element c or element compute.

PR #805, opened 2023-02-06
Remove unused header files in fMHA example

masahi:
Not sure if they are intentionally there, but these files are not used by the example.

@danthe3rd @hwu36


danthe3rd:
These headers will be used for the BW pass. We can remove them but I'll add them again later

masahi:
I see, no need to remove and add back.

PR #828, opened 2023-02-16
fMHA: Sync FW with xFormers

danthe3rd:
**Refactor:**
* Remove `scaling_coefs_updater` and split it into (1) A helper class to iterate on GEMM accumulators and (2) A function that does the iterative softmax
* Moved files around in subfolders

**Added:** [1]
* Support for an attention bias (thanks @jfc4050)
* Support for dropout (thanks @jfc4050) - requires pytorch at the moment, but 99% of the work is there
* Support for a shifted causal mask - useful for doing prefixLM in NLP for instance, see illustration
image
* Support for shifting the causal bias by a custom offset for each batch

[1] (all of this is tested in xFormers, and now available in the CUTLASS example with `kernel_forward.h`, but not exposed through the example CLI binary)

**Improved:**
* Improved performance when there is only a single query AND a single head for K/V. This is useful for LLM decoding with multiquery models (eg multiple heads for Q, but sharing a single head for K/V), as it increases tensorcore utilization

**API breaking changes when using `kernel_forward.h`** (see changes in [fused_multihead_attention_fixed_seqlen.cu](https://github.com/NVIDIA/cutlass/compare/master...danthe3rd:cutlass:fmha_fw2?expand=1#diff-e82b6e1f0e4df2881fdb4ba0a15d69f5e562906e032409d7e72ec47c74bac155))
* `causal` boolean has been replaced with `custom_mask_type`
* `o_strideM` needs to be specified
* If you don't use dropout/attention bias, I recommend setting `kSupportsDropout = false` and `kSupportsBias = false` in the template arguments for optimal performance

cc @hwu36 @mnicely @terrychenism @tianleiwu @MARD1NO @fmassa


mnicely:
@danthe3rd 
> Improved performance when there is only a single query AND a single head for K/V. This is useful for LLM decoding with multiquery models (eg multiple heads for Q, but sharing a single head for K/V), as it increases tensorcore utilization

Are there any performance results to add to PR?

hwu36:
@Laurawly

danthe3rd:
> Are there any performance results to add to the PR?

When using 16 heads, I expect the kernel to be roughly 16 times faster. Instead of doing 16xGEMM{M=1, N, K} we do 1xGEMM{M=16, N, K}. In practice, we saw ~10-15% more throughput for decoding jobs end-to-end.

examples/41_fused_multi_head_attention/gemm_kernel_utils.h
```diff
@@ -141,6 +141,37 @@ constexpr CUTLASS_HOST_DEVICE integer align_up(integer n, integer m) {
   return ((n + m - 1) / m) * m;
 }
 
+inline int32_t getMaximumSharedMemoryPerBlockKb(int cc) {
```
hwu36, line 144: @jackkosaian , there is a cuda api call that can get these numbers, correct?

hwu36: i think we can query device property and get `sharedMemPerBlockOptin`

hwu36: @danthe3rd , if we can get the value from a cuda call, is it better to call the api instead of looking up a table here?

hwu36: one example usage is https://github.com/NVIDIA/cutlass/blob/master/test/unit/conv/device/conv2d_testbed.h#L211

jackkosaian: Yes, `cudaGetDeviceProperties(properties, ...)` and the `properties.sharedMemPerBlockOptin` should be what is needed.

jackkosaian: @danthe3rd, is `getMaximumSharedMemoryPerBlockKb()` expected to be called before every kernel launch? If so, and if changing to use the CUDA API, it may be beneficial to restructure the code so that `cudaGetDeviceProperties()` is called only once and the `properties.sharedMemPerBlockOptin` property is saved somewhere. This avoids having the API call on the critical path.

mnicely: You might want to use `cudaDeviceGetAttribute` for better performance. See [blog ](https://developer.nvidia.com/blog/cuda-pro-tip-the-fast-way-to-query-device-properties/) for more details

hwu36 (approved):

Issue #783, opened 2023-01-21
[QST] Per-channel scale epilogue 


david-macleod:
**What is your question?**

Is it possible to implement an epilogue which applies the following `y[i,j].float().mul(s[j]).half()`, so essentially a per-channel scale multiplication (with some type casting). I can see examples of per-channel addition and also examples of elemement-wise scaling, but nothing that has per-channel scaling.

If it's possible pointers in the correct direction would be greatly appreciated, thanks!

jackkosaian:
Hi, @david-macleod.

It sounds like the functionality you're looking for can be achieved by using `DefaultConv2dFpropWithBroadcast`. You can use this kernel with the `LinearCombinationBiasElementwise`, which allows one to perform an [arbitrary binary operation](https://github.com/NVIDIA/cutlass/blob/master/include/cutlass/epilogue/thread/linear_combination_bias_elementwise.h#L64) between the result of the convolution and a bias term on a per-channel basis.

An example of using this method for adding a bias term on a per-channel basis is [here](https://github.com/NVIDIA/cutlass/blob/master/test/unit/conv/device/conv2d_fprop_with_broadcast_sm75.cu#L48-L90). For your usecase, assuming that you don't want to use a particular activation function, you could replace `ReLu` [here](https://github.com/NVIDIA/cutlass/blob/master/test/unit/conv/device/conv2d_fprop_with_broadcast_sm75.cu#L65) with `cutlass::epilogue::thread::Identity<float>` (defined [here](https://github.com/NVIDIA/cutlass/blob/master/include/cutlass/epilogue/thread/activation.h#L86-L102)). You could then add an additional template parameter to specialize the `BinaryOp` used by the epilogue to be `multiplies<float>` (defined [here](https://github.com/NVIDIA/cutlass/blob/master/include/cutlass/functional.h#L77-L84)), rather than the default of `plus`. Populating the bias vector with scaling terms should achieve the functionality you're looking for.

For an example of setting the arguments for the resulting kernel, you can refer to the testbed corresponding to the unit test for Conv2d+broadcast [here](https://github.com/NVIDIA/cutlass/blob/master/test/unit/conv/device/conv2d_with_broadcast_testbed.h#L305-L317).

I hope this helps!

david-macleod:
[...]

PR #838, opened 2023-02-24
expose StoreT parameter for potential speed

erees1:
As per the suggestion in this question https://github.com/NVIDIA/cutlass/issues/783#issuecomment-1408687889 I experimented with toggling storing the T matrix off for speed. This gave us 5-6% throughput speed up for our problem size int8-int8-fp16 matmul. Wonder if it would be worth upstreaming exposing this option?


hwu36:
i change the template arguments order to match the other elementwise ops.

PR #841, opened 2023-02-27
ex24[gemm_grouped]: Allow to change layout/dtype

danthe3rd:
Changing all layouts from ColumnMajor to RowMajor was causing compilation errors because we didn't pass the workspace.

### DIFF
--- a/examples/24_gemm_grouped/gemm_grouped.cu
+++ b/examples/24_gemm_grouped/gemm_grouped.cu
@@ -960,8 +960,10 @@ public:
       };
 
       Gemm gemm_op;
+      size_t workspace_size = gemm_op.get_workspace_size(arguments);
+      cutlass::DeviceAllocation<uint8_t> workspace(workspace_size);
 
-      cutlass::Status status = gemm_op.initialize(arguments);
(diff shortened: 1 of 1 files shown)


hwu36:
@jackkosaian

jackkosaian:
Thanks for reporting this, @danthe3rd.

It seems that this is actually an issue in the underlying `device::GemmUniversalBase` class not setting the `workspace` argument to `nullptr` by default [here](https://github.com/NVIDIA/cutlass/blob/92ebbf1dc4612bf838ace6f2e6d262919f0abd63/include/cutlass/gemm/device/gemm_universal_base.h#L309).

Would you mind changing the PR to set that default parameter to `nullptr`? That way, we don't need to call `get_workspace` etc. unnecessarily in the example.

Apart from that, the changes to use `ElementA` and `ElementB` in place of `half_t` look good to me.

jackkosaian:
Thanks, @danthe3rd. This looks good to me.

@hwu36, I ran through our test suite, and this change does not cause anything to break, so it should be good to merge.

hwu36 (approved):

PR #845, opened 2023-02-28
Reduce versbosity in manifest.py

yinghai:
Currently it will print a bunch of "Culled ..." which is spammy.

### DIFF
--- a/tools/library/scripts/manifest.py
+++ b/tools/library/scripts/manifest.py
@@ -14,10 +14,13 @@ from rank_k_operation import *
 from rank_2k_operation import *
 from trmm_operation import *
 from symm_operation import *
-from conv2d_operation import *  
-from conv3d_operation import *  
+from conv2d_operation import *
+from conv3d_operation import *
+import logging
 
 ###################################################################################################
+_LOGGER = logging.getLogger(__name__)
+
 
 class EmitOperationKindLibrary:
   def __init__(self, generated_path, kind, args):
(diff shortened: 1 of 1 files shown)


tools/library/scripts/manifest.py
```diff
@@ -352,17 +355,17 @@ def append(self, operation):
       self.operations[operation.operation_kind][configuration_name].append(operation)
       self.operation_count += 1
     else:
-      print("Culled {} from manifest".format(operation.procedural_name()))
+      _LOGGER.debug("Culled {} from manifest".format(operation.procedural_name()))
```
yinghai, line 358: Only change here. Rest are whitespace normalization.

thakkarV: Hi! The file this outputs to is already intended to be a log file. Is this something that gets in your workflow? We intended these to be logged because Hopper kernels won't generate without 90a as the arch and we wanted to provide info to the user which kernels are not getting generated etc.

thakkarV: Ah I think this would be an issue if someone runs this script 
outside of the CMake build too to create custom manifests. Fair

yinghai: Yeah, we (AITemplate) run this out of CMake so would be good to provide a way to turn this off.

thakkarV: Agreed! Sorry for not foreseeing this usecase

PR #853, opened 2023-03-07
Hide streams and typinfo from nvrtc

kroburg:
Hello.

There are no "complex" c++ headers during nvrtc compilation. Looks like iostream, core_io and trace.h are used only for Debug Print().

So I hide them.

### DIFF
--- a/include/cutlass/gemm/threadblock/threadblock_swizzle_streamk.h
+++ b/include/cutlass/gemm/threadblock/threadblock_swizzle_streamk.h
@@ -43,10 +43,11 @@
 #include "cutlass/conv/conv3d_problem_size.h"
 #include "cutlass/gemm/threadblock/index_remat.h"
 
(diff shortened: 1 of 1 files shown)


include/cutlass/gemm/threadblock/threadblock_swizzle_streamk.h
```diff
@@ -43,10 +43,11 @@
 #include "cutlass/conv/conv3d_problem_size.h"
 #include "cutlass/gemm/threadblock/index_remat.h"
 
+#ifndef __CUDA_ARCH__
```
hwu36, line 46: can we use `#if !defined(__CUDACC_RTC__)` instead?

kroburg: Sure. Please find second commit.

hwu36 (approved):

Issue #711, opened 2022-11-24
[QST] dimension of A matrix in implict GEMM doc


ijpq:
**What is your question?**
is it supposed [NPQ, RSC] dimension of A matrix rather [NHW, RSC]? "A is a row-major matrix of extent NHW-by-RSC containing activations" indicated in https://github.com/NVIDIA/cutlass/blob/master/media/docs/implicit_gemm_convolution.md

hwu36:
you can understand it as npq.  what really happens is that it loads from nhw tensor, but only npq is inbound.

mnicely:
@ijpq is your questions answered?

ijpq:
yes, thanks a lot.

PR #855, opened 2023-03-08
Fix dimension error in implicit_gemm_convolution.md

zihaomu:
In `A` x `B` = C, the dimension is `NPQ-by-RSC` x `RSC-by-K` = `NPQ-by-K`.

After im2col, the `matrix A` should be `[N, P, Q, RSC]`. `P` and `Q` is the output height and width, instead of the input `H` and `W`.


hwu36:
See #711

PR #879, opened 2023-03-19
[layout] Fix AffineRank2ColumnMajor::packed()

sjfeng1999:


hwu36 (approved):

hwu36:
i made a mistake when i implemented affine2.  affine2row and affine2col are the same thing.  we actually just need one affine2 layout.

PR #892, opened 2023-03-26
Stream-K with broadcast

alihassanijr:
Adds Stream-K GEMM with broadcast support to allow residual block epilogue and therefore residual add.

~~Only single source is implemented at this time, and I think it could use a second set of eyes before proceeding to multiple sources.~~

Both single and multi source (two sources) are supported.

## Changes
**gemm**:
  * `device`
    * `GemmUniversalStreamkWithBroadcast`
  * `kernel`
    * `DefaultGemmStreamkWithBroadcast`
    * `GemmStreamkWithFusedEpilogue`

**epilogue**:
  * `threadblock`:
    * `DefaultEpilogueWithBroadcastTensorOp`
    * `EpilogueStreamkWithBroadcast `

## Possible Issues

**Naming convention**: I was unsure so I tried keeping names as consistent with existing kernels as I could, but we can always change those if necessary.

**Example**: There's a new example, 47-a, which is example 47 (stream k) with residual add.
Outputs do not match exactly with reference code if using floats and not small ints.
I think this expected behavior, but don't want to assume.
~~However, there's two issues:~~
1. ~~There is no device reference for GEMM with broadcast. I'm using the basic split-K as reference for now.~~
2. ~~Outputs do not match.~~
  a. ~~Basic split-K with 2 splits does not match Basic split-K with 1 split (my reference).~~
      ~~Mean squared error is 4e-1.~~
  b. ~~Stream-K (new) does not match Basic split-K with 1 split (my reference).~~
      ~~Mean squared error is 3e-3.~~


hwu36:
@jackkosaian , could you please take a look first?

hwu36:
@alihassanijr , could you please use real float instead of small int in the example and compare the result of not using streamk? the base line can still be `reference::device::gemm` with a for loop to do the customized epilogue.  the goal is to prove no numeric issues.

alihassanijr:
@hwu36 Absolutely, I'll get started on that.

alihassanijr:
Okay so I brought back the reference kernel, and added a host-side for loop to mimic the residual add.
Now it's showing a higher MSE, but it's higher for both basic split-K and stream-K which has me thinking whether or not my for loop is incorrect. I even tried zeroing out both Z and T, error drops but it's still ~ 2e-1 which makes me think something else other than the fused epilogue might be the issue. I'll keep digging.

jackkosaian:
Hi, @alihassanijr. Thank you for contributing this PR.

I'll take a closer look at your implementation soon. First, I wanted to follow up on the following notes from the "Possible Issues" you mentioned.

> a. Basic split-K with 2 splits does not match Basic split-K with 1 split (my reference).
Mean squared error is 4e-1.

What GEMM sizes did you measure this with? Different splitting factors for a split-K parallel GEMM can result in different numerical behavior due to the different floating point roundoff error behavior that will be observed with different splitting factors. For example, the floating point roundoff error that will occur from accumulating all K multiplied elements of two vectors may be different than if one performs two dot products of length K/2 and sums these partial results together.

> b. Stream-K (new) does not match Basic split-K with 1 split (my reference).
Mean squared error is 3e-3.

Does stream K match your reference if you initialize operands with small ints?

examples/47a_ampere_gemm_universal_streamk_residual/ampere_gemm_universal_streamk_residual.cu
```diff
@@ -0,0 +1,703 @@
...
+    options.split_k_factor,                   // batch count / splitk slices
+    {                                         // epilogue parameters
+      ElementAccumulator(options.alpha),
+      ElementAccumulator(options.beta)
+    },
+    tensor_a.device_data(),                   // ptr_A
+    tensor_b.device_data(),                   // ptr_B
+    tensor_c.device_data(),                   // ptr_C
+    tensor_d.device_data(),                   // ptr_D
+    tensor_z.device_data(),                   // ptr_Vector
```
jackkosaian, line 322: Nit: Usually the Z/T terminology is used in the case where we output both the result before and after an elementwise function (e.g., `z = relu(x)`; see [here](https://github.com/NVIDIA/cutlass/blob/77549ae6c8cf31c7ac4c8b88180a8708a8683da4/include/cutlass/epilogue/thread/linear_combination_bias_elementwise.h#L215-L217)). It's probably best to call the tensor pointing to the vector `tensor_Vector`.

examples/47a_ampere_gemm_universal_streamk_residual/ampere_gemm_universal_streamk_residual.cu
```diff
@@ -0,0 +1,703 @@
...
+  }
+
+  TensorMSEFunc<Element, Layout> func(lhs, rhs);
+  cutlass::reference::host::TensorForEach(
+    lhs.extent(),
+    func
+  );
+
+  return func.get_err();
+}
```
jackkosaian, line 445: We could probably use the existing `TensorRelativeErrorMetric` (defined [here](https://github.com/NVIDIA/cutlass/blob/66d9cddc832c1cdc2b30a8755274f7f74640cfe6/tools/util/include/cutlass/util/reference/host/error_metrics.h#L51)) instead. This metric is used in existing CUTLASS examples that test for numerical accuracy (see the 3xTF32 example [here](https://github.com/NVIDIA/cutlass/blob/77549ae6c8cf31c7ac4c8b88180a8708a8683da4/examples/27_ampere_3xtf32_fast_accurate_tensorop_gemm/27_ampere_3xtf32_fast_accurate_tensorop_gemm.cu#L633-L638)).

examples/47a_ampere_gemm_universal_streamk_residual/ampere_gemm_universal_streamk_residual.cu
```diff
@@ -0,0 +1,703 @@
...
+  options.tensor_z.resize(options.problem_size.mn());       // <- Create residual matrix Z with dimensions M x N
+  options.tensor_t.resize(options.problem_size.mn());       // <- Create residual matrix T with dimensions M x N
+
+  // Fill matrix A on host with uniform-random data [4, -4]
+  cutlass::reference::host::TensorFillRandomUniform(
+      options.tensor_a.host_view(),
+      1,
+      ElementA(2),
+      ElementA(-2),
+      8);
```
jackkosaian, line 589: Since this example is initializing floating point values (rather than with small ints), you can remove the `bits` parameter (8, here). See description [here](https://github.com/NVIDIA/cutlass/blob/77549ae6c8cf31c7ac4c8b88180a8708a8683da4/tools/util/include/cutlass/util/reference/host/tensor_fill.h#L779-L781).

examples/47a_ampere_gemm_universal_streamk_residual/ampere_gemm_universal_streamk_residual.cu
```diff
@@ -0,0 +1,703 @@
...
+  // Initialize tensors using CUTLASS helper functions
+  options.tensor_a.resize(options.problem_size.mk());       // <- Create matrix A with dimensions M x K
+  options.tensor_b.resize(options.problem_size.kn());       // <- Create matrix B with dimensions K x N
+  options.tensor_c.resize(options.problem_size.mn());       // <- Create matrix C with dimensions M x N
+  options.tensor_d.resize(options.problem_size.mn());       // <- Create matrix D with dimensions M x N used to store output from CUTLASS kernel
+  options.tensor_ref_d.resize(options.problem_size.mn());   // <- Create matrix D with dimensions M x N used to store output from reference kernel
+  options.tensor_z.resize(options.problem_size.mn());       // <- Create residual matrix Z with dimensions M x N
+  options.tensor_t.resize(options.problem_size.mn());       // <- Create residual matrix T with dimensions M x N
+
+  // Fill matrix A on host with uniform-random data [4, -4]
```
jackkosaian, line 583: Nit: should be `[-2, 2]` based on the values in the calls.

examples/47a_ampere_gemm_universal_streamk_residual/ampere_gemm_universal_streamk_residual.cu
```diff
@@ -0,0 +1,703 @@
...
+  // Initialize GEMM datasets
+  //
+
+  // Initialize tensors using CUTLASS helper functions
+  options.tensor_a.resize(options.problem_size.mk());       // <- Create matrix A with dimensions M x K
+  options.tensor_b.resize(options.problem_size.kn());       // <- Create matrix B with dimensions K x N
+  options.tensor_c.resize(options.problem_size.mn());       // <- Create matrix C with dimensions M x N
+  options.tensor_d.resize(options.problem_size.mn());       // <- Create matrix D with dimensions M x N used to store output from CUTLASS kernel
+  options.tensor_ref_d.resize(options.problem_size.mn());   // <- Create matrix D with dimensions M x N used to store output from reference kernel
+  options.tensor_z.resize(options.problem_size.mn());       // <- Create residual matrix Z with dimensions M x N
```
jackkosaian, line 580: Since this is the allocation being performed for the vector parameter, should it instead be `{m, 1}`? You can see an example from one of the broadcast unit tests [here](https://github.com/NVIDIA/cutlass/blob/77549ae6c8cf31c7ac4c8b88180a8708a8683da4/test/unit/gemm/device/testbed_gemm_with_broadcast.h#L208-L211).

jackkosaian (commented):
Left a few comments on the example file. I'll look through the implementation later.

examples/47a_ampere_gemm_universal_streamk_residual/ampere_gemm_universal_streamk_residual.cu
alihassanijr: Yes, just found this right now and I'm not surprised if this is what's causing a lot of the issues.

alihassanijr:
Thanks for the notes @jackkosaian .
Regarding a., yes that's pretty much what I expected, which is why I was surprised to see example 47 check for equivalence as opposed to approximation.
One thing I forgot to mention is that the example is mostly based on the existing stream-K example, which is why I was a bit surprised (I could be wrong but I thought one split-K mode even uses atomics, and we'd even have to rule out determinism in that case). 

So I think my reference is incorrect right now, because at the time I started the PR my reference was split-K with a factor of 1, and stream-K had a 3e-3 MSE and did not match (small int operands).
I reverted back to the reference GEMM last night and added in a for loop to handle the vector broadcast and residual add, but unfortunately I'm still seeing a non zero MSE.

I'll keep digging.

examples/47a_ampere_gemm_universal_streamk_residual/ampere_gemm_universal_streamk_residual.cu
alihassanijr: Yeah that was in the original commit. I made it 8 per @hwu36's request. I'll revert back for this stage.

examples/47a_ampere_gemm_universal_streamk_residual/ampere_gemm_universal_streamk_residual.cu
alihassanijr: Although I am a little confused, wouldn't the broadcast vector be of length N instead of M?

https://github.com/NVIDIA/cutlass/blob/77549ae6c8cf31c7ac4c8b88180a8708a8683da4/test/unit/gemm/device/gemm_f16t_f16n_f16t_tensor_op_f16_broadcast_sm80.cu#L152

hwu36: i think it needs to be N

alihassanijr:
Okay so just a summary of the last few commits (my previous comment was a bit verbose so summarizing here):

* One problem was that the broadcast vector size was set to M x N instead of 1 x N.
* Not sure how I missed relative error in utils, but thanks @jackkosaian . I removed the manual MSE and replaced it with that.
* Also changed the vector and tensor references from Z/T to Vector/Tensor.
  * Got rid of tensor for now -- we can worry about that in tests?
* Optional small int (read more below): `--small-int`
* Optional broadcast (as a sanity check): `--disable-broadcast` 

Right now small int values without broadcast match the reference GEMM -- just confirming (kind of) that the GEMM logic wasn't broken.

However, if I switch to floats we get roughly 7e-3 relative error in DP/split-K GEMM and Stream-K GEMM.
I'm not sure if this is reasonable given that DP/split-K are not even modified in this PR.
I'm thinking either it's expected or I'm setting up the example incorrectly.

Once broadcast vector is introduced, whether small int or floats, again none of the kernels match reference.
This could be in part due to the reference (the host for loop added to mimic LinearCombinationWithResidualBlock).
I'm getting [3e-3, 4e-3] error with small int and [2e-2, 3e-2] with floats.

I'll keep digging.

jackkosaian:
@alihassanijr, Thank you for continuing to dig into this.

I think that one component causing a mismatch in the example is that the reference code is applying residual summation regardless of the value of `beta`. However, the broadcast epilogue multiplies the residual by `beta` (see [here](https://github.com/NVIDIA/cutlass/blob/77549ae6c8cf31c7ac4c8b88180a8708a8683da4/include/cutlass/epilogue/thread/linear_combination_residual_block.h#L288-L289)).

(As an aside, you can also use `tensor_d_ref.at({i, j})` in the reference code in your example, rather than calculating the pointer offsets manually. See [this](https://github.com/NVIDIA/cutlass/blob/77549ae6c8cf31c7ac4c8b88180a8708a8683da4/examples/35_gemm_softmax/gemm_softmax.cu#L532-L537) for example.)

alihassanijr:
@jackkosaian Thank you for the pointers.

I do have a question though, in this case `residual` would be `tensor_C`, right?
Because it's single-source.
So `beta * residual` would be `beta * tensor_C`.

For what it's worth I did double check this earlier by looking deeper into the unit test and epilogue operator.

And while doing so I did notice better ways to reference tensor values... I guess I'm still getting used to cute. I'll refactor those.

jackkosaian:
Good catch. Yes, you'll need to pass in the tensor that you want to use as a residual as `ptrC` to `GemmWithBroadcast`.

For the reference calculation, you could run the reference GEMM (alpha * A * B + beta * C) with beta = 0 and then add the residual the way you've been doing so now (but multiplying by beta, and using the same tensor for residual as you used in the `GemmWithBroadcast`).

alihassanijr:
So I should mention I changed a few things since earlier today, so now there's only one for loop for broadcast (no for lop for residual). The reason is that if both GemmWithBroadcast and the reference Gemm are adding `tensor_C` as a residual and multiplying by beta, as long as `Activation` is `Identity` and the `BinaryOp` is `plus`, I should be able to get away with only adding broadcast vector to the reference output. Is that correct?

https://github.com/alihassanijr/cutlass/blob/a3841d91493a3ee5ebfcb0f480060dad206f396c/examples/47a_ampere_gemm_universal_streamk_residual/ampere_gemm_universal_streamk_residual.cu#L597-L611

include/cutlass/gemm/kernel/default_gemm_streamk_with_broadcast.h
```diff
@@ -0,0 +1,143 @@
...
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ **************************************************************************************************/
+
+/*! \file
+  \brief 
+    Defines a Stream-K GEMM with Reduction based on an existing UniversalGemm kernel.
```
jackkosaian, line 34: Nit: Add information about broadcast in this comment.

include/cutlass/gemm/kernel/gemm_streamk_with_fused_epilogue.h
```diff
@@ -0,0 +1,1221 @@
...
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ **************************************************************************************************/
+/*! \file
+    \brief Stream-K Gemm kernel with fused reduction operation.
```
jackkosaian, line 32: Nit: Add information about broadcast in this comment.

include/cutlass/gemm/kernel/gemm_streamk_with_fused_epilogue.h
```diff
@@ -0,0 +1,1221 @@
...
+struct GemmStreamkWithFusedEpilogue;
+
+// GemmStreamkWithFusedEpilogue with multiple sources
+// TODO: Implement
+template <
+  typename Mma_,                  ///! Threadblock-scoped matrix multiply-accumulate 
+  typename Epilogue_,             ///! Epilogue
+  typename ThreadblockSwizzle_    ///! Threadblock swizzling function
+>
+struct GemmStreamkWithFusedEpilogue<Mma_, Epilogue_, ThreadblockSwizzle_, false>;
```
jackkosaian, line 73: Perhaps it's better to remove this for now and to add a `static_assert` somewhere in the epilogue that `kIsSingleSource` is true.

include/cutlass/gemm/kernel/gemm_streamk_with_fused_epilogue.h
```diff
@@ -0,0 +1,1221 @@
...
+        params.params_D,
+        ptr_D,
+        params.block_mapping.problem_size.mn(),
+        thread_idx,
+        threadblock_item_begin);
+
+    // Additional tensor to load from
+    typename Epilogue::TensorTileIterator tensor_iterator(
+        params.params_Tensor,
+        // Only the --first-- (not final like GemmWithFusedEpilogue) block outputs Tensor
```
jackkosaian, line 885: @alihassanijr , could you please explain why this is the case (both for `ptr_Tensor` and `ptr_Vector`) and how this is being enforced in the code? Thanks!

jackkosaian (commented):
Added a few more comments

include/cutlass/gemm/kernel/gemm_streamk_with_fused_epilogue.h
alihassanijr: That's actually an old comment I meant to remove. It is actually the final block and not the first block exactly like GemmWithFusedEpilogue.
The difference is that the StreamK GEMM CTA processing one block might not call the corresponding epilogue, so there's the additional `reduce` method in their epilogues that aggregates the partial sums and then performs the epilogue.
I misunderstood something earlier when I was working on this and thought the original StreamK GEMM called the epilogue in the first block and not the last.
I'm removing that comment.

include/cutlass/gemm/kernel/gemm_streamk_with_fused_epilogue.h
alihassanijr: Good point. I forgot I added that, it serves no purpose for now.
I could add the static assert but I figured once single source was reviewed I'd do multi-source as well, so it wouldn't be missing for long, and until then trying multi-source would just throw template errors instead of a static assert error.
But sure, either way works. I commented out the base and added the assertion.

include/cutlass/gemm/kernel/default_gemm_streamk_with_broadcast.h
alihassanijr: Done. I could get into more details if you'd like.

https://github.com/NVIDIA/cutlass/blob/4518e99f07c7a58ad77a07560d1c3a01da9dbbd5/include/cutlass/gemm/kernel/default_gemm_streamk_with_broadcast.h#L33-L37

include/cutlass/gemm/kernel/gemm_streamk_with_fused_epilogue.h
alihassanijr: Done. I could get into more details if you'd like.

https://github.com/NVIDIA/cutlass/blob/4518e99f07c7a58ad77a07560d1c3a01da9dbbd5/include/cutlass/gemm/kernel/gemm_streamk_with_fused_epilogue.h#L32-L33

include/cutlass/gemm/kernel/gemm_streamk_with_fused_epilogue.h
alihassanijr: https://github.com/NVIDIA/cutlass/blob/4518e99f07c7a58ad77a07560d1c3a01da9dbbd5/include/cutlass/gemm/kernel/gemm_streamk_with_fused_epilogue.h#L75-L84

hwu36:
Splitk does not support relu. Why do we need splitk for verification?

alihassanijr:
Well so we're not doing ReLu yet, activation is `Identity` now.
I'm fine getting rid of split-K to be honest, I just kept it because it was in the original stream-K example (47) to show stream-K can emulate split-K, and compare runtimes.

alihassanijr:
Just pushed two-source. They're also matching with small ints.

hwu36:
@alihassanijr , does this PR still has numeric issues?  I am starting working on this one now.

alihassanijr:
@hwu36 It does not if we're doing small ints. Both split-K and stream-K have the errors if we do floats.

hwu36:
> Both split-K and stream-K have the errors if we do floats.

what kind of error?  i don't expect bit by bit accuracy if using real data.

alihassanijr:
About `1e-2` relative error (cutlass's reference relative error):

```
./examples/47a_ampere_gemm_universal_streamk_residual/47a_ampere_gemm_universal_streamk_residual
10000 timing iterations of 2048 x 2048 x 2048 matrix-matrix multiply

Basic data-parallel GEMM
  Disposition: Failed    Relative error: 0.00701756
  Avg runtime: 0.204862 ms
  GFLOPs: 83860.6

StreamK GEMM with default load-balancing
  Disposition: Failed    Relative error: 0.00696323
  Avg runtime: 0.193961 ms
  GFLOPs: 88573.6
  Speedup vs Basic-DP: 1.056

StreamK emulating basic data-parallel GEMM
  Disposition: Failed    Relative error: 0.00701756
  Avg runtime: 0.206931 ms
  GFLOPs: 83022.1
  Speedup vs Basic-DP: 0.990

Basic split-K GEMM with tile-splitting factor 2
  Disposition: Failed    Relative error: 0.0203117
  Avg runtime: 0.203547 ms
  GFLOPs: 84402.4

StreamK emulating Split-K GEMM with tile-splitting factor 2
  Disposition: Failed    Relative error: 0.00682363
  Avg runtime: 0.206958 ms
  GFLOPs: 83011.3
  Speedup vs Basic-SplitK: 0.984
```

Same example with ints:
```
./examples/47a_ampere_gemm_universal_streamk_residual/47a_ampere_gemm_universal_streamk_residual --small-int
10000 timing iterations of 2048 x 2048 x 2048 matrix-matrix multiply

Basic data-parallel GEMM
  Disposition: Passed    Relative error: 0
  Avg runtime: 0.192819 ms
  GFLOPs: 89098.4

StreamK GEMM with default load-balancing
  Disposition: Passed    Relative error: 0
  Avg runtime: 0.17819 ms
  GFLOPs: 96413
  Speedup vs Basic-DP: 1.082

StreamK emulating basic data-parallel GEMM
  Disposition: Passed    Relative error: 0
  Avg runtime: 0.194421 ms
  GFLOPs: 88364.1
  Speedup vs Basic-DP: 0.992

Basic split-K GEMM with tile-splitting factor 2
  Disposition: Failed    Relative error: 0.0311875
  Avg runtime: 0.189879 ms
  GFLOPs: 90478

StreamK emulating Split-K GEMM with tile-splitting factor 2
  Disposition: Passed    Relative error: 0
  Avg runtime: 0.193384 ms
  GFLOPs: 88838.2
  Speedup vs Basic-SplitK: 0.982
```

hwu36:
the stride of a vector is 0.

alihassanijr:
Yes but if it's always a vector, does it need to be explicitly there?

hwu36:
we cannot change this api now due to internal reasons.

include/cutlass/epilogue/threadblock/epilogue_streamk_with_broadcast.h
```diff
@@ -0,0 +1,1782 @@
+/***************************************************************************************************
```
hwu36, line 1: i understand some code duplication is inevitable, but we want to keep it low.

is there any chance to share the code with the streamk epilogue?  e.g. use the streamk epilogue as the base class, then we can call `Base::reduce()`?

can we refactor the vector load function from non streamk broadcast and share between streamk and non streamk version?

include/cutlass/gemm/kernel/gemm_streamk_with_fused_epilogue.h
```diff
@@ -0,0 +1,2404 @@
...
+
+
+  protected:
+
+    //
+    // Host-only dispatch-utilities
+    //
+
+    /// Pad the given allocation size up to the nearest cache line
+    static size_t cacheline_align_up(size_t size)
```
hwu36, line 297: shall we have a base class so that methods like this can be reused?  maybe something like https://github.com/NVIDIA/cutlass/blob/main/include/cutlass/gemm/device/gemm_universal_base.h ?

examples/47_ampere_gemm_universal_streamk/ampere_gemm_universal_streamk.cu
```diff
@@ -495,23 +495,23 @@ int main(int argc, const char **argv)
   options.tensor_d.resize(options.problem_size.mn());       // <- Create matrix D with dimensions M x N used to store output from CUTLASS kernel
   options.tensor_ref_d.resize(options.problem_size.mn());   // <- Create matrix D with dimensions M x N used to store output from reference kernel
 
-  // Fill matrix A on host with uniform-random data [4, -4]
```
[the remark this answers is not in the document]
alihassanijr: The range was the same, the comment was incorrect in 47.

examples/47a_ampere_gemm_universal_streamk_residual/CMakeLists.txt
```diff
@@ -0,0 +1,35 @@
+
```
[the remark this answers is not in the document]
alihassanijr: Will do.

alihassanijr:
Thank you for the comments. I'll look into merging the classes as mentioned.

> we cannot change this api now due to internal reasons. we can add a default value if you'd like.

Oh yeah absolutely, I just wanted to bring it up. I realize the difficulty in changing signatures. I'll fix the stream-K one so that it's consistent with this as well.

include/cutlass/gemm/kernel/gemm_streamk_with_fused_epilogue.h
alihassanijr: Just refactored common methods in Arguments and Params into `StreamkArgumentsBase` and `StreamkParamsBase`.

There was one slight hitch in getting `StreamkParamsBase`: `kWorkspaceBytesPerBlock` is in the kernel struct, but used in Params. For now I just pass it down to Params through the ctor, but I think there might be a beter way to handle it.

include/cutlass/epilogue/threadblock/epilogue_streamk_with_broadcast.h
alihassanijr: I might be misunderstanding, but we actually are already doing that already. EpilogueStreamkWithBroadcast inherits from both EpilogueBase and EpilogueStreamkBase:

https://github.com/NVIDIA/cutlass/blob/17cb1fb0bb292da26dd0d0467b02a69d41c5e857/include/cutlass/epilogue/threadblock/epilogue_streamk_with_broadcast.h#L226-L238

and calls EpilogueStreamkBase's reduce:

https://github.com/NVIDIA/cutlass/blob/17cb1fb0bb292da26dd0d0467b02a69d41c5e857/include/cutlass/epilogue/threadblock/epilogue_streamk_with_broadcast.h#L461

So it's effectively only adding vector (and tensor) load/store to the StreamK one.

That said, what we definitely can do is inherit some of the methods from EpilogueWithBroadcast to minimize duplication. I'll get started on that.

alihassanijr:
@hwu36 I committed some changes that should address your requests for the most part, at least as far as I understood.
Here's a summary:

* The example was merged into 47, and now it has two binaries: the original and the new one with broadcast
* Vector row offset was added to GemmStreamkWithBroadcast in order to be consistent with universal.
* Base classes for the gemm kernel's arguments and params were created.
* A new reduce method was added to EpilogueWithBroadcast to handle the vector load and epilogue op for a specific fragment offset, which is what Stream-K's reduce requires in order to finalize a partial block. This allows the new StreamK epilogue to be completely based on the existing EpilogueWithBroadcast and EpilogueBaseStreamK, and its reduce method just accums partial sums and calls the reduce in EpilogueWithBroadcast.

I also merged with main to resolve a conflict.

Please let me know if there's anything I missed.

include/cutlass/gemm/kernel/gemm_streamk_with_fused_epilogue.h
```diff
@@ -0,0 +1,2126 @@
...
+
+  /// Argument structure
+  struct Arguments : StreamkArgumentsBase
+  {
+
+    //
+    // Data members
+    //
+
+    GemmUniversalMode mode;
```
hwu36, line 131: my confusion came from this line.  this one is in the base and also here.

alihassanijr: Oh that was my bad. Those should have been removed.
Should I go ahead and remove them, or do we want to add them back to the main arguments class?

hwu36: i am undoing your refactoring of the params.  let us move all refactoring work to the next pr.

alihassanijr: Sorry this should've been sent days ago.

Absolutely, I'll steer clear of pushing more commits to avoid a merge issue, and rebase once you're done to start working on the next PR.

hwu36:
@alihassanijr , would you please do a final check and then we can merge?

alihassanijr (commented):
Looks all good to me.
Checks in the example pass.

PR #903, opened 2023-04-02
Set larger dynamic shared memory size in GemmSoftmax

aakhundov:
Currently, the [`GemmSoftmax`](https://github.com/NVIDIA/cutlass/blob/master/examples/35_gemm_softmax/gemm_with_softmax.h#L579) in [`examples/35_gemm_softmax`](https://github.com/NVIDIA/cutlass/tree/master/examples/35_gemm_softmax) only works with up to 49152 bytes of shared memory. This is problematic, as some of the parameter combinations generated by the [`GenerateSM80`](https://github.com/NVIDIA/cutlass/blob/master/tools/library/scripts/generator.py#L3979) require more shared memory.

This PR suggests setting the `cudaFuncAttributeMaxDynamicSharedMemorySize` to a larger value if required. E.g., if the `kStages = 3` [here](https://github.com/NVIDIA/cutlass/blob/master/examples/35_gemm_softmax/gemm_softmax.cu#L227) in the `gemm_softmax.cu` is increased to 10, the example fails to run with the current code:

```
$ nvcc -arch sm_80 cutlass/examples/35_gemm_softmax/gemm_softmax.cu ...
Device execution failed.

Not verified
```

But it works as expected after the suggested change:

```
$ nvcc -arch sm_80 cutlass/examples/35_gemm_softmax/gemm_softmax.cu ...
         Problem: 16-by-24-by-64, batch size: 16
         Runtime: 0.0199088 ms

          GFLOPs: 39.5017  GFLOPs
Memory bandwidth: 1.72448  GiB/s

Passed
```

### DIFF
--- a/examples/35_gemm_softmax/gemm_with_softmax.h
+++ b/examples/35_gemm_softmax/gemm_with_softmax.h
@@ -578,9 +578,21 @@ public:
 
     int gemm_smem_size = int(sizeof(typename GemmKernel::SharedStorage));
 
+    cudaError_t result;
+
+    if (gemm_smem_size >= (48 << 10)) {
+      result = cudaFuncSetAttribute(cutlass::Kernel<GemmKernel>,
+                                    cudaFuncAttributeMaxDynamicSharedMemorySize,
+                                    gemm_smem_size);
+
+      if (result != cudaSuccess) {
+        return Status::kErrorInternal;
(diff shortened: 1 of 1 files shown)


hwu36:
thanks.  this should improve the performance by increasing the occupancy.

PR #916, opened 2023-04-18
Update functional.h

PityQAQ:
Line1426： - reinterpret_cast<__half const &>(x);   will cause an error: "no operator "-" matches these operands: operand types are: - const __half. "   when arch == "sm80/compute80".    
Moving the negative sign up one line(1425) can solve this error.  Take a minus sign for '__half' instead of a minus sign for "const __half"

Discovered by @furong@pjlab.org.cn @suzhongling@pjlab.org.cn @xuhaoran@pjlab.org.cn

### DIFF
--- a/include/cutlass/functional.h
+++ b/include/cutlass/functional.h
@@ -1422,8 +1422,8 @@ struct negate<Array<half_t, N>> {
     }
 
     if (N % 2) {
-      half_t x = lhs[N - 1];
-      __half lhs_val = -reinterpret_cast<__half const &>(x);
+      half_t x = -lhs[N - 1];
+      __half lhs_val = reinterpret_cast<__half const &>(x);
       result[N - 1] = reinterpret_cast<half_t const &>(lhs_val);
     }
 


hwu36:
this function is moved to `include/cutlass/array.h`.  Do you want to rebase and try again?

mnicely:
@PityQAQ, wanted to follow up based on @hwu36's last comment.

PityQAQ:
Sorry for the late reply. I noticed that in the latest version of the Cutlass library, the function has been moved to `include/cutlass/array.h`. 
However, when I use `cutlass::epilogue::thread::LinearCombinationSigmoid`, I still encounter the same error during compilation:

> cutlass/include/cutlass/array.h(1549): error: no operator "-" matches these operands

And when I use the previous method to add and remove the negative sign, the issue can be resolved.

hwu36:
could you please rebase so your PR is mergeable?

PityQAQ:
@hwu36 oh, I have created a new PR to update this change.  https://github.com/NVIDIA/cutlass/pull/1091

PR #918, opened 2023-04-18
Fix for dangling references in the MHA example

alexander-zinoviev:
There is a dangling reference for the params_ class member that does not survive the iterator's constructor. This PR replaces the dangling reference with a copy.


hwu36:
@danthe3rd

danthe3rd (approved):
LGTM - Thanks!

hwu36 (approved):

PR #925, opened 2023-04-25
Add missing #include directive.

Gregory-Meyer:
Currently [`cutlass/epilogue/thread/linear_combination_clamp.h`](include/cutlass/epilogue/thread/linear_combination_clamp.h) cannot be compiled standalone. This PR addresses this issue; see the commit message.

### DIFF
--- a/include/cutlass/epilogue/thread/linear_combination_clamp.h
+++ b/include/cutlass/epilogue/thread/linear_combination_clamp.h
@@ -40,6 +40,7 @@
 #include "cutlass/array.h"
 #include "cutlass/functional.h"
 #include "cutlass/numeric_conversion.h"
+#include "cutlass/epilogue/thread/scale_type.h"
 
 /////////////////////////////////////////////////////////////////////////////////////////////////
 


hwu36:
I can take the first part.  As to the 2nd part, i prefer it is fixed by llvm.

Gregory-Meyer:
> As to the 2nd part, i prefer it is fixed by llvm.

Sure, I'll talk to @Artem-B and see if we can merge something into LLVM.

Artem-B:
We can take a look at what exactly happens in this particular case, but I'm not very optimistic. Function call ABI assumes that all pointers are generic (except for the kernels), so any non-inlined function call the pointer passes through will block an attempt to infer the address space. 

We may get lucky and find out that we can make things work by force-inlining something in this case, but at the moment we don't have particularly robust general solution. 

To make it work somewhat consistently we'd probably need to figure out a way to automatically specialize functions based on known pointer address space, and allow functions to accept pointers in specific address space. That would likely take a nontrivial amount of work for somewhat limited benefit, with a source-level workaround. I guess. Source-level specialization for specific use cases tends to be easier to accomplish, compared to teaching compiler to do it.

Hardcoding address space casting may not be the best implementation of such specialization, but in general, I think that explicitly telling the compiler that we are guaranteed to work with a shared pointer would probably be the best practical option we have at the moment. 

That said, I don't really know how to implement it portably. C++ type system is generally not aware of address spaces. NVCC often hides things behind inline asm, so that does not help clang. Clang does have an extension that allows specifying pointer address space, but it's not portable to NVCC and I'm not completely sure it will play nicely with C++ types. We may have to specialize the load/store operations themselves and do explicit casting there.

hwu36:
All cutlass functions are force inlined.  If we don't do that, nvcc has the same problem.

Gregory-Meyer:
I was seeing these lines getting compiled to 4x `ST.E.32` instructions instead of 1x `STS.128` instruction on SM 7.5 before, but when I tried to reconstruct a reproducer now I'm no longer able to do that. Must have been some strange interaction that I happened to pick up on -- it's no longer an issue for my work, so, dropped the commit.

hwu36 (approved):

PR #936, opened 2023-05-05
Make operator() const-correct and add missing static functions.

Gregory-Meyer:
Currently, `*Converter::operator()` requires a mutable object to invoke, and there are missing `static result_type convert(source_type const & source)` overloads for certain partial specializations of `*Converter` objects. This PR makes `operator()` const-correct and adds missing function overloads where appropriate.

### DIFF
--- a/include/cutlass/numeric_conversion.h
+++ b/include/cutlass/numeric_conversion.h
@@ -75,13 +75,13 @@ struct NumericConverter {
   static FloatRoundStyle const round_style = Round;
 
   CUTLASS_HOST_DEVICE
-    static result_type convert(source_type const & s) {
+  static result_type convert(source_type const & s) {
 
     return static_cast<result_type>(s);
   }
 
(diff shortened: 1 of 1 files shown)


include/cutlass/numeric_conversion.h
```diff
@@ -732,9 +732,14 @@ struct NumericConverterClamp<cutlass::half_t, S> {
   using source_type = S;
 
   CUTLASS_HOST_DEVICE
-    result_type operator()(source_type const &s) {
+    static result_type convert(source_type const & source) {
     return static_cast<cutlass::half_t>(s);
```
hwu36, line 736: `s` => `source`

Gregory-Meyer: To clarify -- you'd like all of these function parameter definitions to be `source_type const &s`?

hwu36 (approved):

PR #937, opened 2023-05-08
Fix device-only compilation for clang

grypp:
To ensure successful device-only compilation with the clang compiler, the CUDA_ARCH macro is required. Without it, the compiler may try to compile for the host, resulting in errors. See the link below. https://llvm.org/docs/CompileCudaWithLLVM.html#detecting-clang-vs-nvcc-from-code

### DIFF
--- a/include/cute/arch/util.hpp
+++ b/include/cute/arch/util.hpp
@@ -34,7 +34,7 @@
 
 #include <cute/numeric/integer_sequence.hpp>
 
-#if defined(__clang__) && defined(__CUDA__)
+#if defined(__clang__) && defined(__CUDA__) && defined(__CUDA_ARCH__)
   //  __cvta_generic_to_shared was added in Clang 14: https://reviews.llvm.org/D111665
   #define CUTE_CLANG_SUPPORTS_CVTA_GENERIC_TO_SHARED (__clang_major__ >= 14)
 


hwu36:
@Gregory-Meyer

hwu36:
@Gregory-Meyer , do you like #939 solution more?

include/cute/arch/util.hpp
```diff
@@ -34,7 +34,7 @@
 
 #include <cute/numeric/integer_sequence.hpp>
 
-#if defined(__clang__) && defined(__CUDA__)
+#if defined(__clang__) && defined(__CUDA__) && defined(__CUDA_ARCH__)
```
d-k-b, line 37: Most likely this qualification on `defined(__CUDA_ARCH__)` should go down on line # 64. The SUPPORTS or SUPPORTED flags define whether the compiler has the support in general, and then the ACTIVATED definitions qualify the supports flag on whether we're doing device compilation (when applicable).

Gregory-Meyer:
Clang has support for `__forceinline__`, `__host__`, and `__device__` anyways, so I would like for #939 to be merged. My ideal solution would be for `cute::cast_smem_ptr_to_uint` to be annotated as `CUTE_DEVICE` rather than `CUTE_HOST_DEVICE`; the only reasonable implementation would be to cast a on-most-systems-64-bits pointer to 32 bits, which I don't think is a useful operation to do in host code.

Sent #940 for review --

hwu36:
#939 and #940 should cover the clang issue.

grypp:
Nice! I am closing this PR then

PR #939, opened 2023-05-08
Fix cuTE compilation with clang

JanuszL:
- clang 1.14 complains about missing function from a host call:
  ```
  cutlass/include/cute/arch/util.hpp:106:32: error: no matching function for call to '__cvta_generic_to_shared' return static_cast<uint32_t>(__cvta_generic_to_shared(ptr));```
- fixes this by defining CUTE_HOST_DEVICE for clang as well

### DIFF
--- a/include/cute/config.hpp
+++ b/include/cute/config.hpp
@@ -30,7 +30,7 @@
  **************************************************************************************************/
 #pragma once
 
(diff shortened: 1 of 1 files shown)


JanuszL:
Or the proper way is https://github.com/NVIDIA/cutlass/pull/937

hwu36 (approved):

PR #940, opened 2023-05-08
Fix host compilation of cute::cast_smem_ptr_to_uint.

Gregory-Meyer:
See the commit messages for why I created this PR and what it does.

### DIFF
--- a/include/cute/arch/util.hpp
+++ b/include/cute/arch/util.hpp
@@ -90,7 +90,7 @@ namespace cute
 {
 
 /// CUTE helper to cast SMEM pointer to unsigned
-CUTE_HOST_DEVICE
+CUTE_DEVICE
 uint32_t
 cast_smem_ptr_to_uint(void const* const ptr)
 {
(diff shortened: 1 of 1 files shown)


include/cute/arch/util.hpp
```diff
@@ -36,39 +36,35 @@
...
+#    define CUTE_CLANG_SUPPORTS_NVVM_GET_SMEM_POINTER (__clang_major__ >= 15)
+#  endif
 #endif
 
 #if defined(__NVCC__) || defined(__CUDACC_RTC__)
   // __cvta_generic_to_shared added in CUDA 11+
-  #if defined(__CUDA_ARCH__) && (__CUDACC_VER_MAJOR__ >= 11)
-    #define CUTE_NVCC_SUPPORTS_CVTA_GENERIC_TO_SHARED 1
-  #endif
+#  define CUTE_NVCC_SUPPORTS_CVTA_GENERIC_TO_SHARED (__CUDACC_VER_MAJOR__ >= 11)
```
hwu36, line 52: we need to write
```
#if xxx
  #define xxx 1
#endif
```
otherwise, some compiler will yell.

Gregory-Meyer: Done -- changed all checks to `#define VAR 1` only if the associated condition is true.

hwu36:
in my small commit, i kept runtime assert part just in case anyone wants to run cutlass with a home brew compiler.

hwu36 (approved):

PR #945, opened 2023-05-12
Make bias stride configurable in EpilogueTensorBroadcast

aakhundov:
Currently, the bias stride is hard-coded to `Stride<_1, _0, _0>{}` (column-major) [here](https://github.com/NVIDIA/cutlass/blob/main/include/cutlass/epilogue/collective/epilogue_tensor_broadcast.hpp#L177) in `EpilogueTensorBroadcast`. This makes it impossible to use a row-major bias vector. This PR adds another parameter to the `EpilogueTensorBroadcast::Arguments`---`dBias`---following the `StrideC` of the source other tensors.

N.B.: I'm not sure the change would leave the unit tests [here](https://github.com/NVIDIA/cutlass/blob/fcfbd23e26328df8c8b720a161ada95a3eb725e8/test/unit/gemm/device/sm90_gemm_f16_f16_f16_tensor_op_f32_tensor_broadcast.cu) in consistent state. If not, and if the PR is acceptable, I'd kindly ask the reviewer to point out the inconsistencies in the tests and a desirable way to resolve them. Thanks!

### DIFF
--- a/include/cutlass/epilogue/collective/epilogue_tensor_broadcast.hpp
+++ b/include/cutlass/epilogue/collective/epilogue_tensor_broadcast.hpp
@@ -112,6 +112,7 @@ public:
     ElementBias* ptr_Bias = nullptr;
     ElementC* ptr_C0 = nullptr;
     ElementC* ptr_C1 = nullptr;
+    StrideC dBias{};
   };
 
   // Device side epilogue params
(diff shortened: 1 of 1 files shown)


hwu36:
@thakkarV

thakkarV:
Hi @aakhundov , thanks a lot for this contribution. I thought a bit more about this after our conversations offline -- and I think it may be better to not support this use case for now. The reason being that for most users who may not be intimately familiar with the kernels, this could result in an unexpected performance cliff, especially because this epilogue can also work provided a full rank - 3 bias tensor in which case each load of that tensor would not be done via a proper tiled gmem load leading to bad perf. That said, I would be happy to accept this as a contribution if we can `static_assert` that only exactly 1 of the strides is set t to _1, and the other two are set to _0 for now.

thakkarV:
Expected indeed :) To summarize, we recommend swapping A and B to compute a transposed output D while still using a column-major bias vector for best performance.

PR #979, opened 2023-06-22
Add RMS norm to device util

masahi:
Add [RMS normalization](https://arxiv.org/abs/1910.07467) kernel to `tools/utils/`, based on the existing layer norm kernel. This normalization method is used, in particular, by llama. It is essentially a simplified layer norm.

Two impls have been added - a naive one that works for any shape / dtype, and a relatively fast one for half + multiple of 8 size. Based on feedbacks I can simply copy more variants from the layer norm kernels.

@hwu36


hwu36:
thank you very much as always, @masahi

PR #995, opened 2023-07-10
fix epilogue iterator error

ChangyouSiom:


hwu36 (approved):

hwu36:
curious, did you hit any error?  the upper dimension is rarely used.

PR #1021, opened 2023-08-02
Fix numeric conversion warning

vincentccc:
fix unused var


include/cutlass/numeric_conversion.h
```diff
@@ -2352,8 +2352,9 @@ struct FastNumericArrayConverter {
   static result_type convert(source_type const &s) {
     result_type result;
```
hwu36, line 2353: Thank you for the PR.  Could we just remove this line?

vincentccc: Thanks for your time to review, sure, I will follow it.

hwu36 (approved):

PR #1037, opened 2023-08-10
torch.bfloat16 support in cutlass python

sophiawisdom:
haven't tested, think this works. will test soon.

Previously:
```
>>> gemm = cutlass.op.Gemm(element_A=torch.bfloat16, element_B=torch.bfloat16, element_C=torch.float16, layout=cutlass.LayoutType.RowMajor)
  File "<stdin>", line 1, in <module>
  File "/workspace/cutlass/python/cutlass/op/gemm.py", line 245, in __init__
    elements.append(datatypes.library_type(elt_to_set))
  File "/workspace/cutlass/python/cutlass/utils/datatypes.py", line 223, in library_type
    raise Exception(f"No available conversion from type {inp} to a library type.")
Exception: No available conversion from type torch.bfloat16 to a library type.
```

### DIFF
--- a/python/cutlass/utils/datatypes.py
+++ b/python/cutlass/utils/datatypes.py
@@ -150,6 +150,10 @@ try:
     import bfloat16
 
     bfloat16_available = True
+    if torch_available:
+        _torch_to_library_dict[torch.bfloat16] = cutlass.DataType.bf16
+        _library_to_torch_dict[cutlass.DataType.bf16] = torch.bfloat16
+    
 except ImportError:
     bfloat16_available = False
 


python/cutlass/utils/datatypes.py
```diff
@@ -150,6 +150,10 @@ def torch_type(inp):
     import bfloat16
 
     bfloat16_available = True
+    if torch_available:
```
jackkosaian, line 153: Thanks for this contribution, @sophiawisdom.

I think that the best place to add this to the `_torch_to_library_dict` and `_library_to_torch_dict` would be here: https://github.com/NVIDIA/cutlass/blob/4575443d446807b8e9d7f77185452899020ac8c9/python/cutlass/utils/datatypes.py#L118-L134.

Could you please change this to add torch.bfloat16 there?

Thanks.

sophiawisdom: Done

sophiawisdom:
@jackkosaian is this suitable for merge?

Issue #1040, opened 2023-08-11
[BUG] CuTE print fails to compile with compiler flags -Wformat   -Wformat-security   -Werror=format-security


ahendriksen:
**Describe the bug**
`cute::print` fails to compile when compiler flags `Wformat   -Wformat-security   -Werror=format-security` are passed. 

**Steps/Code to reproduce bug**
Code:
``` cuda
#include <cute/tensor.hpp>
int main() {
  using namespace cute;
  auto layout = make_layout(make_shape (_2{}, _3{}), make_stride(_3{}, _1{}));
  print(layout);
}
```

Compile with :
```
# This is fine:
g++ -I/path/to/cutlass -std=c++17 repro.cc             
# This warns (can fail with -Werr)
g++ -I/path/to/cutlass -std=c++17 -Wformat   -Wformat-security    repro.cc      
# This fails:
g++ -I/path/to/cutlass -std=c++17 -Wformat   -Wformat-security   -Werror=format-security repro.cc
```

**Expected behavior**
I expect the code to continue to work and issue no warning with these flags.

**Environment details (please complete the following information):**
This issue can occur on:
- Debian-based GCC, which [turns some of these flags on](https://salsa.debian.org/toolchain-team/gcc/-/blob/master/debian/patches/gcc-distro-specs.diff#L45).
- Occurs on Nix-based systems which turns on `-Werror=format-security` (fails to compile completely).

The reproducer was with nvcc 12.2, GCC 12.2.0.

ahendriksen:
Fixed in PR #1041

PR #1041, opened 2023-08-11
Avoid cute::print compiler warnings with -Wformat-security

ahendriksen:
Fixes issue #1040.

`cute::print` failed to compile when compiler flags `-Wformat   -Wformat-security   -Werror=format-security` were passed. 

With this commit, `cute::print` works also when it is passed a single string and does not emit warnings (even when these compiler flags are present).

### DIFF
--- a/include/cute/util/print.hpp
+++ b/include/cute/util/print.hpp
@@ -145,4 +145,10 @@ print(T const&... t) {
   (print(t), ...);
 }
 
+CUTE_HOST_DEVICE
+void
+print(char const* format) {
+  printf("%s", format);
+}
+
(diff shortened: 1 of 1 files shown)


thakkarV:
@hwu36 this is a warning that @mhoemmen and I just attempted to fix ourselves too. We should accept this MR

mhoemmen:
@thakkarV I agree -- thanks @ahendriksen ! : - D

thakkarV:
@hwu36 @kerrmudgeon this MR has merged into the internal repo, we are good to merge this.

hwu36 (approved):

Issue #1042, opened 2023-08-12
[BUG] Compilation Times too Damn High. 


manishucsd:
CUTLASS commit
```
(main) $ git log
commit 4575443d446807b8e9d7f77185452899020ac8c9 (HEAD -> main, origin/main, origin/HEAD)
Author: ANIKET SHIVAM <3268307+ANIKET-SHIVAM@users.noreply.github.com>
Date:   Mon Aug 7 14:50:32 2023 -1000

    CUTLASS 3.2 (#1024)
    
    * CUTLASS 3.2
```

NVCC version
```bash
$ nvcc --version
nvcc: NVIDIA (R) Cuda compiler driver
Copyright (c) 2005-2023 NVIDIA Corporation
Built on Tue_Jun_13_19:16:58_PDT_2023
Cuda compilation tools, release 12.2, V12.2.91
Build cuda_12.2.r12.2/compiler.32965470_0
```

cmake
```bash
build $ cmake ../cutlass/ -DCUTLASS_NVCC_ARCHS='80' -DCUTLASS_LIBRARY_KERNELS="tensorop_i16832gemm_s8_128x128*tn*align16" -DCUTLASS_NVCC_KEEP=OFF -DCUTLASS_LIBRARY_IGNORE_KERNELS="cutlass_tensorop_h16816gemm_grouped*,cutlass_tensorop_h16816gemm_planar*"
```

Repro
```bash
$ time make cutlass_profiler -j
[  0%] Building CXX object tools/library/CMakeFiles/cutlass_library_objs.dir/src/manifest.cpp.o
[  0%] Building CUDA object tools/library/CMakeFiles/cutlass_library_objs.dir/src/handle.cu.o
[  0%] Building CUDA object tools/library/CMakeFiles/cutlass_library_objs.dir/src/singleton.cu.o
[  0%] Building CUDA object tools/library/CMakeFiles/cutlass_library_objs.dir/src/util.cu.o
[  0%] Building CUDA object tools/library/CMakeFiles/cutlass_library_objs.dir/src/reference/gemm.cu.o
[  0%] Building CUDA object tools/library/CMakeFiles/cutlass_library_objs.dir/src/operation_table.cu.o
[  0%] Building CUDA object tools/library/CMakeFiles/cutlass_library_objs.dir/src/reference/gemm_fp8.cu.o
[  0%] Building CUDA object tools/library/CMakeFiles/cutlass_library_objs.dir/src/reduction/reduction_device.cu.o
[...]

PR #1043, opened 2023-08-12
Shard gemm reference templates into multiple TUs for parallel compilation

thakkarV:


manishucsd:
fb572ddda341bf7b4621dfb4ca2e798466137c1b brings it down to 4 mins for the machine I am using. The previous commit was around 8 mins and we started at 17 mins. It looks good and we can close the issue #1042 once the PR 1043 merges. Thanks again!

tools/library/scripts/generator.py
```diff
@@ -4632,20 +4632,21 @@ def GenerateSM90_TensorOp_fp8_WGMMA_gemm(manifest, cuda_version):
     else:
       assert False, "math inst is not supported"
 
+    # some schedules disabled to save on library size
     if CudaToolkitVersionSatisfies(cuda_version, 12, 1):
       schedules = [
-        [KernelScheduleType.ScheduleAuto, EpilogueScheduleType.ScheduleAuto],
+        # [KernelScheduleType.ScheduleAuto, EpilogueScheduleType.ScheduleAuto],
```
IonThruster, line 4638: Would recommend not to comment the auto schedules.

thakkarV: added back for 12.1.

hwu36 (approved):

PR #1072, opened 2023-08-29
Make files under tools/library/scripts/ use relative import 

ipiszy:
## Description
This is another attempt to use relative import for files under `tools/library/scripts/`, so that it's easier to import cutlass python scripts in other project. e.g. A related PR: https://github.com/pytorch/pytorch/pull/107802/.

Also updated `tools/library/CMakeLists.txt` to add `tools/library` to `PYTHONPATH` and use `python -m package.module` to execute `generator.py` to solve previous import errors.

## Testing
Commandlines below runs successfully:
```
cmake .. -DCUTLASS_NVCC_ARCHS=80
make cutlass_profiler -j16
```

### DIFF
--- a/tools/library/CMakeLists.txt
+++ b/tools/library/CMakeLists.txt
@@ -42,8 +42,8 @@ target_include_directories(
   )
 
 target_link_libraries(
-  cutlass_library_includes 
-  INTERFACE 
+  cutlass_library_includes
+  INTERFACE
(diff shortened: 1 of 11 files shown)


ipiszy:
cc @jackkosaian @hwu36

mnicely:
Added in v3.2.1

PR #1100, opened 2023-09-19
Fix infinite print call for tiled copy

reed-lau:
```c++
#include <cute/tensor.hpp>

int main() {
  using T = cute::half_t;
  using namespace cute;
  using X = Underscore;

  using copy_op = SM75_U32x1_LDSM_N;
  using traits = Copy_Traits<copy_op>;
  using atom = Copy_Atom<traits, T>;
  auto tiled_copy1 =
      TiledCopy<Copy_Atom<traits, T>,
                Layout<Shape<_32, _2>, Stride<_2, _1>>, Shape<_8, _8>>{};

  auto tiled_copy2 =
      TiledCopy<Copy_Atom<copy_op, T>,
                Layout<Shape<_32, _2>, Stride<_2, _1>>, Shape<_8, _8>>{};

  print(tiled_copy1);
  print(tiled_copy2);

  return 0;
}
```
when I print tiled_copy2 with print api, the compiler will choose the overload one from https://github.com/NVIDIA/cutlass/blob/8783c41851cd3582490e04e69e0cd756a8c1db7f/include/cute/util/print.hpp#L144C1-L144C1 , and it will cause a infinite loop.

compile command: `nvcc test.cu -arch=sm_80 -std=c++17 -Iinclude`

### DIFF
--- a/include/cute/atom/copy_atom.hpp
+++ b/include/cute/atom/copy_atom.hpp
@@ -646,10 +646,14 @@ void
 print(TiledCopy<Atom, Args...> const& copy, char const* pad = "")
 {
   using Copy = TiledCopy<Atom, Args...>;
+  using Traits = typename Atom::Traits;
+  using ValType = typename Atom::ValType;
+  using CopyAtomT = Copy_Atom<Traits, ValType>;
+
   print("TiledCopy\n");
   print("  Tiler_MN:       "); print(typename Copy::Tiler_MN{});       print("\n");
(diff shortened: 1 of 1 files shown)


reed-lau:
this crash disappared when apply the patch at https://github.com/NVIDIA/cutlass/pull/1104

thakkarV:
looks like #1104 is a more general fix, can we close this?

reed-lau:
It have been fixed in the newest main branch, so I will close this PR.

PR #1101, opened 2023-09-19
Suppress missing return value warning for cute::get

reed-lau:
when I compile a test code in linux with g++9.4.0 and nvcc11.4, there is a warning
image

I fixed it following the other code blocks.


thakkarV:
@hwu36 this looks good to me, we can merge.

thakkarV:
EDIT: with 3.2.1 we have refactored tuple get() and this WAR should no longer be needed. @reed-lau can you please confirm that 3.2.1 no longer has this warning?

reed-lau:
It had been fixed. I will close this PR.

PR #1102, opened 2023-09-19
Fix GmmaDescriptor print format string error

reed-lau:
When compiling tensor.hpp related code, a warning will appear.
image

I fixed it. But I'm not sure if this is the expected format.

### DIFF
--- a/include/cute/arch/mma_sm90_desc.hpp
+++ b/include/cute/arch/mma_sm90_desc.hpp
@@ -140,7 +140,7 @@ union GmmaDescriptor
   CUTE_HOST_DEVICE friend void print(GmmaDescriptor const& t)
   {
     #if !defined(__CUDACC_RTC__)
-    printf("GmmaDescriptor: 0x%016 %lli\n", static_cast<long long>(t.desc_));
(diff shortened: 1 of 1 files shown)


mhoemmen:
@hwu36 Thanks for the heads-up!  I'll communicate offline.

include/cute/arch/mma_sm90_desc.hpp
```diff
@@ -140,7 +140,7 @@ union GmmaDescriptor
   CUTE_HOST_DEVICE friend void print(GmmaDescriptor const& t)
   {
     #if !defined(__CUDACC_RTC__)
-    printf("GmmaDescriptor: 0x%016 %lli\n", static_cast<long long>(t.desc_));
```
hwu36, line 143: could you please change it to `    printf("GmmaDescriptor: 0x%016llx\n",   static_cast<unsigned long long>(t.desc_));`, then i will just merge?

hwu36 (changes requested):

include/cute/arch/mma_sm90_desc.hpp
reed-lau: done

hwu36 (approved):

PR #1116, opened 2023-09-28
Fix Parallel Split-K Profiling for Mixed-precision GEMMs 

manishucsd:
# Summary
This PR fixes support for mixed-precision GEMMs with parallel split-k reduction. In scenarios where `ElementOutput = F16` and `ElementAccumulator = F32`, to enable parallel split-k reduction, the `underlying_operation` should have both `ElementOutput` and `ElementAccumulator` set to `F32`. The conversion from `F32` to `F16` is addressed during the reduction process.

In profiling/autotuning mixed-precision GEMMs with parallel split-k reduction, it's crucial to adjust only the `ElementOutput` from `F16` to `F32`. Other parameters like tile shape, warp count, and instruction shape should remain consistent with the original `operation`. This PR ensures this behavior.

## CMake Build Command 
```bash
make --no-warn-unused-cli -DCMAKE_BUILD_TYPE:STRING=Debug -DCUTLASS_NVCC_ARCHS:STRING=80 -DCUTLASS_NVCC_KEEP:STRING=OFF -DCUTLASS_ENABLE_F16C:STRING=ON -DCUTLASS_DEBUG_TRACE_LEVEL:STRING=0 -DCUTLASS_LIBRARY_KERNELS:STRING=s16816gemm_bf16_*_tn_align8,s16816gemm_f16_*_tn_align8 
```

## Behavior Before the PR
When profiling the operation `cutlass_tensorop_f16_s16816gemm_f16_64x64_32x10_tn_align8` with parallel split-k reduction, the expected GEMM tile shape should be `64x64_32x10`. Yet, enabling `--split-k-mode=parallel --split-k-slices=2` unexpectedly altered the threadblock shape of the `underlying_operation` to `256x128_32x3`. This behavior was consistent for all parallel split-k reductions, always selecting the first threadblock from the `GemmOperationVectorMap`. While, the profiler (auto-tuner) built on top of the profiler should select `64x64_32x10` with `ElementOutput=F32`.

## Behavior After the PR
Post this PR, the operation and `underlying_operation` maintain the anticipated tile shape of `64x64_32x10` when using parallel split-k reduction.

```bash
manigupta@manigupta-gpu /mnt/disks/gcloud_workspace/repos/cutlass/cutlass_tree_1/build $ ./tools/profiler/cutlass_profiler --kernels=cutlass_tensorop_f16_s16816gemm_f16_64x64_32x10_tn_align8 --split-k-mode=parallel --split-k-slices=2
operation           : cutlass_tensorop_f16_s16816gemm_f16_64x64_32x10_tn_align8
underlying_operation: cutlass_tensorop_s16816gemm_f16_64x64_32x10_tn_align8
```

**Note**: This PR currently includes two debug prints which will be removed prior to final merging after review.


hwu36:
you could remove/comment out the prints now.

manishucsd:
> you could remove/comment out the prints now.

DONE. I have removed the comments for now, but can I have feature request here :). Cam we have a verbose/logging mode in CUTLASS profiler to dump such information for debugging. 

- Details from initialize_workspace such as:
  - Global memory allocations details: size in bytes for each operand/results, pointer address.
  - Underlying Operation / Operation
  - Reduction Operation
 
We can add more debug info in this mode if we decide to have such a feature. It maybe helpful in debugging such issues as I think  the CI only test basic profiler health, i.e., it runs without crashing.

PR #1119, opened 2023-09-29
Fix type typo in rmsnorm

abcdabcd987:
Initially the variable `h4` is `half4`, but its last two fields are not used. Based on the semantics and the context, I believe it should be `half2`.


hwu36:
@masahi

masahi (approved):
Thanks.

hwu36 (approved):

PR #1120, opened 2023-09-30
Fix example 47

apuaaChen:
This pull request fixes the error of example 47. Originally the GemmUniversalAdapter performs internal transpose for row-major layout C. Yet under EVT with row-major output, internal transpose is not required. Using GemmUniversalBase can avoid this internal transpose.


hwu36:
it is already part of 3.3

PR #1124, opened 2023-10-05
fix missing return warning for repeat and axpby

reed-lau:
When I compile a test code in linux with g++9.4.0 and nvcc11.4, there are serval warnings:

image
image

I fixed it following the other code blocks.


hwu36:
@thakkarV

thakkarV:
@hwu36 yeah this seems fine to patch in

hwu36 (approved):

PR #1127, opened 2023-10-05
Add missing int64 and uint64 overloads for conj

klecki:
In 3.2.1 the `conj` implementation was changed from generic, templated one to listing trivial overloads for scalar types. There were no overloads for 64 bit integer types, so when the code was instantiated with such types the compiler found several close-matching overloads that would require conversion and report an error.
This PR adds the missing overloads.


ANIKET-SHIVAM (approved):

ANIKET-SHIVAM:
@hwu36 feel free to merge.

hwu36 (approved):

PR #1140, opened 2023-10-13
change 'not' to symbol '!' and unify the code style for 'else'

reed-lau:
1. While the `not` operator is functional in C++, its usage is rather restricted. Therefore, I recommend employing the more prevalent `!` symbol to denote logical negation, which would enhance comprehension for subsequent readers.
2. make the `else` code style following the other code blocks in this file.

Reference:
https://en.cppreference.com/w/cpp/language/operator_alternative

### DIFF
--- a/include/cute/algorithm/tuple_algorithms.hpp
+++ b/include/cute/algorithm/tuple_algorithms.hpp
@@ -392,7 +392,7 @@ CUTE_HOST_DEVICE constexpr
 auto
 none_of(T const& t, F&& f)
 {
-  return not any_of(t, f);
+  return !any_of(t, f);
 }
 
 //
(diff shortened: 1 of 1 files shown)


thakkarV:
Thanks for your PR. Unfortunately, we will most likely not be accepting it at this time for two reasons:

- CuTe headers have a different  code style that is subtly but consistently different from that of rest of CUTLASS 
- the change from using `not` to `!` is not desirable. `not` kw was used intentionally in these places as it is much more clear to see the negation operation when reading source

reed-lau:
the 'not symbol fix' is reverted.

reed-lau:
@thakkarV could you help review?

d-k-b (approved):

PR #1147, opened 2023-10-17
Fix is_zero

cyyever:
Fix incorrect use of masks in is_zero

### DIFF
--- a/include/cutlass/predicate_vector.h
+++ b/include/cutlass/predicate_vector.h
@@ -178,8 +178,9 @@ struct PredicateVector {
   /// Returns mask of last word.
   CUTLASS_HOST_DEVICE static constexpr bool computeLastWordMask() {
     Storage mask(0);
+    constexpr int count = (kBytes % sizeof(Storage) == 0) ? sizeof(Storage) : (kBytes % sizeof(Storage));
     CUTLASS_PRAGMA_UNROLL
-    for (int byte = 0; byte < kBytes % sizeof(Storage); ++byte) {
+    for (int byte = 0; byte < count; ++byte) {
       mask |= (kByteMask << (byte * 8));
     }
     return mask;
@@ -515,19 +516,12 @@ struct PredicateVector {
   CUTLASS_HOST_DEVICE bool is_zero() const {
     constexpr Storage mask = computeWordMask();
     Storage result = 0;
-    if constexpr (kBytes % sizeof(Storage) == 0) {
-      CUTLASS_PRAGMA_UNROLL
-      for (int word = 0; word < kWordCount; ++word) {
-        result |= (storage(word) & mask);
-      }
-    } else {
-      CUTLASS_PRAGMA_UNROLL
-      for (int word = 0; word < kWordCount - 1; ++word) {
-        result |= (storage(word) & mask);
-      }
-      constexpr Storage last_word_mask = computeLastWordMask();
-      result |= (storage(kWordCount - 1) & last_word_mask);
+    CUTLASS_PRAGMA_UNROLL
+    for (int word = 0; word < kWordCount - 1; ++word) {
+      result |= (storage(word) & mask);
     }
+    constexpr Storage last_word_mask = computeLastWordMask();
+    result |= (storage(kWordCount - 1) & last_word_mask);
     return result == 0;
   }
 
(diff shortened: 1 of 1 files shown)


include/cutlass/predicate_vector.h
```diff
@@ -490,16 +490,12 @@ struct PredicateVector {
...
   /// Returns true if entire predicate array is zero.
   CUTLASS_HOST_DEVICE bool is_zero() const {
-    Storage mask(0);
-    for (int byte = 0; byte < sizeof(Storage); ++byte) {
-      Storage byte_mask = (((1 << kPredicatesPerByte) - 1) << kPredicateStart);
-      mask |= (byte_mask << (byte * 8));
-    }
-    uint32_t result = 0;
     for (int word = 0; word < kWordCount; ++word) {
-      result |= storage(word);
```
hwu36, line 500: i think the right fix is to apply mask here.

thakkarV:
Short circuiting the loop is not necessarily an optimization as it may cause warp divergence.

cyyever:
Recheck the code and fix it again.

include/cutlass/predicate_vector.h
```diff
@@ -490,15 +490,20 @@ struct PredicateVector {
...
     Storage mask(0);
     for (int byte = 0; byte < sizeof(Storage); ++byte) {
-      Storage byte_mask = (((1 << kPredicatesPerByte) - 1) << kPredicateStart);
       mask |= (byte_mask << (byte * 8));
     }
-    uint32_t result = 0;
-    for (int word = 0; word < kWordCount; ++word) {
-      result |= storage(word);
+    Storage result = 0;
+    for (int word = 0; word < kWordCount - 1; ++word) {
```
hwu36, line 499: why do we need to minus 1?

cyyever: To  exclude the last word which is handled below

hwu36: i see. thanks.

looking at the other methods in this struct, I think we can assume that we don't have a partial word.

cyyever: > i see. thanks.
> 
> looking at the other methods in this struct, I think we can assume that we don't have a partial word.

I disagree, may be there are hidden bugs in other methods because of word handling.

cyyever:
@thakkarV No, I reverted early stop. Now the new code should have the same cost of the old code with negligible overhead of OR masking in the loop.

include/cutlass/predicate_vector.h
hwu36: the register size is always 32bit.  it does not make sense if predicatevector is not multiple of 32bit.  you can add an assert to assure it if you want.

cyyever: I mean kBytes may not be a multiple of 32bit. So we have to deal with the last word because it is possible to contain out-of-bound bytes.

include/cutlass/predicate_vector.h
```diff
@@ -490,14 +511,18 @@ struct PredicateVector {
...
-    for (int word = 0; word < kWordCount; ++word) {
-      result |= storage(word);
+    constexpr Storage mask = computeWordMask();
+    Storage result = 0;
+    if constexpr (kBytes % sizeof(Storage) == 0) {
+      for (int word = 0; word < kWordCount; ++word) {
+        result |= (storage(word) & mask);
+      }
+    } else {
+      for (int word = 0; word < kWordCount - 1; ++word) {
```
hwu36, line 521: just use this part.  if...else can make the compiler life a little harder.

include/cutlass/predicate_vector.h
```diff
@@ -162,6 +165,24 @@ struct PredicateVector {
...
+    for (int byte = 0; byte < sizeof(Storage); ++byte) {
+      mask |= (kByteMask << (byte * 8));
+    }
+    return mask;
+  }
+
+  /// Returns mask of last word.
+  CUTLASS_HOST_DEVICE static constexpr bool computeLastWordMask() {
+    Storage mask(0);
+    for (int byte = 0; byte < kBytes % sizeof(Storage); ++byte) {
```
hwu36, line 180: all for loops with compile time constant trip count needs `CUTLASS_PRAGMA_UNROLL`

cyyever: This is a constexpr function, I am not sure whether it works well with CUTLASS_PRAGMA_UNROLL. But I will test it locally and submit.

hwu36 (approved):

PR #1153, opened 2023-10-23
Support ElementD to be void for tma

kongroo:
We can use Sm90AuxStore to customize writing the D tensor (e.g. scatter the D tensor to multiple pointers). In such cases, it is necessary to remove the original writing logic of the D tensor. This PR allowes passing a nullptr for ptr_D to make storing-to-D a noop.


hwu36:
@thakkarV @yzhaiustc

thakkarV:
Thanks a lot for your contribution, I will review your MR

include/cutlass/epilogue/collective/sm90_epilogue_tma_warpspecialized.hpp
```diff
@@ -73,7 +74,7 @@ template <
   class CopyOpR2S_
 >
 class CollectiveEpilogue<
-    Sm90TmaWarpSpecialized<StagesC_,StagesD_,FragmentSize_,ReuseSmemC_>,
```
thakkarV, line 76: The way we have done this is by setting the type of `ElementD = void`. You can see this for the `ElementC` residual load as well. Can we follow the same pattern here for consistency?

include/cutlass/epilogue/collective/sm90_epilogue_tma_warpspecialized.hpp
```diff
@@ -600,7 +612,14 @@ class CollectiveEpilogue<
         }
 
         // Copy tile from register to smem
-        copy(tiled_r2s, tRS_rD, tRS_sD(_,_,_,store_pipe_producer_state.index()));
+        [&]() -> void {
+          if constexpr (DispatchPolicy::EnableNullptrD) {
+            if (params.is_nullptr_d) {
```
thakkarV, line 617: is a dynamic branch needed here? looks like you always check against the template parameter only. So we can totally avoid any dynamic checks within the kernel which will ensure no perf regressions.

include/cutlass/epilogue/collective/sm90_epilogue_tma_warpspecialized.hpp
```diff
@@ -209,6 +210,7 @@ class CollectiveEpilogue<
     typename FusionCallbacks::Params thread{};
     TMA_C tma_load_c;
     TMA_D tma_store_d;
+    bool is_nullptr_d = false;
```
thakkarV, line 213: since you are only using the template parameter, this is not needed.

include/cutlass/epilogue/dispatch_policy.hpp
```diff
@@ -131,13 +131,15 @@ template<
   int StagesC_,
   int StagesD_,
   int FragmentSize_,
-  bool ReuseSmemC_
+  bool ReuseSmemC_,
+  bool EnableNullptrD_ = true
```
thakkarV, line 135: See comment above about using the element D to detect this instead of adding a new parameter here.

yzhaiustc, line 135: Nit: should the `EnableNullptrD_` be `false` by default?

richardmcai:
if you know that storing D will be disabled then adding support for `void` D is probably the better way here, instead of adding these runtime checks. You can follow how `is_source_supported` is used for reference, you could create a similar `is_destination_supported` variable for void D

include/cutlass/epilogue/collective/sm90_epilogue_tma_warpspecialized.hpp
kongroo: I find it not easy to make both C and D void because ElementD is used in many places such as construction tensors and calculating tile shapes in the builder.

include/cutlass/epilogue/collective/sm90_epilogue_tma_warpspecialized.hpp
kongroo: I followed the code pattern in Sm90AuxStore. I think there are two possible implementations here. 1. use EnableNullptrD = true as default and dynamic check; 2. use DisableWriteD = false as default and constant check; I think It makes sense to prefer the second one for D.

kongroo: I have updated to use static branches. Could you please take a look?

include/cutlass/epilogue/dispatch_policy.hpp
kongroo: Updated

yzhaiustc: Thank you. Cleared on my side.

thakkarV:
@richardmcai thoughts on this MR now?

richardmcai:
> @richardmcai thoughts on this MR now?

It should be fine, a couple points
* It's not clear to me how this feature is going to be exposed to the user since there are no builder modifications. Is this collective being instantiated manually?
* Ultimately this flag should be deprecated in favor of supporting void D to maintain consistency with void C behavior. Fine with accepting it for now with the understanding that it may be deprecated in a future release (with appropriate prior notice or compile warnings)

kongroo:
> > @richardmcai thoughts on this MR now?
> 
> It should be fine, a couple points
> 
> * It's not clear to me how this feature is going to be exposed to the user since there are no builder modifications. Is this collective being instantiated manually?
> * Ultimately this flag should be deprecated in favor of supporting void D to maintain consistency with void C behavior. Fine with accepting it for now with the understanding that it may be deprecated in a future release (with appropriate prior notice or compile warnings)

Yes we manually construct the collective. Should I add some comments to suggest that NoWriteD maybe remoted if D=void is supported?

hwu36:
@kongroo , would you please rebase your code?  sm90 epilogue code changed quite a bit.  there are conflicts.  i tried to solve it myself, but ends up with mismatch results.

kongroo:
@hwu36 I just rebased the code. (but haven't checked the correctness yet)

kongroo:
@hwu36 It looks like version 3.4 has been delayed? Do you know when it will get ready? I'll rebase then

kongroo:
@hwu36 @richardmcai @yzhaiustc @thakkarV  I have added support for passing void to ElementD, and I've also added a unit test, which compares the output of writing D using normal kernel with using AuxStore and void ElementD. Could you please review again?

hwu36 (approved):

PR #1173, opened 2023-11-03
Fix missing return warning for tuple_cat

reed-lau:
image

fix a missing return warning.


hwu36:
@thakkarV

thakkarV:
@hwu36 this is good to merge now

hwu36 (approved):

Issue #1170, opened 2023-11-02
[DOC] CUTLASS 3.3 changelog change request 


manishucsd:
Suggesting a change to the changelog for CUTLASS 3.3

## Current ChangeLog
```md
New [Mixed Precision Hopper GEMMs](https://github.com/NVIDIA/cutlass/blob/main/examples/55_hopper_mixed_dtype_gemm) support covering 16-bit x 8-bit input types with optimal performance.
New [Mixed Precision Ampere GEMMs](https://github.com/NVIDIA/cutlass/commit/7d8317a63e0a978a8dbb3c1fb7af4dbe4f286616) with support for canonical layouts (TN) and {fp16, bf16} x {s8/u8}. They also include fast numeric conversion recipes and warp level shuffles to achieve optimal performance.
```

## Suggested change
```
New [Mixed-Input Hopper GEMMs](https://github.com/NVIDIA/cutlass/blob/main/examples/55_hopper_mixed_dtype_gemm) support covering 16-bit x 8-bit input types with optimal performance.
New [Mixed-Input Ampere GEMMs](https://github.com/NVIDIA/cutlass/pull/1084) with support for canonical layouts (TN) and {fp16, bf16} x {s8/u8}. They also include fast numeric conversion recipes and warp level shuffles to achieve optimal performance.
```

### Summary and rationale for the suggested changes
1. `Mixed-Precision` to `Mixed-Input`. [`Mixed-Precision`](https://arxiv.org/abs/1710.03740) is taken by the GEMM data-type where inputs (DataType(`operandA`) == DataType(`operandB`) are mixed with a different accumulation data type (`F16*F16+F32` and `BF16*BF16+F32`). The code uses [cutlass::arch::OpMultiplyAddMixedInputUpcast](https://github.com/manishucsd/cutlass/blob/main/include/cutlass/arch/mma.h#L73) tag to navigate and communicate that input data types are mixed. It would be good to set a nomenclature that is consistent and distinguishes between `Mixed-Precision` and `Mixed-Input` use-case. 

2. Update the hyperlink for `Mixed Precision Ampere GEMMs` to the [PR#1084](https://github.com/NVIDIA/cutlass/pull/1084) which has detailed description, steps to only compile Ampere mixed-input GEMMs, reproduce performance results, and a performance graph.

mnicely:
We've moved to Mixed "Input"

PR #1084, opened 2023-09-11
Support for Mixed Input TensorOp 

manishucsd:
This PR introduces support for mixed input data types for the NVIDIA A100. The PR is motivated by rising trend of utilizing activations in 16-bit floating-point (either F16 or BF16) alongside weights in 8-bit integer. 

# Supported GEMM datatype configurations
### Upcast on Operand B
- `[F16|F32] <= F16 * S8 + [F16|F32]`
- `[F16|F32] <= F16 * U8 + [F16|F32]`
### Upcast on Operand A
- `[F16|F32] <= S8 * F16 + [F16|F32]`
- `[F16|F32] <= U8 * F16 + [F16|F32]`

### Upcast on Operand B
- `[BF16|F32] <= BF16 * S8 + F32`
- `[BF16|F32] <= BF16 * U8 + F32`
### Upcast on Operand A
- `[BF16|F32] <= S8 * BF16 + F32`
- `[BF16|F32] <= U8 * BF16 + F32`

## Key Features of the Mixed Input Computation:

- **Memory Layout:** TN layout, which implies RowMajor for matrix A and ColumnMajor for matrix B when they're in the Global Memory.
  
- **Mainloop Execution Blueprint:** 
    1. **Data Loading:** Copies data from Global Memory to Shared Memory (using either `cp.async` [or `cp.async.bulk`]. `cp.async`(A100 version) is implemented in this PR).
    2. **Shared Memory to Register:** Loads data from Shared Memory to Registers through the `ldmatrix` instruction.
    3. **Layout Adaptation for `mma` Operation:** 
        - Incorporates the `FragmentShuffler` component, which efficiently reorganizes data within a warp to align with the thread-value mapping needed by the `mma` operation. 
        - Given that `ldmatrix` operates on 8-bit values (`S8` or `U8`) and `mma.sync` works on 16-bit values (`F16` or `BF16`), there's a mismatch in thread ownership post `ldmatrix`. The data isn't aligned with what `mma.sync.[f16|f32].f16.f16.[f16|f32]` or `mma.sync.f32.bf16.bf16.f32` expects. The `FragmentShuffler` addresses this challenge. As all necessary data is within the warp (32 threads), `FragmentShuffler` shuffles it to achieve a thread layout that conforms to `mma.sync`.
[...]

PR #1180, opened 2023-11-08
Doc only change to CUTLASS 3.3 changelog

manishucsd:
This doc-only PR addresses the discussion at #1170 

### Summary and rationale for the suggested changes
1. `Mixed-Precision` to `Mixed-Input`. [`Mixed-Precision`](https://arxiv.org/abs/1710.03740) is taken by the GEMM data-type where inputs (DataType(`operandA`) == DataType(`operandB`) are mixed with a different accumulation data type (`F16*F16+F32` and `BF16*BF16+F32`). The code uses [cutlass::arch::OpMultiplyAddMixedInputUpcast](https://github.com/manishucsd/cutlass/blob/main/include/cutlass/arch/mma.h#L73) tag to navigate and communicate that input data types are mixed. It would be good to set a nomenclature that is consistent and distinguishes between `Mixed-Precision` and `Mixed-Input` use-case. 

2. Update the hyperlink for `Mixed Precision Ampere GEMMs` to the [PR#1084](https://github.com/NVIDIA/cutlass/pull/1084) which has detailed description, steps to only compile Ampere mixed-input GEMMs, reproduce performance results, and a performance graph.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,16 +1,16 @@
 # NVIDIA CUTLASS Changelog
 
 ## [3.3](https://github.com/NVIDIA/cutlass/releases/tag/v3.3) (2023-10-31)
-* [Mixed Precision Hopper GEMMs](/examples/55_hopper_mixed_dtype_gemm) support covering 16-bit x 8-bit input operand types.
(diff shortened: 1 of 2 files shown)


hwu36:
could you pleae also change CHANGELOG.md?

manishucsd:
> could you pleae also change CHANGELOG.md?

done

PR #1187, opened 2023-11-14
fix cp.async L2 prefetch typo

reed-lau:
fix a typo for cp.async L2 prefetch.

And I also found that, the copy_traits will always forward the SM80_CP_ASYNC_CACHEGLOBAL to SM80_CP_ASYNC_CACHEGLOBAL_ZFILL(https://github.com/NVIDIA/cutlass/blob/main/include/cute/atom/copy_traits_sm80.hpp#L57, https://github.com/NVIDIA/cutlass/blob/main/include/cute/atom/copy_traits_sm80.hpp#L79).
Thus the SM80_CP_ASYNC_CACHEGLOBAL's implementation will never be called any more, should we delete it?


hwu36:
@thakkarV  for the question above

thakkarV:
I don't think we should delete the unused ones, as someone else may need them at some point. but this is indeed a typo.

reed-lau:
Hi, @hwu36 could you help merge it? or do I need more modification?

hwu36 (approved):

PR #1189, opened 2023-11-15
Add support for sparse GEMM with visitor epilogue

alexsamardzic:
@apuaaChen

### DIFF
--- a/examples/15_ampere_sparse_tensorop_gemm/CMakeLists.txt
+++ b/examples/15_ampere_sparse_tensorop_gemm/CMakeLists.txt
@@ -33,3 +33,8 @@ cutlass_example_add_executable(
   ampere_sparse_tensorop_gemm.cu
   )
 
+cutlass_example_add_executable(
+  15_ampere_sparse_tensorop_gemm_with_visitor
+  ampere_sparse_tensorop_gemm_with_visitor.cu
+  )
+
--- a/examples/15_ampere_sparse_tensorop_gemm/ampere_sparse_tensorop_gemm_with_visitor.cu
+++ b/examples/15_ampere_sparse_tensorop_gemm/ampere_sparse_tensorop_gemm_with_visitor.cu
@@ -1,0 +1,374 @@
+/***************************************************************************************************
+ * Copyright (c) 2017 - 2023 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: BSD-3-Clause
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ * list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ *
+ * 3. Neither the name of the copyright holder nor the names of its
+ * contributors may be used to endorse or promote products derived from
+ * this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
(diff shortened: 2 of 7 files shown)


alexsamardzic:
This PR makes it possible to combine sparse GEMM with EVT epilogues.  Besides adding `SparseGemmWithVisitor` and `DefaultSparseGemmWithVisitor` classes to `cutlass::gemm::kernel` namespace, as well as `SparseGemmWithVisitor` class to `cutlass::gemm::device` namespace, it also includes an example of use, modeled after existing `15_ampere_sparse_tensorop_gemm`, but with bias addition implemented as an EVT epilogue.

Some items to discuss about PR:

1. I'm not sure that I understand properly the purpose of the first argument of `cutlass::epilogue::threadblock::EpilogueWithVisitorCallbacks` template, so instantiation of this template in `include/cutlass/gemm/kernel/default_gemm_sparse_with_visitor.h` may need an adjustment.  In particular, I find it confusing that `ElementC`, `LayoutC` and `ElementEpilogue` are to be specified here, when these have no clear meaning for the EVT epilogue.
2. Tried first to go without `cutlass::gemm::device::SparseGemmWithVisitor` class, and to utilize `SparseGemmWithVisitor` kernel from `GemmUniversal` device class instead, but to no avail; would appreciate comments on is it possible and how actually to use `SparseGemm`/`SparseGemmWithVisitor` kernels this way.
3. For intended use of `SparseGemmWithVisitor` in my case, and that is about supporting sparse GEMM on tensor cores in PyTorch project, it is imprortant too to have it exposed through `cutlass_library`, so I would also appreciate any suggestions on how to do it.  In particular, `cutlass::gemm::device::SparseGemmWithVisitor` template arguments differ from corresponding `cutlass::gemm::device::SparseGemm` class (for example, `ElementC`/`LayoutC` arguments are missing in my implementation of `SparseGemmWithVisitor` counterpart, and I'm not sure is that going to be a problem).

hwu36:
@apuaaChen @jackkosaian @richardmcai , could you please review it first?  we can include it in 3.4

apuaaChen:
Hi @alexsamardzic Thank you for your effort! This MR looks good to me. My small suggestions are listed above, the basic idea is to use the original non-visitor implementation from CUTLASS as the base class and derive anything unrelated to epilogue from them. This could avoid some duplications.

alexsamardzic:
Hi @apuaaChen, thanks for your comments.  Just to make it clear, when you mention derivation in your first comment, do you have `cutlass::gemm::kernel` or `cutlass::gemm::device` classes on your mind (or both)?

apuaaChen:
> Hi @apuaaChen, thanks for your comments. Just to make it clear, when you mention derivation in your first comment, do you have `cutlass::gemm::kernel` or `cutlass::gemm::device` classes on your mind (or both)?

Hi @alexsamardzic, I think both `kernel` and `device` should be benefited from derivation.

alexsamardzic:
Sorry for the delay, I just pushed a commit that added `ElementC`/`LayoutC` into visitor version of sparse GEMM, and also refactored classes within `cutlass/gemm/kernel` so that the relation between non-visitor and visitor version is alike to relation between corresponding non-visitor and visitor universal GEMM classes (which means: visitor version inherits non-visitor version, base `Params` class is extracted into separate file etc.).  So - it would be great if you could check it now (please look [here](https://github.com/NVIDIA/cutlass/pull/1189/files), i.e. look at these two commits as whole, as I'm going to squash them if this looks OK), and let me know your comments.

apuaaChen:
> Sorry for the delay, I just pushed a commit that added `ElementC`/`LayoutC` into visitor version of sparse GEMM, and also refactored classes within `cutlass/gemm/kernel` so that the relation between non-visitor and visitor version is alike to relation between corresponding non-visitor and visitor universal GEMM classes (which means: visitor version inherits non-visitor version, base `Params` class is extracted into separate file etc.). So - it would be great if you could check it now (please look [here](https://github.com/NVIDIA/cutlass/pull/1189/files), i.e. look at these two commits as whole, as I'm going to squash them if this looks OK), and let me know your comments.

Hi @alexsamardzic. I went through the updated code, it looks much better now. Two minor comments:

- For the kernel-level template, you can take a look at https://github.com/NVIDIA/cutlass/blob/main/include/cutlass/gemm/kernel/default_gemm_universal_with_visitor.h, which can further simplify your code
- I like the `params_sparse_base.h`, which helps avoid code duplication. But I'm not sure if it is ok to change the `params` struct of `sparse_gemm.h`, maybe @hwu36 can give more comments here

alexsamardzic:
Thanks for checking latest changes.

>     * For the kernel-level template, you can take a look at https://github.com/NVIDIA/cutlass/blob/main/include/cutlass/gemm/kernel/default_gemm_universal_with_visitor.h, which can further simplify your code

I've added `cutlass/gemm/kernel/default_gemm_sparse_with_visitor.h`, I think it's even simpler than corresponding `default_gemm_universal_with_visitor.h` - namely, I'm skipping `using GemmBase = ...` and am building my `BaseEpilogue` directly.  Do you have anything specific that you want changed here?

>     * I like the `params_sparse_base.h`, which helps avoid code duplication. But I'm not sure if it is ok to change the `params` struct of `sparse_gemm.h`, maybe @hwu36 can give more comments here

Ok.  Just a comment here:  The `SparseGemm::Params` is exactly the same as before, it's just that what it shares with `SparseGemmWithEpilogueVisitor::Params` is now extracted into common base class.  This is, again, exactly how `GemmUniversal::Params`/`GemmWithEpilogueVisitor::Params` work (in their case, there is even some duplication actually).

alexsamardzic:
FWIW, [here](https://github.com/NVIDIA/cutlass/files/13601599/sparse-patch.txt) is a patch to be applied on top of f87bf8ae0a0b56a8a033335bc45c2f24053a640f, that makes both `cutlass::gemm::kernel::SparseGemm` and `cutlass::gemm::kernel::SparseGemmWithVisitor` inheriting from common base class (instead of later inheriting from former, that I posted previously).  This may be also an approach to consider.

hwu36:
fyi,  i am working on this one now.

alexsamardzic:
> fyi, i am working on this one now.

Thanks.  Just to clarify - by "working on this", do you mean reviewing my PR or working on this feature independently?

hwu36:
i mean working on merging your MR.  you know there is a process that i have to go through.  if i need to change your code, i can do it directly in this pr.

alexsamardzic:
Of course, thanks!   Please let me know if further changes are needed from my side, in particular w.r.t. @apuaaChen request to refactor the code so that duplication between existing sparse GEMM and newly added sparse GEMM with visitor epilogue.

Also, it was my intention to check is there anything needed to add support for sparse GEMM with visitor epilogue into `cutlass_library` - this is something that I also need for my PyTorch work.  But I can do that through a new PR after this one merged.

hwu36 (approved):

PR #1218, opened 2023-11-29
fix(type_traits): fix traits for 64 bit int

JieRen98:


hwu36:
cublas does not support `int64_t` gemm, it does not have `CUDA_R_64I`

hwu36:
> shall we still update the corresponding cuBLAS dtype?

I don't know what is the best solution here.  maybe just write a comment here?

Skylion007:
> > shall we still update the corresponding cuBLAS dtype?
> 
> I don't know what is the best solution here. maybe just write a comment here?

Was int64 supported added in CUBLAS12? https://docs.nvidia.com/cuda/cublas/index.html#bit-integer-interface

PR #1224, opened 2023-12-01
Fix Stream-K reduce bug in epilogue with broadcast

alihassanijr:
Fixes an issue that existed since Stream-K support was added to epilogue with broadcast in #892 .

`reduce` [is supposed to offset destination iterator by the reduction fragment index](https://github.com/NVIDIA/cutlass/blob/main/include/cutlass/epilogue/threadblock/epilogue.h#L381-L382), not by 1 like the typical epilogue call. 

@jackkosaian @hwu36


jackkosaian:
LGTM. Thanks!

hwu36 (approved):

PR #1225, opened 2023-12-01
Fixed illegal PTX syntax

hwu36:


kerrmudgeon (approved):

include/cutlass/arch/barrier.h
```diff
@@ -267,13 +267,12 @@ struct ClusterBarrier {
   static void arrive(ValueType const* smem_ptr) {
 #if CUDA_BARRIER_ENABLED
     uint32_t smem_addr = cute::cast_smem_ptr_to_uint(smem_ptr);
-    uint64_t state = 0;
     asm volatile(
         "{\n\t"
-        "mbarrier.arrive.shared.b64 %1, [%0];\n\t"
+        "mbarrier.arrive.shared.b64 _, [%0];\n\t"
```
manishucsd, line 272: What does `_` mean here? Why can't it be just `mbarrier.arrive.shared.b64 [%0];\n\t`

KTong821: See https://docs.nvidia.com/cuda/parallel-thread-execution/#parallel-synchronization-and-communication-instructions-mbarrier-arrive

PR #1275, opened 2023-12-19
Allow per-column bias in EpilogueTensorBroadcast

alihassanijr:
EpilogueTensorBroadcast only supports per-row vector broadcast, because the bias stride is hardcoded.

It can easily support both if the bias stride is made conditional, and the original behavior is maintained by defaulting to per-row.

Existing unit tests with `EpilogueTensorBroadcast` pass, and equivalent tests with per-col bias have been added.

### DIFF
--- a/include/cutlass/epilogue/collective/epilogue_tensor_broadcast.hpp
+++ b/include/cutlass/epilogue/collective/epilogue_tensor_broadcast.hpp
@@ -69,7 +69,8 @@ template <
   class StrideC_,
   class StrideD_,
   class ThreadEpilogueOp_,
-  class EpilogueSchedule_
+  class EpilogueSchedule_,
+  bool PerColumnBias_ = false
 >
 class EpilogueTensorBroadcast {
 public:
@@ -100,6 +101,9 @@ public:
   static constexpr bool IsBinaryOp0Enabled = ThreadEpilogueOp::IsBinaryOp0Enabled;
   static constexpr bool IsBinaryOp1Enabled = ThreadEpilogueOp::IsBinaryOp1Enabled;
   static constexpr bool IsUnaryOpEnabled = ThreadEpilogueOp::IsUnaryOpEnabled;
+
+  static constexpr bool PerColumnBias = PerColumnBias_;
+  using BiasStride = typename std::conditional<PerColumnBias, Stride<_0, _1, _0>, Stride<_1, _0, _0>>::type;
 
   struct SharedStorage { };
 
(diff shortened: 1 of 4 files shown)


include/cutlass/epilogue/collective/epilogue_tensor_broadcast.hpp
```diff
@@ -101,6 +102,9 @@ class EpilogueTensorBroadcast {
   static constexpr bool IsBinaryOp1Enabled = ThreadEpilogueOp::IsBinaryOp1Enabled;
   static constexpr bool IsUnaryOpEnabled = ThreadEpilogueOp::IsUnaryOpEnabled;
 
+  static constexpr bool PerColumnBias = PerColumnBias_;
+  using BiasStride = typename std::conditional<PerColumnBias, Stride<_0, _1, _0>, Stride<_1, _0, _0>>::type;
```
jackkosaian, line 106: We often prefer to use the `cute::` or `platform::` variants of things like `std::contditional` to work better with NVRTC.
```suggestion
  using BiasStride = cute::conditional_t<PerColumnBias, Stride<_0, _1, _0>, Stride<_1, _0, _0>>;
```

test/unit/gemm/device/gemm_testbed_3x_tensor_broadcast.hpp
```diff
@@ -263,7 +266,9 @@ struct Testbed3xTensorBroadcast {
         decltype(dummy_Aux),      
         decltype(dummy_Valpha),
         decltype(dummy_Vbeta),
-        ActivationFunctor> epilogue_params{
+        ActivationFunctor,
+        cutlass::plus<ElementCompute>,
+        /* PerColBias = */ PerColBias> epilogue_params{
```
jackkosaian, line 271: This inline comment seems unnecessary.
```suggestion
        PerColBias> epilogue_params{
```

test/unit/gemm/device/sm90_gemm_f16_f16_f16_tensor_op_f32_tensor_broadcast.cu
```diff
@@ -292,6 +292,253 @@ TEST(SM90_Device_Gemm_f16t_f16t_f32n_tensor_op_gmma_f32_warpspecialized_tensor_b
   EXPECT_TRUE(test::gemm::device::TestAllTensorBroadcast<Gemm>());
 }
 
+/////////////////////////////////////////////////////////////////////////////////////////////////
+///////////////////////////////////// per-column bias ///////////////////////////////////////////
+/////////////////////////////////////////////////////////////////////////////////////////////////
```
jackkosaian, line 297: It seems like this PR adds versions of all of the original tests, but with `PerColBias = true`. Would it be possible to reduce the number of tests added so that test count in these files doesn't double?

jackkosaian (changes requested):
Thank you for your contribution.

test/unit/gemm/device/sm90_gemm_f16_f16_f16_tensor_op_f32_tensor_broadcast.cu
alihassanijr: Thank you; yeah I just figured I'd be thorough, but most are just unnecessary. I removed everything but the single fp32 test, and applied the rest of the suggestions.

PR #1278, opened 2023-12-25
Fix flops calculation and tensor b stride calculation in the example 36

getianao:
The count of FMAs depends on the index-tensor-B rather than the entire tensor B. 
Additionally, there was an error in using 'n' instead of 'k' when determining the width of tensor B in the calculation of the stride.

### DIFF
--- a/examples/36_gather_scatter_fusion/gather_scatter_fusion.cu
+++ b/examples/36_gather_scatter_fusion/gather_scatter_fusion.cu
@@ -174,7 +174,7 @@ struct Options {
   double gflops(double runtime_s) const {
 
     // Number of real-valued multiply-adds
-    int64_t fmas = problem_size.m() * index_size * problem_size.k();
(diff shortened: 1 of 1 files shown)


examples/36_gather_scatter_fusion/gather_scatter_fusion.cu
```diff
@@ -173,8 +173,8 @@ struct Options {
   /// Compute performance in GFLOP/s
   double gflops(double runtime_s) const {
 
-    // Number of real-valued multiply-adds 
-    int64_t fmas = problem_size.product();
+    // Number of real-valued multiply-adds
+    int64_t fmas = int64_t(index_size) * problem_size.n() * problem_size.k();
```
hwu36, line 177: should it be 
`m * index_size * k` ?

getianao: fixed

hwu36 (approved):

PR #1287, opened 2023-12-31
expose stream API in python kernel call interfaces

K-Wu:
Hi there,

I exposed the stream as an option in the python kernel call interfaces. This is needed for my research project at U of I. Please let me know if this could be merged into your main branch with shepherding. 

Thank you.

Best Regards,
Kun

### DIFF
--- a/python/cutlass/op/conv.py
+++ b/python/cutlass/op/conv.py
@@ -131,6 +131,7 @@ from cutlass.backend.library import TensorDescription, TileDescription
 from cutlass.op.op import OperationBase
 from cutlass.shape import Conv2DProblemSize, MatrixCoord
 from cutlass.utils import check, datatypes
+from cuda import cuda
 
 
 class Conv2d(OperationBase):
@@ -733,7 +734,8 @@ class Conv2d(OperationBase):
             stride=(1, 1), padding=(0, 0), dilation=(1, 1),
             alpha=None, beta=None,
             split_k=("serial", 1), sync: bool = True,
-            print_module: bool = False) -> Conv2dArguments:
+            print_module: bool = False,
+            stream: cuda.CUstream = cuda.CUstream(0)) -> Conv2dArguments:
         """
         Runs the kernel currently specified. If it has not already been, the kernel is emitted and
         compiled. Tensors holding operands and outputs of the kernel are sourced either from the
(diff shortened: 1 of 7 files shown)


jackkosaian:
Thanks for your contribution, @K-Wu. I'll work on testing this out and will get back to you with any feedback.

python/cutlass/op/conv.py
```diff
@@ -733,7 +734,8 @@ def run(self, A=None, B=None, C=None, D=None,
             stride=(1, 1), padding=(0, 0), dilation=(1, 1),
             alpha=None, beta=None,
             split_k=("serial", 1), sync: bool = True,
-            print_module: bool = False) -> Conv2dArguments:
+            print_module: bool = False,
+            stream: cuda.CUstream = cuda.CUstream(0)) -> Conv2dArguments:
```
jackkosaian, line 738: Could you please update the docstrings for this and other methods that now take in `stream` as an argument?

K-Wu: Hi Jack, sure! This is addressed in commit 91becf9022ce9235d8af6fe7c4e648e6f7e8cb96.

K-Wu:
Hi @jackkosaian, thank you for your review! I have addressed your comment in commit 91becf9022ce9235d8af6fe7c4e648e6f7e8cb96.

Let me know if you need any further changes/clarifications. Thank you again.

jackkosaian:
Could you also add the following inside the `GemmGroupedArguments.__init__` in `python/cutlass/backend/gemm_operation.py`?

```python
        self.stream = kwargs.get("stream", cuda.CUstream(0))
```

This is needed because `GemmGroupedArguments` do not derive from `ArgumentsBase`, and, thus, do not have the `stream` member set.

K-Wu:
Nice catch! I have addressed it in the new commit.

> Could you also add the following inside the `GemmGroupedArguments.__init__` in `python/cutlass/backend/gemm_operation.py`?
> 
> ```python
>         self.stream = kwargs.get("stream", cuda.CUstream(0))
> ```
> 
> This is needed because `GemmGroupedArguments` do not derive from `ArgumentsBase`, and, thus, do not have the `stream` member set.

hwu36 (approved):

PR #1303, opened 2024-01-16
fix unrecognized print format specifier for int8/uint8

reed-lau:
when I debug an int8 kernel, I found the print(int8_tensor) could not perfom as expected(show in the following picture). Although it's a feature support problem for nvcc. I think we can work around it first.
The following snippet can reproduce the problem with nvcc-11.4 and nvcc-12.1.

```cpp
#include <cute/tensor.hpp>

__global__ void foo() {
  using namespace cute;

  int8_t i8 = 1;
  int16_t i16 = 1;
  int32_t i32 = 1;

  uint8_t u8 = 1;
  uint16_t u16 = 1;
  uint32_t u32 = 1;

  short s = 1;
  unsigned short us = 1;

  print(i8); print("\n");
  print(i16); print("\n");
  print(i32); print("\n");

  print(u8); print("\n");
  print(u16); print("\n");
  print(u32); print("\n");

  print(s); print("\n");
  print(us); print("\n");
}

int main() {
  foo<<<1, 1>>>();
  cudaDeviceSynchronize();
}
```
image

### DIFF
--- a/include/cute/util/print.hpp
+++ b/include/cute/util/print.hpp
@@ -70,13 +70,13 @@ print(char c) {
 CUTE_HOST_DEVICE
 void
 print(signed char a) {
-  printf("%hhd", a);
+  printf("%d", (int)(a));
 }
 
 CUTE_HOST_DEVICE
 void
 print(unsigned char a) {
-  printf("%hhu", a);
+  printf("%u", (unsigned int)(a));
 }
 
 CUTE_HOST_DEVICE
--- a/include/cute/util/print.hpp
+++ b/include/cute/util/print.hpp
@@ -70,13 +70,13 @@ print(char c) {
 CUTE_HOST_DEVICE
 void
 print(signed char a) {
-  printf("%d", (int)(a));
+  printf("%d", static_cast<int>(a));
 }
 
 CUTE_HOST_DEVICE
 void
 print(unsigned char a) {
-  printf("%u", (unsigned int)(a));
+  printf("%u", static_cast<unsigned int>(a));
 }
 
 CUTE_HOST_DEVICE


include/cute/util/print.hpp
```diff
@@ -70,13 +70,13 @@ print(char c) {
 CUTE_HOST_DEVICE
 void
 print(signed char a) {
-  printf("%hhd", a);
+  printf("%d", (int)(a));
```
thakkarV, line 73: we do not use c style casts in our code. can you please convert these to static casts instead?

reed-lau: done

reed-lau:
@thakkarV could you help review it?

thakkarV:
looks good, I will work on integrating this

thakkarV:
does this problem exist for signed/unsigned short as well?
```
CUTE_HOST_DEVICE
void
print(short a) {
  printf("%hd", a);
}

CUTE_HOST_DEVICE
void
print(unsigned short a) {
  printf("%hu", a);
}
```

reed-lau:
> does this problem exist for signed/unsigned short as well? Can you try out those two as well and fix them in this MR if need be?
> 
> ```
> CUTE_HOST_DEVICE
> void
> print(short a) {
>   printf("%hd", a);
> }
> 
> CUTE_HOST_DEVICE
> void
> print(unsigned short a) {
>   printf("%hu", a);
> }
> ```

It seems like NVCC recognizes the short format specifier, as shown in the picture I attached at the very beginning.

reed-lau:
@hwu36 can we merge it now？

thakkarV:
@hwu36 good to merge now

hwu36 (approved):

reed-lau:
@hwu36  and this one.

PR #1384, opened 2024-03-06
Support use of external/system GTest installation

iskunk:
I am building the project in a restricted environment that does not allow network access (so FetchContent is not usable), and I already have a pre-compiled build of GTest ready to go. With this change, I can pass `-DGTest_ROOT=/path/to/gtest-install` to CMake, and the tests will build using the specified install. A system install of GTest will also be picked up. If neither is available, the configuration falls back to FetchContent.

The second commit fixes a related issue. I was running into this problem:
```
$ make test ARGS="--exclude-regex fmha_backward_python"
make[1]: Entering directory '.../cutlass-3.4.1-build'
Running tests...
Test project .../cutlass-3.4.1-build
                    Start   1: ctest_profiler_gemm
Failed to change working directory to ./bin : No such file or directory
  1/176 Test   #1: ctest_profiler_gemm ....................................***Not Run   0.00 sec
                    Start   2: ctest_profiler_conv2d
Failed to change working directory to ./bin : No such file or directory
  2/176 Test   #2: ctest_profiler_conv2d ..................................***Not Run   0.00 sec
                    Start   3: ctest_profiler_conv3d
Failed to change working directory to ./bin : No such file or directory
[...]
176/176 Test #176: ctest_unit_nvrtc_thread ................................***Not Run   0.00 sec

0% tests passed, 174 tests failed out of 174

Total Test time (real) =   0.03 sec
```

I've observed that the `./bin` directory is created when building from a default configuration, but once I customized the build to my needs, it was absent. It's not clear where this is coming from, so I added an explicit `MAKE_DIRECTORY` to ensure that the tests will run.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -689,7 +689,10 @@ include(CTest)
 enable_testing()
 
 if (CUTLASS_ENABLE_GTEST_UNIT_TESTS)
-  include(${CMAKE_CURRENT_SOURCE_DIR}/cmake/googletest.cmake)
+  find_package(GTest QUIET)
+  if (NOT GTest_FOUND)
+    include(${CMAKE_CURRENT_SOURCE_DIR}/cmake/googletest.cmake)
+  endif()
 endif()
 
 if (NOT TARGET test_all)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -737,6 +737,7 @@ set(CUTLASS_DEFAULT_ACTIVE_TEST_SETS "default" CACHE STRING "Default
   with CUTLASS_TEST_SETS environment variable when running the ctest
   executable.")
 
+file(MAKE_DIRECTORY "${CMAKE_BINARY_DIR}/${CMAKE_INSTALL_BINDIR}")
 set(CUTLASS_CTEST_TEMPLATE_FILE ${CMAKE_CURRENT_LIST_DIR}/cmake/CTestTestfile.configure.cmake)
 set(CUTLASS_CTEST_GENERATED_FILES "" CACHE INTERNAL "")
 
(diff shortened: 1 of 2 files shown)


hwu36:
@d-k-b

d-k-b:
Closing for now, we can reopen if the suggestion above would be useful.

iskunk:
Hi @d-k-b, I've rebased the PR. I am unable to reopen the ticket; could you please do so?

Nothing about the need for this issue has changed. It is customary to allow the use of a pre-built dependency, and in my environment, it is required. I currently have to apply a patch with this change in order to build the project.

d-k-b:
Ah, OK, I partially misunderstood the original request. I thought you had already pulled the source locally and just need to point the CUTLASS build at it. The FetchContent steps above I mentioned would work with that. However, if you want to integrate with a pre-built  GTest library, then we should go with something similar to the last snippet I posted above. Using the pre-built library should be opt-in via a CMake option to ensure CUTLASS does not attempt to use an installed library by default.

```
if (CUTLASS_USE_SYSTEM_GOOGLETEST)
  find_package(GTest)
  return()
endif()
```

d-k-b:
Reopening for further investigation.

iskunk:
Thanks @d-k-b. I'll update this to use a `CUTLASS_USE_SYSTEM_GOOGLETEST` flag; that's perfectly reasonable.

iskunk:
All right, I've updated my push accordingly. Please let me know if you need anything more here.

d-k-b:
I see that you made changes in another commit, but I'm guessing this PR won't update for it. Do you mind creating a new PR?

iskunk:
No problem; the new PR is up at #1469.

PR #1350, opened 2024-02-16
Add couple configs into generator.py for mixed input MM

alexsamardzic:
I'm adding (PR [here](https://github.com/pytorch/pytorch/pull/119986)) CUTLASS kernels as an auto-tune option for PyTorch compiler, and it would be nice to have these additional configurations available.  This is not urgent, and more of alike changes may be further desired, so if it's OK to make changes like this then this PR could be kept open for while, and I'll make further additions, as needed, to it.

@manishucsd : Would it make sense for `GenerateSM80_TensorOp_16816_mixed_input_upcast_a` and `GenerateSM80_TensorOp_16816_mixed_input_upcast_b` to be symmetric w.r.t. `math_instructions` and `tile_descriptions`?  I can change it through this PR too.

PR #1413, opened 2024-03-19
Add support for mixed 4-bit/8-bit data types GEMM

alexsamardzic:


alexsamardzic:
More to come here: support for `U4`, support for generator in the CUTLASS library, etc.  Still, opening PR to solicit feedback for `S8`/`S4` and `S4/S8` GEMMs that are now available; in particular, I'm interested in eventual suggestions for a faster approach to `S4`->`S8` conversion.

alexsamardzic:
Added more tests.

test/unit/gemm/device/gemm_universal_s4t_s8n_s32t_mixed_input_tensor_op_s32_sm80.cu
```diff
@@ -0,0 +1,97 @@
...
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ **************************************************************************************************/
+/*! \file
+    \brief Tests for device-wide GEMM interface
+    
+*/
+
+#include <iostream>
```
lezcano, line 36: leftover? (just lurking around)

alexsamardzic: Yep.  But it's there pretty much in all the tests in `test/unit/gemm/device`, it seems we've been just copying it around...  Would be the best to remove all of them in a separate PR.

alexsamardzic:
Added generator support for S8/S4 and S4/S8.

---

AFAIK, implementing generator support for given operation is not specifically documented, so I want to clarify the steps I've taken here.  Basically, I've copied code from `GenerateSM80_TensorOp_16832_TN` method into `GenerateSM80_TensorOp_16832_TN_mixed_input_upcast_(a|b)`, and then made some changes:

- Obviously, I've changed `math_instructions` assignments according to data types actually used for mixed input data types.
- I'm not sure from where `smem_usage = 164` in `GenerateSM80_TensorOp_16832_TN`  comes, and this variable is not further used anyway, so I skipped it in new methods.
- I've used two sets of alignment constraints.  The alignments for operands A and B are the same, but operand C (and thus the result too) could be either 32-bit or 8-bit.  The code at the end of the mixed input methods, within the last `if` statement is handling the later case, and alignments are changed here accordingly.  (Note that for `GenerateSM80_TensorOp_16816_mixed_input_upcast_(a|b)` there are snippets of code at the end of methods doing alike thing, but they're slightly different from each other, and also from what I did here.)
- The `tile_descriptions` were initially copied from `GenerateSM80_TensorOp_16832_TN`, and then I would make sure that all the relevant kernels would be compiled (through adding `CUTLASS_LIBRARY_KERNELS="*i16832gemm*"` to the CMake command line), and would remove tiles that would fail to compile.

I did the verification as @manishucsd suggested [here](https://github.com/NVIDIA/cutlass/pull/1350#issuecomment-1977910775):  As mentioned above, I did the build with all the relevant kernels included, and then I verified that `cutlass_profiler` would run all the tile variations that are specified in `GenerateSM80_TensorOp_16832_TN_mixed_input_upcast_(a|b)`.  Note that the profiler would produce `Disposition: Incorrect` for all the kernels with 8-bit output; I suppose it's related to saturation - I'm not sure if I should actually come up with applying saturation somehow for this combination of input data types?

---

Overall, this PR now contains everything that I intended to do for `S4`/`S8` and `S8`/`S4` GEMM, and it's ready for review.  It has grown somewhat large, so I'd suggest to have it reviewed and eventually merged, and then I can add `U4`/`U8` and `U8`/`U4`, and maybe `U4`/`S8` and `S8`/`U4`, support in follow-up PR(s).

andrewor14:
Hi @alexsamardzic, thanks for working on this. Just wanted to clarify, will this kernel support int4 grouped per channel weight quantization + int8 per token dynamic activation quantization?

alexsamardzic:
> Hi @alexsamardzic, thanks for working on this. Just wanted to clarify, will this kernel support int4 grouped per channel weight quantization + int8 per token dynamic activation quantization?

This kernel is just int4/int8 GEMM, producing int32 (or int8) result.  Quantization is not to be supported by CUTLASS directly, but could be implemented using an EVT epilogue.  In particular, I'm trying to get this feature into CUTLASS mainly in order to have this particular operation supported in PyTorch, with using it along with quantization as primary motivator.

alexsamardzic:
@manishucsd, @hwu36:  Would it be possible for someone to review this PR (and eventually #1350 too)?  These should not be controversial, are needed by PyTorch, and for this one I'd like to proceed with another PR to add other 4-bit/8-bit integer combinations that make sense.

hwu36:
working on it now.

Hongbosherlock:
> > Hi @alexsamardzic, thanks for working on this. Just wanted to clarify, will this kernel support int4 grouped per channel weight quantization + int8 per token dynamic activation quantization?
> 
> This kernel is just int4/int8 GEMM, producing int32 (or int8) result. Quantization is not to be supported by CUTLASS directly, but could be implemented using an EVT epilogue. In particular, I'm trying to get this feature into CUTLASS mainly in order to have this particular operation supported in PyTorch, with using it along with quantization as primary motivator.

Great job! How can I integrate this PR with PyTorch? Are there any example codes available ? @alexsamardzic

alexsamardzic:
> How can I integrate this PR with PyTorch? Are there any example codes available ? @alexsamardzic

The primary motivation for this PR is to have this combination of operands supported by PyTorch, so the integration should be coming soon.

Hongbosherlock:
> > How can I integrate this PR with PyTorch? Are there any example codes available ? @alexsamardzic
> 
> The primary motivation for this PR is to have this combination of operands supported by PyTorch, so the integration should be coming soon.

I'm a beginner with Cutlass, I have on idea how to use my own constructed s4/s8 data to run this GEMM.
Could you please provide an example code for testing this s4/s8 GEMM? like the official example here:
https://github.com/NVIDIA/cutlass/blob/main/examples/55_hopper_mixed_dtype_gemm/README.md

alexsamardzic:
> I'm a beginner with Cutlass, I have on idea how to use my own constructed s4/s8 data to run this GEMM. Could you please provide an example code for testing this s4/s8 GEMM? like the official example here: https://github.com/NVIDIA/cutlass/blob/main/examples/55_hopper_mixed_dtype_gemm/README.md

These changes are not for Hopper, but for Ampere architecture.  The code to run s4/s8 GEMM would be the same as for any other GEMM, for example s8/s8, except that when a GEMM template instantiated, data type and other argument should be specified accordingly.  For some examples of this, see `using Gemm = cutlass::gemm::device::GemmUniversal...` template instantiations in the test cases added by this PR into `test/unit/gemm/device` directory.  As far as your data concerned, s4 data should be provided as two successive values packed into single byte, and that's all.

zkf331:
## Question

Hi @alexsamardzic,

This is an excellent piece of work, making significant contributions to the quantized inference of the entire LLM model. I have some questions while using it and would like to consult you. My code is as follows:

Using PyTorch's C++/CUDA Extension.

matmul.cu

```C++
torch::Tensor matmul_w4a8(const torch::Tensor &A, const torch::Tensor &B) {
  torch::checkAllSameGPU("W4A8Matmul", {{A, "A", 0}, {B, "B", 1}});
  auto M = A.size(0);
  auto N = B.size(0);
  auto K = A.size(1) * 2;  // 4bit packing is on the columns
  auto C = torch::empty({M, N}, torch::dtype(torch::kInt32).device(A.device()));

  using ElementOutput = int32_t;
  using ElementAccumulator = int32_t;

  using Gemm = cutlass::gemm::device::GemmUniversal<
      int8_t,                // ElementA
      cutlass::layout::RowMajor,       // LayoutA
      cutlass::int4b_t,                // ElementB
      cutlass::layout::ColumnMajor,    // LayoutB
      int32_t,                         // ElementOutput
      cutlass::layout::RowMajor,       // LayoutOutput
      int32_t,                         // ElementAccumulator
      cutlass::arch::OpClassTensorOp,  // tag indicating Tensor Cores
      cutlass::arch::Sm80,  // tag indicating target GPU compute architecture
      cutlass::gemm::GemmShape<128, 128, 64>,
      cutlass::gemm::GemmShape<64, 64, 64>,
      cutlass::gemm::GemmShape<16, 8, 32>,
      cutlass::epilogue::thread::LinearCombination<
          ElementOutput, 128 / cutlass::sizeof_bits<ElementOutput>::value,
          ElementAccumulator, ElementAccumulator>,
      cutlass::gemm::threadblock::GemmIdentityThreadblockSwizzle<>,
      4,   // Stages
      16,  // AlignmentA
      32,  // AlignmentB
      cutlass::arch::OpMultiplyAddMixedInputUpcast,
      cutlass::ComplexTransform::kNone,
      cutlass::ComplexTransform::kNone
    >;

  Gemm gemmOp;
  using GemmCoord = cutlass::gemm::GemmCoord;

  cutlass::gemm::GemmUniversalMode mode = cutlass::gemm::GemmUniversalMode::kGemm;
  int batch_count = 1;
  int64_t stride_A = M * K;
  int64_t stride_B = K * N;
  int64_t stride_C = M * N;
  int64_t stride_D = M * N;

  typename Gemm::Arguments arguments{
      mode,
      {static_cast<GemmCoord::Index>(M), static_cast<GemmCoord::Index>(N),
       static_cast<GemmCoord::Index>(K)},
      batch_count,
      {1, 0},
      A.data_ptr<int8_t>(),
      (cutlass::int4b_t *)B.data_ptr<uint8_t>(),
      C.data_ptr<int32_t>(),
      C.data_ptr<int32_t>(),
      stride_A,
      stride_B,
      stride_C,
      stride_D,
      0,
      0,
      0,
      0
      };

  const cudaStream_t stream = at::cuda::getCurrentCUDAStream();

  auto status = gemmOp(arguments, nullptr, stream);

  TORCH_CHECK(status == cutlass::Status::kSuccess,
              cutlassGetStatusString(status))

  return C;
}
```

The code for packing int4 to uint8 is as follows:

```python
def to_int4(x, bits):
    return torch.where(x < 0, x + 2**bits, x)

def pack_int4_to_uint8(x):
    x_int4 = to_int4(x, 4).to(torch.uint8)
    x_uint8 = x_int4[:, 0::2] | (x_int4[:, 1::2] << 4)
    return x_uint8.to(x.device)
```

I want to implement int32 = s8 * s4 in torch, the code is as follows:

```python
input_int8 = torch.randint(-8, 7, (2, 4), dtype=torch.int8, requires_grad=False).to("cuda:0")
weight_int4 = torch.randint(-8, 7, (4, 4), dtype=torch.int8, requires_grad=False).to("cuda:0")

weight_int4_pack_uint8 = pack_int4_to_uint8(weight_int4)

results = ops.matmul_w4a8(input_int8, weight_int4_pack_uint8)
```

However, the results is incorrect. I would like to ask which part of my code has a problem.

alexsamardzic:
On a quick look, your strides may be wrong.

zkf331:
> On a quick look, your strides may be wrong.

Thank you for your prompt reply. I don't know much about this parameter, and I can't find many references. Could you give me some more details? Thank you very much.

Hongbosherlock:
> > On a quick look, your strides may be wrong.
> 
> Thank you for your prompt reply. I don't know much about this parameter, and I can't find many references. Could you give me some more details? Thank you very much.

@zkf331 I think it maybe like this:
```c++
  int64_t stride_A = M * K;
  int64_t stride_B = N * K;
  int64_t stride_C = M * N;
```
I'm working on it too, maybe we can discuss on WeChat( hongbosherlock)

Hongbosherlock:
> > I'm a beginner with Cutlass, I have on idea how to use my own constructed s4/s8 data to run this GEMM. Could you please provide an example code for testing this s4/s8 GEMM? like the official example here: https://github.com/NVIDIA/cutlass/blob/main/examples/55_hopper_mixed_dtype_gemm/README.md
> 
> These changes are not for Hopper, but for Ampere architecture. The code to run s4/s8 GEMM would be the same as for any other GEMM, for example s8/s8, except that when a GEMM template instantiated, data type and other argument should be specified accordingly. For some examples of this, see `using Gemm = cutlass::gemm::device::GemmUniversal...` template instantiations in the test cases added by this PR into `test/unit/gemm/device` directory. As far as your data concerned, s4 data should be provided as two successive values packed into single byte, and that's all.

I have two s4 values packed in a single byte(uint8). Do I need to unpack the uint8 data to get s4 data before GEMM manually?

alexsamardzic:
> > > I'm a beginner with Cutlass, I have on idea how to use my own constructed s4/s8 data to run this GEMM. Could you please provide an example code for testing this s4/s8 GEMM? like the official example here: https://github.com/NVIDIA/cutlass/blob/main/examples/55_hopper_mixed_dtype_gemm/README.md
> > 
> > 
> > These changes are not for Hopper, but for Ampere architecture. The code to run s4/s8 GEMM would be the same as for any other GEMM, for example s8/s8, except that when a GEMM template instantiated, data type and other argument should be specified accordingly. For some examples of this, see `using Gemm = cutlass::gemm::device::GemmUniversal...` template instantiations in the test cases added by this PR into `test/unit/gemm/device` directory. As far as your data concerned, s4 data should be provided as two successive values packed into single byte, and that's all.
> 
> I have two s4 values packed in a single byte(uint8). Do I need to unpack the uint8 data to get s4 data before GEMM manually?

No, s4 values should be packed, two values per byte.

Hongbosherlock:
https://github.com/NVIDIA/FasterTransformer/blob/df4a7534860137e060e18d2ebf019906120ea204/src/fastertransformer/kernels/cutlass_kernels/int8_gemm/int8_gemm_template.h#L142

Hongbosherlock:
> > I have two s4 values packed in a single byte(uint8). Do I need to unpack the uint8 data to get s4 data before GEMM manually?
> 
> No, s4 values should be packed, two values per byte.

Thanks for your help !  I can get correct result now.  but I have another question:
Assuming that  A is int8 and (M, K),  B is int4 and (K, N), after GEMM:` C = A·B`, and C will be (M, N). Now, I have another matrix E, which is fp32. I want to perform **element-wise multiplication** : `E * C`. Can I complete this element-wise multiplication within the this s4/s8 GEMM operation ?  for example by passing matrix E to`Arguments`?  I am not sure how to do this.
Maybe here is an example what I want to do :https://github.com/NVIDIA/TensorRT-LLM/blob/5d8ca2faf74c494f220c8f71130340b513eea9a9/cpp/tensorrt_llm/kernels/cutlass_kernels/int8_gemm/int8_gemm_template.h#L131

alexsamardzic:
> Assuming that A is int8 and (M, K), B is int4 and (K, N), after GEMM:` C = A·B`, and C will be (M, N). Now, I have another matrix E, which is fp32 and also (M,N). I want to perform **element-wise multiplication** : `E * C`. Can I complete this element-wise multiplication within the this s4/s8 GEMM operation ? for example by passing matrix E to`Arguments`?

If matrix `E` is really `MxN` (i.e. not broadcasted), it doesn't seem that the code you linked is doing this exact operation.  I'd say the simplest way to achieve this would be through EVT epilogues, these are exactly for the purpose of fusing matrix multiplications with arbitrary operations.  For Ampere, there is `examples/47_ampere_gemm_universal_streamk/ampere_gemm_universal_streamk_broadcast.cu` example demonstrating how to use EVT epilogues, you'd have to remove everything related to `Bias`/`C1` matrices in this example, to use `C2` as your matrix `E`, and then to replace `cutlass::plus` with `cutlass::multiplies` in `using Compute2 = ...` (also, you should take care that all of the data types in the template instantiations are correctly specified).

Hongbosherlock:
> > Assuming that A is int8 and (M, K), B is int4 and (K, N), after GEMM:` C = A·B`, and C will be (M, N). Now, I have another matrix E, which is fp32 and also (M,N). I want to perform **element-wise multiplication** : `E * C`. Can I complete this element-wise multiplication within the this s4/s8 GEMM operation ? for example by passing matrix E to`Arguments`?
> 
> If matrix `E` is really `MxN` (i.e. not broadcasted), it doesn't seem that the code you linked is doing this exact operation. I'd say the simplest way to achieve this would be through EVT epilogues, these are exactly for the purpose of fusing matrix multiplications with arbitrary operations. For Ampere, there is `examples/47_ampere_gemm_universal_streamk/ampere_gemm_universal_streamk_broadcast.cu` example demonstrating how to use EVT epilogues, you'd have to remove everything related to `Bias`/`C1` matrices in this example, to use `C2` as your matrix `E`, and then to replace `cutlass::plus` with `cutlass::multiplies` in `using Compute2 = ...` (also, you should take care that all of the data types in the template instantiations are correctly specified).

Thanks, I’m trying this, but it’s not going well currently.
To make it clearer, what I want to do is exactly the following：
```
    // inputs
    //     A           [M, K]    int8
    //     B           [N, K]    int4
    //     alphaCol    [M, 1]    fp32
    //     alphaRow    [1, N]    fp32
    // outputs
    //     mat [M, N]            fp32
```
That is:` (alphaCol x alphaRow) * (A x B)`
 I think here is a s8/s8 example(A and B are all int8):https://github.com/NVIDIA/TensorRT-LLM/blob/5d8ca2faf74c494f220c8f71130340b513eea9a9/cpp/tensorrt_llm/kernels/cutlass_kernels/int8_gemm/int8_gemm_template.h#L131,  which also uses EVT, and the inputs are passed from [here]( https://github.com/NVIDIA/TensorRT-LLM/blob/5d8ca2faf74c494f220c8f71130340b513eea9a9/cpp/tensorrt_llm/plugins/smoothQuantGemmPlugin/smoothQuantGemmPlugin.cpp#L225)
I wonder if I could use the same EVT code and  ` using Gemm = cutlass::gemm::device::GemmUniversalBaseCompat<GemmKernel>` with this s4/s8 GEMM.

alexsamardzic:
> Thanks, I’m trying this, but it’s not going well currently. To make it clearer, what I want to do is exactly the following：
> 
> ```
>     // inputs
>     //     A           [M, K]    int8
>     //     B           [N, K]    int4
>     //     alphaCol    [M, 1]    fp32
>     //     alphaRow    [1, N]    fp32
>     // outputs
>     //     mat [M, N]            fp32
> ```
> 

Well, that's not element-wise multiplication with `MxN` tensor, as stated initially...  The example from TensorRT-LLM that you're linking to is not using EVT, but some kind of their CUTLASS extension instead (probably because CUTLASS had no EVT support at that time).  I don't have time to look into these, so I can only point you again to `examples/47_ampere_gemm_universal_streamk/ampere_gemm_universal_streamk_broadcast.cu` example, but now you should take a look into how `Bias` is applied, and ignore everything `C1`/`C2` related.  With `cutlass::plus` replaced with `cutlass::multiplies`, your `alphaCol` will be applied the same way, and once you understand how this works, it should be easy to interchange offsets and apply `alphaRow` accordingly too.

Hongbosherlock:
> > Thanks, I’m trying this, but it’s not going well currently. To make it clearer, what I want to do is exactly the following：
> > ```
> >     // inputs
> >     //     A           [M, K]    int8
> >     //     B           [N, K]    int4
> >     //     alphaCol    [M, 1]    fp32
> >     //     alphaRow    [1, N]    fp32
> >     // outputs
> >     //     mat [M, N]            fp32
> > ```
> 
> Well, that's not element-wise multiplication with `MxN` tensor, as stated initially... The example from TensorRT-LLM that you're linking to is not using EVT, but some kind of their CUTLASS extension instead (probably because CUTLASS had no EVT support at that time). I don't have time to look into these, so I can only point you again to `examples/47_ampere_gemm_universal_streamk/ampere_gemm_universal_streamk_broadcast.cu` example, but now you should take a look into how `Bias` is applied, and ignore everything `C1`/`C2` related. With `cutlass::plus` replaced with `cutlass::multiplies`, your `alphaCol` will be applied the same way, and once you understand how this works, it should be easy to interchange offsets and apply `alphaRow` accordingly too.

I got errors with `ElementB = cutlass::int4b_t` when I tried to follow this [example](https://github.com/NVIDIA/cutlass/blob/main/examples/47_ampere_gemm_universal_streamk/ampere_gemm_universal_streamk_broadcast.cu#L241):
```
    using EVTKernelStreamK =
        typename cutlass::gemm::kernel::DefaultGemmWithVisitor<
        ElementA, LayoutA, cutlass::ComplexTransform::kNone, AlignmentA,
        ElementB, LayoutB, cutlass::ComplexTransform::kNone, AlignmentB,
        ElementC, LayoutC, AlignmentC,
        ElementAccumulator,
        ElementCompute,
        cutlass::arch::OpClassTensorOp,
        cutlass::arch::Sm80,
        ThreadblockShape,
        WarpShape,
        InstructionShape,
        EVTD,
        cutlass::gemm::threadblock::ThreadblockSwizzleStreamK,
        NumStages,
        cutlass::arch::OpMultiplyAdd,
        EVTEpilogueStages
    >::GemmKernel;
```
error message:
>cutlass/include/cutlass/gemm/warp/mma_tensor_op_policy.h(58): error: incomplete type is not allowed
          detected during:
            instantiation of class "cutlass::gemm::warp::MmaTensorOpPolicy<Operator_, OpDelta_> [with Operator_=cutlass::arch::Mma<cutlass::gemm::GemmShape<16, 8, 32>, 32, int8_t, cutlass::layout::RowMajor, cutlass::int4b_t, cutlass::layout::ColumnMajor, int32_t, cutlass::layout::RowMajor, cutlass::arch::OpMultiplyAdd>, OpDelta_=cutlass::MatrixShape<1, 1>]" 

The complete log is [here](https://pastebin.com/ZV3YzM6K). 
When I set `ElementB = int8_t `, It's OK.  Is it because `DefaultGemmWithVisitor` doesn't support `s8/s4`?
If it's possible pointers in the correct direction would be greatly appreciated, thanks!

alexsamardzic:
> I got errors with `ElementB = cutlass::int4b_t` when I tried to follow this [example](https://github.com/NVIDIA/cutlass/blob/main/examples/47_ampere_gemm_universal_streamk/ampere_gemm_universal_streamk_broadcast.cu#L241):
> 

Are you using CUTLASS main, or the branch from this PR?

Hongbosherlock:
> > I got errors with `ElementB = cutlass::int4b_t` when I tried to follow this [example](https://github.com/NVIDIA/cutlass/blob/main/examples/47_ampere_gemm_universal_streamk/ampere_gemm_universal_streamk_broadcast.cu#L241):
> 
> Are you using CUTLASS main, or the branch from this PR?

I'm using this PR branch: `alexsamardzic:add-mixed-4bit-8bit-gemm`, I can get the right s4/s8 GEMM result  with `using Gemm = cutlass::gemm::device::GemmUniversal` as added in the `test/unit/gemm/device`.  
But when I try to use`GemmUniversalBase` or `GemmUniversalAdapter` which need to specify a `GemmKernel`, It can't work.

alexsamardzic:
> But when I try to use`GemmUniversalBase` or `GemmUniversalAdapter` which need to specify a `GemmKernel`, It couldn't work with `cutlass::int4b_t`, while `int8_t` could.
> 

Can you post your full code here?

Hongbosherlock:
> > But when I try to use`GemmUniversalBase` or `GemmUniversalAdapter` which need to specify a `GemmKernel`, It couldn't work with `cutlass::int4b_t`, while `int8_t` could.
> 
> Can you post your full code here?

```c++
#include <torch/extension.h>
#include <cuda.h>
#include <cuda_runtime.h>
#include "gemm_cuda.h"
#include <ATen/cuda/CUDAContext.h>
#include <cutlass/core_io.h>
#include <cutlass/cutlass.h>
#include <cutlass/half.h>
#include <cutlass/gemm/device/gemm.h>
#include <cutlass/numeric_types.h>
#include <cutlass/util/host_tensor.h>

#include <cutlass/gemm/device/gemm_universal.h>
#include <cutlass/util/reference/host/gemm.h>
#include <cutlass/util/reference/host/tensor_compare.h>
#include <cutlass/util/reference/host/tensor_copy.h>
#include <cutlass/util/reference/host/tensor_fill.h>
#include <cutlass/util/tensor_view_io.h>

#include <cutlass/gemm/device/gemm_universal_with_broadcast.h>
#include <cutlass/gemm/device/gemm_universal_streamk_with_broadcast.h>

#include <cutlass/util/reference/host/error_metrics.h>
#include <cutlass/util/reference/host/tensor_foreach.h>
#include <cutlass/epilogue/threadblock/fusion/visitors.hpp>
#include <cutlass/gemm/kernel/default_gemm_universal_with_visitor.h>
#include <cutlass/gemm/device/gemm_universal_adapter.h>


torch::Tensor matmul_w4a8(const torch::Tensor &A, const torch::Tensor &B, const torch::Tensor &alphaCol, const torch::Tensor &alphaRow) {
    torch::checkAllSameGPU("W4A8Matmul", {{A, "A", 0}, {B, "B", 1}});
    auto M = A.size(0);
    auto N = B.size(0);
    auto K = A.size(1);  // 4bit packing is on the columns
    auto D = torch::empty({M, N}, torch::dtype(torch::kFloat32).device(A.device())); 

    // A matrix configuration
    using         ElementA         = int8_t;                                    // Element type for A matrix operand
    using         LayoutA          = cutlass::layout::RowMajor;                        // Layout type for A matrix operand
    constexpr int AlignmentA       = 128 / cutlass::sizeof_bits<ElementA>::value;      // Memory access granularity/alignment of A matrix in units of elements (up to 16 bytes)

    // B matrix configuration
    using         ElementB         = cutlass::int4b_t;                                  // Element type for B matrix operand
    using         LayoutB          = cutlass::layout::ColumnMajor;                        // Layout type for B matrix operand
    constexpr int AlignmentB       = 128 / cutlass::sizeof_bits<ElementB>::value;      // Memory access granularity/alignment of B matrix in units of elements (up to 16 bytes)

    // C1/C2/D matrix configuration
    using         ElementC         = float;        //cutlass::half_t;                   // Element type for C matrix operands
    using         LayoutC          = cutlass::layout::RowMajor;                        // Layout type for C matrix operands
    constexpr int AlignmentC       = 128 / cutlass::sizeof_bits<ElementC>::value;      // Memory access granularity/alignment of C matrices in units of elements (up to 16 bytes)

    // Output matrix configuration
    using         ElementOutput    = float;                                          // Element type for output matrix operands
    using         LayoutOutput     = cutlass::layout::RowMajor;                        // Layout type for output matrix operands
    // constexpr int AlignmentOutput  = 128 / cutlass::sizeof_bits<ElementOutput>::value; // Memory access granularity/alignment of output matrices in units of elements (up to 16 bytes)

    // Multiply-accumulate blocking/pipelining details
    using ElementAccumulator  = int32_t;                                 // Element type for internal accumulation
    using ElementCompute      = float;  //cutlass::half_t;                          // Element type for compute
    using ArchTag             = cutlass::arch::Sm80;                      // Tag indicating the minimum SM that supports the intended feature
    using OperatorClass       = cutlass::arch::OpClassTensorOp;           // Operator class tag
    using ThreadblockShape    = cutlass::gemm::GemmShape<128, 128, 64>;   // Threadblock-level tile size (concept: GemmShape)
    using WarpShape           = cutlass::gemm::GemmShape<64, 64, 64>;     // Warp-level tile size (concept: GemmShape)
    using InstructionShape    = cutlass::gemm::GemmShape<16, 8, 32>;      // Instruction-level tile size (concept: GemmShape)
    constexpr int NumStages   = 4;                                        // Number of global->shared pipeline stages used in the GEMM mainloop
    constexpr int EVTEpilogueStages = 1;   

    // StreamK device GEMM implementation type with EVT
    using namespace cute;

    using OutputTileThreadMap = cutlass::epilogue::threadblock::OutputTileThreadLayout<
    ThreadblockShape, 
    WarpShape, 
    ElementC,  
    AlignmentC,  //4
    EVTEpilogueStages
    >;

    using Accum = cutlass::epilogue::threadblock::VisitorAccFetch;

    // alphaCol    [M, 1]    fp32
    using V1Broadcast = cutlass::epilogue::threadblock::VisitorColBroadcast<
        OutputTileThreadMap, ElementC,
        cute::Stride<int32_t, _1, _0>  // StrideMNL
    >;

    // alphaRow    [1, N]    fp32
    using V2Broadcast = cutlass::epilogue::threadblock::VisitorRowBroadcast<
        OutputTileThreadMap, ElementC,
        cute::Stride<_0, _1, int32_t>  // StrideMNL
    >;

    // mul
    using Compute0 = cutlass::epilogue::threadblock::VisitorCompute<
        cutlass::multiplies, ElementCompute, ElementCompute,
        cutlass::FloatRoundStyle::round_to_nearest
    >;

    // alphaCol * accumulator
    using EVTCompute0 = cutlass::epilogue::threadblock::Sm80EVT<
        Compute0,
        V1Broadcast,
        Accum>;

    // mul
    using Compute1 = cutlass::epilogue::threadblock::VisitorCompute<
        cutlass::multiplies, ElementOutput, ElementCompute,
        cutlass::FloatRoundStyle::round_to_nearest
    >;

    // alphaRow * alphaCol * accumulator
    using EVTCompute1 = cutlass::epilogue::threadblock::Sm80EVT<
        Compute1,
        V2Broadcast,
        EVTCompute0>;

    using StoreD = cutlass::epilogue::threadblock::VisitorAuxStore<
        OutputTileThreadMap, ElementOutput, cutlass::FloatRoundStyle::round_to_nearest,
        cute::Stride<int64_t, _1, int64_t> // StrideMNL
    >;

    using EVTD = cutlass::epilogue::threadblock::Sm80EVT<
        StoreD,
        EVTCompute1>;

    using EVTKernelStreamK =
        typename cutlass::gemm::kernel::DefaultGemmWithVisitor<
        ElementA, LayoutA, cutlass::ComplexTransform::kNone, AlignmentA,
        ElementB, LayoutB, cutlass::ComplexTransform::kNone, AlignmentB,
        ElementC, LayoutC, AlignmentC,
        ElementAccumulator,
        ElementCompute,
        cutlass::arch::OpClassTensorOp,
        cutlass::arch::Sm80,
        ThreadblockShape,
        WarpShape,
        InstructionShape,
        EVTD,
        cutlass::gemm::threadblock::ThreadblockSwizzleStreamK,
        NumStages,
        cutlass::arch::OpMultiplyAddSaturate,
        EVTEpilogueStages
    >::GemmKernel;

    using DeviceGemmStreamK = cutlass::gemm::device::GemmUniversalAdapter<EVTKernelStreamK>;

    // Populates a DeviceGemmStreamK::Arguments structure from the given commandline options

    // Ensure the input tensors are in the correct device and layout
    auto tensor_a = A.contiguous();
    auto tensor_b = B.contiguous();
    auto tensor_v1 = alphaCol.contiguous();
    auto tensor_v2 = alphaRow.contiguous();
    auto tensor_d = D.contiguous();                                                                                              // EVTD

    typename EVTD::Arguments callback_args{
        {
            {
                {
                    {},                                                                                                          // Accum
                    {tensor_v1.data_ptr<ElementC>(), ElementC(0), {int32_t(M), _1{}, _0{}}},                                    // V1 Broadcast
                    {}                                                                                                           // Compute0
                },                                                                                                             // EVTCompute0
                {tensor_v2.data_ptr<ElementC>(), ElementC(0), {_0{}, _1{}, int32_t(N)}},                                      // V2 Broadcast
                {}                                                                                                             // Compute1
            },                                                                                                               // EVTCompute1
            {}                                                                                                               // Compute2
        },                                                                                                                // EVTCompute2
        {tensor_d.data_ptr<ElementC>(), {int32_t{N}, _1{}, int32_t{M*N}}}                                                                  // D
    };                                                                                                                   // EVTD
    
                  
    using GemmCoord = cutlass::gemm::GemmCoord;
    cutlass::gemm::GemmUniversalMode mode = cutlass::gemm::GemmUniversalMode::kGemm;
    int batch_count = 1;
    // Construct Gemm ProblemSize with user defined output size
    // cutlass::gemm::GemmCoord problem_size = {M, N, K};
    cutlass::gemm::GemmCoord problem_size(M, N, K);

    int64_t stride_A = M * K;
    int64_t stride_B = N * K;
    // int64_t stride_C = M * N;
    // int64_t stride_D = M * N;                                                                                               // EVTD
    int     avail_sms = -1;
    typename DeviceGemmStreamK::Arguments arguments(
        mode,                                     // universal mode
        problem_size,                             // problem_size
        batch_count,                              // batch count / splitk slices
        callback_args,                            // argument of EVT callbacks
        tensor_a.data_ptr<ElementA>(),            // ptr_A
        (cutlass::int4b_t *)tensor_b.data_ptr<uint8_t>(),            // ptr_B
        nullptr,                                  // ptr_C (unused)
        nullptr,                                  // ptr_D (unused)
        stride_A,                                 // batch_stride_A
        stride_B,                                 // batch_stride_B
        0,                                        // batch_stride_C (unused)
        0,                                        // batch_stride_D (unused)
        tensor_a.stride(0),                       // stride_a
        tensor_b.stride(0),                       // stride_b
        0,                                        // stride_c (unused)
        0);                                      // stride_d (unused)
                            

    DeviceGemmStreamK gemm_op;
    
    auto stream = at::cuda::getCurrentCUDAStream(A.get_device());

    // Using the arguments, query for extra workspace required for matrix
    // multiplication computation
    size_t workspace_size = DeviceGemmStreamK::get_workspace_size(arguments);

    // Allocate workspace memory
    cutlass::device_memory::allocation<uint8_t> workspace(workspace_size);

    // Check the problem size is supported or not
    cutlass::Status status = gemm_op.can_implement(arguments);
    if (status != cutlass::Status::kSuccess) {
        throw std::runtime_error("cutlass cannot implement");
    }

    // Initialize CUTLASS kernel with arguments and workspace pointer
    status = gemm_op.initialize(arguments, workspace.get(), stream);
    if (status != cutlass::Status::kSuccess) {
        throw std::runtime_error("cutlass cannot initialize");
    }

    status = gemm_op(stream);
    if (status != cutlass::Status::kSuccess) {
        throw std::runtime_error("cutlass cannot run");
    }

    return tensor_d;
}
```
I got errors about ` incomplete type is not allowed`.

alexsamardzic:
This code uses PyTorch, can you post a reproducible example that uses CUTLASS only?

Hongbosherlock:
> This code uses PyTorch, can you post a reproducible example that uses CUTLASS only?

Hi @alexsamardzic , I have pushed my code here: https://github.com/Hongbosherlock/cutlass/blob/add-mixed-4bit-8bit-gemm/examples/61_s4s8_gemm/s4s8_gemm.cu#L104
>when `ElementB` is `int8_t`, everything is ok

you can add this example , then complie and run it:
```
# cutlass/build$ cmake .. -DCUTLASS_NVCC_ARCHS=80

# cutlass/build$ make 61_s4s8_gemm 

#  cutlass/build$ ./examples/61_s4s8_gemm/61_s4s8_gemm
```
you can get the result:
image

But when `ElementB = cutlass::int4b_t`,  lots of compilation errors occur.
I am really at a loss and would greatly appreciate any guidance or help you can provide. Thank you very much in advance for your time and assistance!

Hongbosherlock:
> > I have pushed my code here: https://github.com/Hongbosherlock/cutlass/blob/add-mixed-4bit-8bit-gemm/examples/61_s4s8_gemm/s4s8_gemm.cu
> 
> Replace `cutlass::arch::OpMultiplyAddSaturate` with `cutlass::arch::OpMultiplyAddMixedInputUpcast`.

Works for me. Thanks!

alexsamardzic:
> Works for me. Thanks!

Good.  Remember that CUTLASS is a heavily templated library, but actually small number of all the possible template argument combination work together - so one cannot just paste pieces of code from different sources, and expect it to work.

Hongbosherlock:
> > Works for me. Thanks!
> 
> Good. Remember that CUTLASS is a heavily templated library, but actually small number of all the possible template argument combination work together - so one cannot just paste pieces of code from different sources, and expect it to work.

Yea, that was a foolish mistake. `OpMultiplyAddMixedInputUpcast` did appear in the `test` folder. I think I was a bit disoriented. There are too many arguments.
I think cutlass is somewhat challenging for beginners. Do you have any recommended learning paths?

alexsamardzic:
> I think CUTLASS is somewhat challenging for beginners. Do you have any recommended learning paths?

Going through relevant examples, as well as unit tests, in the CUTLASS source tree is probably still the best way to start.

Hongbosherlock:
Hi @alexsamardzic , in fact, I am working in the GEMM+dequantization fusion kernel for W4A8 based on your PR, similar to the W8A8 [here](https://github.com/vllm-project/vllm/commit/2060e93659f1f63a3d2a76aee61559ccb1fe732e#diff-d40ac83826ce61086d3e34733c0d805eefa01c96e635a08f1e07b7b6ae93e5a6R120) (GEMM+DQ), which also used EVT.

I have completed most of the work and used EVT to finish the DQ. 
For simplicity, I started with W8A8 to verify that my EVT is correct, and then I will modify it for W4A8.
[Here](https://github.com/Hongbosherlock/cutlass/blob/add-mixed-4bit-8bit-gemm/examples/61_s4s8_gemm/s4s8_gemm.cu) is the W8A8 GEMM+DQ code. The code can be compiled, but the results are incorrect compared to the Python simulation results. 
I tried printing some elements, and they are very large values like `18766625`, which might indicate an address access error.

Can you please have a look what the possible issues might be? 
Thanks!

Hongbosherlock:
Hi @alexsamardzic ,thanks for your help, I have make it done.

When profiling a single GEMM, do you think the performance of `s4/s8` will be better than that of of `s8/s8`?
In my test s8/s8(int8 GEMM) is faster.

alexsamardzic:
> When profiling a single GEMM, do you think the performance of `s4/s8` will be better than that of of `s8/s8`? In my test s8/s8(int8 GEMM) is faster.

The GEMM actually performed is the same: `S8`/`S8` in both cases.  The bandwidth used in transfers between global and shared memory, and then between shared memory and registers, is smaller in `S4`/`S8` case, but on the other side there are additional calculations for `S4`->`S8` conversion in this case.  Thus, very broadly: `S4`/`S8` and `S8`/`S8` should be in the same ballpark regarding performance, but it is not unusual if for specific input shapes, and tile sizes selected, one is faster than the other one.

include/cutlass/gemm/warp/mma_mixed_input_tensor_op.h
```diff
@@ -122,8 +123,10 @@ struct FragmentShuffler <ElementMma_, ElementLoad_,
                          NumElementsInWarpFragment, 
                          NumElementsInMmaFragment,
                          Operand::kA,
-                         typename platform::enable_if<(sizeof_bits<ElementMma_>::value == 16) &&
-                                                 (sizeof_bits<ElementLoad_>::value == 8)>::type> {
+                         typename platform::enable_if<((sizeof_bits<ElementMma_>::value == 16) &&
+                                                       (sizeof_bits<ElementLoad_>::value == 8)) ||
+                                                      ((sizeof_bits<ElementMma_>::value == 8) &&
+                                                       (sizeof_bits<ElementLoad_>::value == 4))>::type> {
```
manishucsd, line 129: It is pretty pleasing to see that the same `FragmentShuffler` that worked for `F16`x`S8` is working for `S8`x`S4`. I have been trying to visualize it in my head. The invariant here is the ratio `bits(ElementMma_) : bits(ElementLoad_) = 2`. Additionally, `16-by-8-by-128b` Tensor Core operation stays fixed. 

For instance, `F16` x `S4` would need a different FragmentShuffler implementation which has different ratio `16-by-8-by-128b`. correct? 

Thinking out loud, This ratio maybe useful to create a single `FragmentShuffler` implementation that will work for `F16` x `S4` as well. 

cc: @hwu36 , @IonThruster

python/cutlass_library/generator.py
```diff
@@ -2836,6 +2836,151 @@ def GenerateSM80_TensorOp_16832_TN(manifest, cuda_version):
...
+      TileDescription([ 64, 128,  64],  6, [2, 2, 1], math_inst, min_cc, max_cc),
+      TileDescription([256, 128, 128],  3, [4, 2, 1], math_inst, min_cc, max_cc),
+      TileDescription([128, 256, 128],  3, [2, 4, 1], math_inst, min_cc, max_cc),
+      TileDescription([256,  64, 128],  4, [4, 1, 1], math_inst, min_cc, max_cc),
+      TileDescription([ 64, 256, 128],  4, [1, 4, 1], math_inst, min_cc, max_cc),
+      TileDescription([256,  32, 128],  4, [4, 1, 1], math_inst, min_cc, max_cc),
+      TileDescription([ 32, 256, 128],  4, [1, 4, 1], math_inst, min_cc, max_cc),
+      TileDescription([128, 128, 128],  4, [2, 2, 1], math_inst, min_cc, max_cc),
+      TileDescription([ 64, 128, 128],  3, [2, 2, 1], math_inst, min_cc, max_cc),
+      TileDescription([128,  32, 128],  4, [4, 1, 1], math_inst, min_cc, max_cc),
```
manishucsd, line 2883: How did you come up with this `TileDescription` list for `S8` x `S4`? I guess you carried these from `S8` x `S8`. Please make sure all of these pass verification. You can follow steps similar to [here](https://github.com/NVIDIA/cutlass/pull/1350#issuecomment-2223601983) to instantiate all the tile shapes listed here by using `-DCUTLASS_LIBRARY_KERNELS="s8_s4,s4_s8"`. By default the build process only instantiate 128x128 tile shape.

manishucsd (approved):
Thank you for working on this. Apologies for a delayed review. LGTM. 
Over to NVIDIA/CUTLASS (cc: @hwu36 ) for merging this.

python/cutlass_library/generator.py
manishucsd: Can you please run the verification and profiling on `--m=3456 --n=4096 --k=2048` on an A100? 

Please compile using `-DCUTLASS_LIBRARY_KERNELS="s8_s4,s4_s8,s4,s8"` to also have s4 x s4 and s8 x s8 kernels in the runs.

include/cutlass/gemm/warp/mma_mixed_input_tensor_op.h
alexsamardzic: Yes, the shuffler is reused from `F16xS8`, namely in both cases, for each 32 bits loaded into a thread registers through `ldmatrix`, two groups of 16 bits are shuffled between threads.  Alike for `F16xS4`, the shuffler that is added into #1190 could be reused for other combinations of operands where the bitwidth of wider operand is 4x the bitwidth of the other one.

(While we're at it: it would be great if #1190 could be revisited between you, @hwu36 and myself - the changes for the case when one operand is 4x wider than the other one are rather invasive, but these combinations of operands are still deemed quite important for PyTorch, and I'm willing to work to find an approach that would be acceptable for merging into CUTLASS.)

python/cutlass_library/generator.py
alexsamardzic: The tiles selection is desribed in a [comment](https://github.com/NVIDIA/cutlass/pull/1413#issuecomment-2015366925) above; also, as mentioned in this comment, I did the verification.  I will repeat the verification procedure, together with profiling, and report the outcome here.

include/cutlass/gemm/warp/mma_mixed_input_tensor_op.h
manishucsd: Yeah, the last time went in looking deep into #1190 it had mainloop changes. Those changes will affect almost all the Ampere kernels. While this is fresh, for me, let me leave a comment on #1190.

python/cutlass_library/generator.py
alexsamardzic: There was a build issue after rebasing on the latest main: basically, `OpMultiplyAddSaturate` for `MmaTensorOpPolicy` in the specialization of `struct DefaultMmaTensorOp` (in `include/cutlass/gemm/warp/default_mma_tensor_op_sm80.h`) seem to be obligatory now, as the build fails if `OpMultiplyAdd` used.  The branch is updated accordingly.

I've configured the build using `-DCUTLASS_NVCC_ARCHS=80 -DCUTLASS_LIBRARY_KERNELS="s8_s4,s4_s8,s4,s8"` CMake options, and then verified that `cutlass_test_unit_gemm_device_mixed_input_tensorop_sm80` unit test passes.  Then, I did profiler runs as follows:
```
./build/tools/profiler/cutlass_profiler --operation=gemm --m=3456 --n=4096 --k=2048 --A=s8:row --B=s8:column >& s8_s8.txt
./build/tools/profiler/cutlass_profiler --operation=gemm --m=3456 --n=4096 --k=2048 --A=s4:row --B=s8:column >& s4_s8.txt
./build/tools/profiler/cutlass_profiler --operation=gemm --m=3456 --n=4096 --k=2048 --A=s8:row --B=s4:column >& s8_s4.txt
```

The corresponding profiler outputs are here: 
[s8_s8.txt](https://github.com/user-attachments/files/16630281/s8_s8.txt)
[s4_s8.txt](https://github.com/user-attachments/files/16630289/s4_s8.txt)
[s8_s4.txt](https://github.com/user-attachments/files/16630290/s8_s4.txt)

The disposition values for mixed data types cases with `s8` accumulator are still `incorrect`.  Also, the timings are somewhat slower than for corresponding `s8xs8` cases (with the same configurations: tile sizes etc.).

manishucsd: Thank you for running and sharing these results. 

Accumulator is for all of these runs should be `S32` as shown at the bottom of the output in csv format with `accum` type = `S32`. The `Incorrect` disposition with mixed-input is happening for only `S8` output, i.e., when the accumulators are `S32` but the output is downcast-ed to `S8`. 

We do *not* see incorrect results for `S8`x`S8` with `S32` accumulators and `S8` output, can you pick one row of incorrect run from 

```
(elementD/elementC type) <= (elementA type) x(elementB type) + (accum type)
S8  <= S8  x S4  + S32 
```

and compare the same kernel configuration against 

```
S8 <= S8 x S8 + S32
```
to find where is the difference?


I believe it is to do with initialization of the operands during profiling or inside the kernel epilogue S32-to-S8. quantization. 

Also, you can just upload the csv that can be produced by adding `--output=filename.csv` to the profiler runs

alexsamardzic: The CSV files are attached below, I'm examining `incorrect` cases.

[s8_s8.gemm.csv](https://github.com/user-attachments/files/16635796/s8_s8.gemm.csv)
[s4_s8.gemm.csv](https://github.com/user-attachments/files/16635801/s4_s8.gemm.csv)
[s8_s4.gemm.csv](https://github.com/user-attachments/files/16635803/s8_s4.gemm.csv)

alexsamardzic: Here is what I found so far regarding `incorrect` cases:

First, I made following change in the code generating inputs, in order to generate the same inputs for profiler for `S4xS8` and `S8xS8` cases:
```
diff --git a/tools/profiler/src/device_context.cu b/tools/profiler/src/device_context.cu
index 2cbfa5d2..7b488fe8 100644
--- a/tools/profiler/src/device_context.cu
+++ b/tools/profiler/src/device_context.cu
@@ -105,7 +105,7 @@ DeviceAllocation *DeviceContext::allocate_tensor(
           data_distribution.set_uniform(-1, 1, 0);
           break;
         case library::NumericTypeID::kS4:
-          data_distribution.set_uniform(-2, 2, 0);
+          data_distribution.set_uniform(-3, 3, 0);
           break;
         case library::NumericTypeID::kU2:
           data_distribution.set_uniform(0, 2, 0);
```

I used following profiler runs to make comparision between `S4xS8` and `S8xS8` cases (BTW, I found that smaller input shapes selection that would still allow for reproducing the problem would be `--m=32 --n=64 --k=512`):
```
cutlass_profiler --operation=gemm --gemm_kind=universal --m=3456 --n=4096 --k=2048 --A=s8:row --B=s8:column --C=s8:column --D=s8:column --accum=s32 --cta_m=256 --cta_n=128 --cta_k=64 --stages=3 --save-workspace=always
cutlass_profiler --operation=gemm --gemm_kind=universal --m=3456 --n=4096 --k=2048 --A=s4:row --B=s8:column --C=s8:column --D=s8:column --accum=s32 --cta_m=256 --cta_n=128 --cta_k=64 --stages=3 --save-workspace=always
```

By comparing saved `.mat` files, I verified that input matrices `A`, `B` and `C` are the same, but also that output matrices `D` are the same.  What differs are actually `Reference` matrices, which means that reference results calculated for `S4xS8` case are wrong.  If I understood it correctly, cuBLAS is used for reference calculations, so I'll check what's going on there...

manishucsd: I am not sure if cuBLAS is called for this for reference check. The output should show which references are called. You can use `--verification-providers=cublas,host,device` to run them all. Is there a host reference for this you must check in here [/tools/library/src/reference](/tools/library/src/reference ) ?

alexsamardzic: Indeed - `device` provider is actually used for reference check here.  I posted an update with the fix in reference calculations, so for most of cases with `S8` output, `cutlass_profiler` reports success now.  However, there are still couple of cases where `incorrect` is reported, I'm looking into this...

alexsamardzic: Pushed another update - the problem with remaining `incorrect` cases was that I haven't copied `C` operand alignment update from `S8xS8` case, in the generator code.  Everything is reported as `passed` now by profiler, the output files are attached below.  I believe this one should be ready for merging now.

[s8_s8.gemm.csv](https://github.com/user-attachments/files/16681357/s8_s8.gemm.csv)
[s4_s8.gemm.csv](https://github.com/user-attachments/files/16681359/s4_s8.gemm.csv)
[s8_s4.gemm.csv](https://github.com/user-attachments/files/16681361/s8_s4.gemm.csv)

manishucsd: @alexsamardzic thank you for digging it through. LGTM!

@hwu36 , @thakkarV , @IonThruster , can you please help it merge it?

alexsamardzic (commented):
Thanks!

hwu36 (approved):

hwu36:
while we are at this, i think we can improve the int4->int8 upcasting.  now we use 11 instructions to upcast 8 elements.  quite a lot. we used a look-up-table method to do int->fp8 upcasting (https://github.com/NVIDIA/cutlass/blob/main/include/cutlass/numeric_conversion.h#L2983-L3027), I think we maybe able to use the same here.

@alexsamardzic , do you want to give it a try?  i am setting up now so it won't take me months to merge your code.

cc @rhenry-nv

alexsamardzic:
> @alexsamardzic , do you want to give it a try? i am setting up now so it won't take me months to merge your code.

Sure.  Below is a patch to implement the look-up table method for int4->int8 (pretty much the same as existing int4->fp8 code), and also the profiler outputs for original and patched version.  It seems that the look-up table method is slower.

I ran the profiler in both cases as follows:
```
cutlass_profiler --operation=Gemm --m=1024 --n=1024 --k=1024 --A=s4:row --B=s8:column
```
and here are mentioned files:
[patch.txt](https://github.com/user-attachments/files/16819978/patch.txt)
[original.csv](https://github.com/user-attachments/files/16819981/original.csv)
[patched.csv](https://github.com/user-attachments/files/16819983/patched.csv)

I was the least happy about the conversion code in this PR, but this is the best I was able to come up with...

PR #1416, opened 2024-03-20
group gemm set stride L = cute::Int<0>

Xseventh:
For group_gemm, specifying L is unnecessary; it can be set to Int<0> for a smaller stride size.
the data layout between cutlass2.x group_gemm lda and cutlass 3.x group_gemm stride can be identical, allowing users to utilize the same arguments when upgrading to the cutlass3.x API.
I have tested this with 57_hopper_grouped_gemm.


thakkarV:
@ANIKET-SHIVAM pretty valid change and usecase imo

hwu36 (approved):

PR #1428, opened 2024-03-26
Allow setting a custom TmaDescriptor for TMAStore.

ipiszy:
ATT, this is to support variable sequence length in the destination tensor.


ipiszy:
cc @thakkarV @hwu36 @ANIKET-SHIVAM

thakkarV:
Yep, a dedicated epilogue for pointer array and grouped Gemm is coming before we tag 3.5

hwu36:
@ipiszy , do you still need it?  it regressed the other kernels.  if you still need it, we have to try a different approach.

PR #1454, opened 2024-04-03
Restore "make_ordered_layout" method removed in 3.5

jayhshah:
From CUTLASS 3.4.1 to 3.5, one of the `make_ordered_layout` methods in `cute/layout.hpp` was removed, which breaks compatibility with the colfax FA-2 implementation. This PR restores that method.


jayhshah:
Sorry, I realized that the design decision must have been to migrate the functionality to `make_layout_like`. So this PR isn't needed.

ccecka:
Confirming that `make_layout_like(Layout)` should be used in place of `make_ordered_layout(Layout)`. Thanks.

PR #1465, opened 2024-04-09
Clarify the meaning of "M-major" in CuTe.

jlebar:


jlebar:
cc @thakkarV

ccecka:
Consider an `MxN` matrix named `A`. We agree that:
- The CuTe Layout `(M,N):(N,1)` corresponds to "row-major".
- The Cute Layout `(M,N):(1,M)` corresponds to "column-major".

> Outside of CuTe, we'd say that a matrix is "row-major" if the stride in the row dimension is `num_cols` and "row-minor" if the stride in the row dimension is 1.

First, row-major certainly does not require "the stride ... is `num_cols`". That is far too strict.

Second, "row dimension" is unused elsewhere and semantically troublesome -- does it mean the mode that I index into to get a row or the mode that I index into to get an element within a row? This is the core of the confusion which I expand on below.

Third, "row-minor" appears nowhere in your linked article (nor in any quick google searching...). This concept of "major" and "minor" appear to be confused. From the article:

> The terms row-major and column-major stem from the terminology related to ordering objects. A general way to order objects with many attributes is to first group and order them by one attribute, and then, within each such group, group and order them by another attribute, etc. If more than one attribute participates in ordering, the first would be called major and the last minor. If two attributes participate in ordering, it is sufficient to name only the major attribute.

which tells me that "major" means "ordered first".

Fourth, row-major is defined in your article as:

> In row-major order, the consecutive elements of a row reside next to each other...

which is expanded on in the next section of the article:

> In the case of arrays, **the attributes are the indices along each dimension**. For matrices in mathematical notation, the first index indicates the row, and the second indicates the column, e.g., given a matrix `A`, the entry `a_12` is in its first row and second column.

Note the highlighted text. While "the first index indicates the row, and the second indicates the column", it is also appropriate to say that "the first index indicates the position within a column, and the second indicates a position within a row". The latter semantically generalizes to tensors much more easily (where indexing into one mode doesn't necessarily give you a row or column as a result).

So, to me, row-major and N-major are very consistent: "As I increase the index within a row (index within the N-mode) the elements reside next to each other."

These confusions highlight why we like to avoid the use of the words "row" and "column" at all. In matrices, those words are duals -- `M` is the number of rows AND the size of a column while `m` is the row index AND the index within a column -- but in tensors these words don't make sense and we prefer to label the modes semantically instead.

jlebar:
Thank you for your reply, Cris.  In my defense, I put this together only after talking to Vijay -- sorry for wasting your time.

> So, to me, row-major and N-major are very consistent: "As I increase the index within a row (index within the N-mode) the elements reside next to each other."

I think this is the point of confusion.  If I have a matrix with R rows and C columns, I had thought that the R mode was the index of the row, i.e. the variable we'd call little-r.  Apparently it's the index *within* the row, i.e. the index of the column?

Does CuTe define "mode" somewhere?

ccecka:
It's an interesting discussion, no defense needed.

On your point, thinking of a `RxC` matrix as having `R` rows and `C` columns is common. Now let's generalize to tensors and think about an `MxNxL` object. Do we want to describe this tensor as having `M` `NL`-matrices and `N` `ML`-matrices and `L` `MN`-matrices... don't think so. I find that similar to saying `R` rows [`C`-modes] and `C` columns [`R`-modes]. That demonstrates that we're describing every mode by the space that is orthogonal to it... which is obviously unnatural after two dimensions.

A mode `M` is anything that can be indexed into with a coordinate `0 <= m < size(M)`.
A tensor is a tensor product of modes with its associated tensor product of coordinates.

thakkarV:
@jlebar My apologies for the confusion. I have not used the terms "row-major" or "col-major" in many years, and in my defense I did mention that "describing every mode by the space that is orthogonal to it" does not scale beyond 2 modes or multi modes but tripped myself up on the same thing. I think I will continue to stray away from the row-col major terminology even more than before :p

jlebar:
Thanks, Chris and Vijay.  I think this answers my questions and fixes my confusion.

PR #1469, opened 2024-04-10
  Support use of external/system GTest installation

iskunk:
(Re-push of #1384)

I am building the project in a restricted environment that does not allow network access (so FetchContent is not usable), and already have a pre-compiled build of GTest ready to go. With this change, I can pass `-DCUTLASS_USE_SYSTEM_GOOGLETEST=ON -DGTest_ROOT=/path/to/gtest-install` to CMake, and the tests will build using the specified install. A system install of GTest can also be picked up.

The second commit fixes a related issue: the test suite breaks when the `bin/` subdirectory is missing from the build tree, but it's the bundled-googletest logic that creates it. So we create that dir explicitly/unconditionally.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -133,6 +133,7 @@ set(CUTLASS_ENABLE_PERFORMANCE ${CUTLASS_ENABLE_PROFILER} CACHE BOOL "Enable CUT
 
 set(CUTLASS_ENABLE_TESTS ${CUTLASS_ENABLE_TESTS_INIT} CACHE BOOL "Enable CUTLASS Tests")
 set(CUTLASS_ENABLE_GTEST_UNIT_TESTS ${CUTLASS_ENABLE_TESTS} CACHE BOOL "Enable CUTLASS GTest-based Unit Tests")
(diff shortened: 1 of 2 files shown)


d-k-b (approved):

iskunk:
Is there an additional step needed for merging? Anything I need to do?

iskunk:
What is the status of development on this project? It does not appear to have received a commit, from anyone, for nearly a month.

thakkarV:
@d-k-b can you please shepherd this MR? I see that you have approved it.  CC: @hwu36 

@iskunk we are about to release 3.5.1 and due to logistics of our release process, we will likely only merge this after 3.5.1 is pushed out next week. Is that ok? Sorry for the delay.

d-k-b:
@hwu36 to confirm, but I believe this fix will come with 3.5.1. If not, we can certainly merge it shortly after.

iskunk:
@thakkarV: After 3.5.1 is fine; just want to make sure this gets in.

hwu36 (approved):

PR #1470, opened 2024-04-10
Add missing #include <memory> for definition of std::addressof.

Gregory-Meyer:
`detail::checked_addressof` uses `std::addressof`; failing to `#include` this header results in a compiler error if it didn't happen to be `#include`d already.

### DIFF
--- a/include/cutlass/cluster_launch.hpp
+++ b/include/cutlass/cluster_launch.hpp
@@ -36,6 +36,9 @@
 #pragma once
 
 #include <cuda_runtime_api.h>
+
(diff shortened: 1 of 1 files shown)


include/cutlass/cluster_launch.hpp
```diff
@@ -36,6 +36,9 @@
 #pragma once
 
 #include <cuda_runtime_api.h>
+
```
hwu36, line 39: could we move it to line 47?  i am afraid nvrtc may not like this header file.

PR #1477, opened 2024-04-12
Fix SMEM index for C in CuTe examples

joerowell:
The static check against `CSmemLayout` in all of the CuTe sgemm examples checks against `size<0>` twice. 

This has only passed thus far because `size<0>(CSmemLayout{}) == size<1>(CSmemLayout{});`, but it isn't strictly correct (it would imply that `CSmemLayout` is an BLK_M x BLK_M matrix, rather than a BLK_M x BLK_N matrix).


thakkarV:
Good catch :)

ccecka (approved):

hwu36 (approved):

Issue #1474, opened 2024-04-12
[QST] PyTorch + CUTLASS Batched GEMM Kernel - Expression Must be Constant


nickjeliopoulos:
Good evening, all. I am attempting to compile a minimal [CUTLASS](https://github.com/NVIDIA/cutlass/releases/tag/v3.4.1) GEMM example in a PyTorch project. I want to write a simple CUTLASS kernel and be able to call it from PyTorch.

I have a minimal example that refuses to compile.
**setup.py**
```
from setuptools import setup
from torch.utils.cpp_extension import BuildExtension, CUDAExtension, CppExtension

setup(
    ### Minimal Example Args
    name="gorby",
    install_requires=["torch >= 2.1", "pybind11"],
    ### PyTorch C++/CUDA Examples
    ext_modules=[
        #CUDAExtension(
        #    name="gorby_vector_add", sources=["gorby_vector_add.cu"]
        #),
        #CUDAExtension(
        #    name="gorby_nvtx", sources=["gorby_nvtx.cpp"]
        #),
        CUDAExtension(
            name="gorby_sdpa", sources=["gorby_sdpa.cu"],
            include_dirs=['C:\\Users\\neliopou\\Documents\\PhD\\Github\\cutlass\\include']
        ),
    ],
    cmdclass={"build_ext": BuildExtension},
)
```

**gorby_sdpa.cu** - taken from CUTLASS https://github.com/NVIDIA/cutlass/blob/main/examples/05_batched_gemm/batched_gemm.cu
```
// CUTLASS
#include "cutlass/cutlass.h"
#include "cutlass/layout/matrix.h"
#include "cutlass/gemm/device/gemm_array.h"
#include "cutlass/gemm/device/gemm_batched.h"

cudaError_t cutlass_array_sgemm(
  int m,
  int n,
  int k,
  float alpha,
  float const * const *A,
  int lda,
  float const * const *B,
  int ldb,
  float * const *C,
  int ldc,
  float beta,
  int batch_count) {

  using Gemm = cutlass::gemm::device::GemmArray<
    float, cutlass::layout::ColumnMajor,
    float, cutlass::layout::ColumnMajor,
    float, cutlass::layout::ColumnMajor
  >;

  Gemm gemm_op;

  cutlass::Status status = gemm_op({
    {m, n, k},
    A, lda,
    B, ldb,
    C, ldc,
    C, ldc,
    {alpha, beta},
    batch_count
  });

[...]

PR #1479, opened 2024-04-12
Fix C++17 version detection in helper_macros.hpp

nickjeliopoulos:
It seems that __cplusplus can be inconsistent with _MSVC_LANG when discerning C++17 version in some cases.
See https://github.com/NVIDIA/cutlass/issues/1474. 
Added switch to check _MSVC_LANG in addition to __cplusplus.

### DIFF
--- a/include/cutlass/detail/helper_macros.hpp
+++ b/include/cutlass/detail/helper_macros.hpp
@@ -146,7 +146,13 @@ namespace cutlass {
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////
 
-#if (201700L <= __cplusplus)
+#ifndef _MSVC_LANG  
+#define MSCV_LANG _MSVC_LANG
+#else
+#define MSCV_LANG 0
+#endif
+
+#if (201700L <= __cplusplus || 201700L <= MSVC_LANG)
 #define CUTLASS_CONSTEXPR_IF_CXX17 constexpr
 #define CUTLASS_CXX17_OR_LATER 1
 #else
--- a/include/cutlass/detail/helper_macros.hpp
+++ b/include/cutlass/detail/helper_macros.hpp
@@ -152,7 +152,7 @@ namespace cutlass {
 #define MSCV_LANG 0
 #endif
 
-#if (201700L <= __cplusplus || 201700L <= MSVC_LANG)
+#if (201700L <= __cplusplus || 201700L <= MSCV_LANG)
 #define CUTLASS_CONSTEXPR_IF_CXX17 constexpr
 #define CUTLASS_CXX17_OR_LATER 1
 #else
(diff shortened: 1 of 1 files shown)


jackkosaian:
@mhoemmen, can you take a look?

d-k-b:
Please set the flag `/Zc:__cplusplus` as shown in https://github.com/NVIDIA/cutlass/blob/7d49e6c7e2f8896c47f586706e67e1fb215529dc/CMakeLists.txt#L439-453 to get MSVC to set the compile options appropriately.

nickjeliopoulos:
I was able to include CUTLASS headers successfully with @d-k-b suggestion. I closed my issue #1474 related to this. 

Is there still some value to this PR for those who want to include CUTLASS headers-only?
Or is this better solved upstream (in something like PyTorch's `torch/utils/cpp_extensions.py`?

mhoemmen:
@jackkosaian Hi!  Thanks for the heads-up!  I can take a look.

include/cutlass/detail/helper_macros.hpp
```diff
@@ -146,7 +146,13 @@ namespace cutlass {
...
 ////////////////////////////////////////////////////////////////////////////////////////////////////
 
-#if (201700L <= __cplusplus)
+#ifdef _MSVC_LANG  
+#define MSVC_LANG _MSVC_LANG
+#else
+#define MSVC_LANG 0
+#endif
+
+#if (201700L <= __cplusplus || 201700L <= MSVC_LANG)
```
mhoemmen, line 155: 1. The C++ Standard reserves identifiers with `_` prefix for the implementation's use.
2. This change defines a macro with a short name and no library prefix (e.g.,` CUTLASS_`) for all downstream code.
3. It looks like this macro is only used on line 155.

As a result, should we consider not defining `MSVC_LANG`, and instead expressing the idea inline?

```suggestion
#if (defined(_MSVC_LANG) && (201700L <= MSVC_LANG)) || (201700L <= __cplusplus)
```

mhoemmen: Alternately, should we consider name-prefixing the macro and defining it as "the way to test the language version for any compiler"?

```suggestion
#if defined(_MSVC_LANG)
#  define CUTLASS_CPLUSPLUS _MSVC_LANG
#else
#  define CUTLASS_CPLUSPLUS __cplusplus
#endif

#if (201700L <= CUTLASS_CPLUSPLUS)
```

mhoemmen (changes requested):
Greetings and thanks for your contribution!

Would you consider the minor changes requested in this review?

mhoemmen:
@nickjeliopoulos Excellent, thanks so much!

mhoemmen (approved):

mhoemmen:
Let me check on the status of this with @hwu36 .

mhoemmen:
@hwu36 I think we can merge this.

PR #1534, opened 2024-05-11
Fix template parameter `IterationsUnroll` type from int to bool

peakcrosser7:


peakcrosser7:
Closed due to being stale.

PR #1539, opened 2024-05-16
Allow scalar broadcasting in VisitorRowBroadcast and VisitorColBroadcast

tlrmchlsmth:
This PR addresses an inconsistency between the VisitorRowBroadcast/VisitorColBroadcast epilogues and the SM90RowBroadcast/SM90ColBroadcast epilogues. 

The inconsistency is that the SM90 epilogues can handle either row/column broadcasting by passing in a nullptr for the first argument, and a float for the second, while the visitor epilogues cannot. This PR adds this functionality to the visitor epilogues.

I am using this for quantized GEMMs that can handle either per-token/per channel quantization or per-tensor quantization without compiling and distributing multiple kernels to handle all cases. 

For reference, I ran into this issue when developing https://github.com/vllm-project/vllm/pull/4749

### DIFF
--- a/include/cutlass/epilogue/threadblock/fusion/visitor_load.hpp
+++ b/include/cutlass/epilogue/threadblock/fusion/visitor_load.hpp
@@ -403,10 +403,29 @@ struct VisitorRowBroadcast {
       auto src_v = filter(tC_gRow);
       auto coord_v = filter(tC_cRow);
       auto dst_v = filter(tC_rRow);
-      CUTLASS_PRAGMA_UNROLL
-      for (int i = 0; i < size(src_v); ++i) {
-        bool guard = get<1>(coord_v(i)) < n;
-        cutlass::arch::global_load<VecType, sizeof(VecType)>(dst_v(i), (void const*)&src_v(i), guard);
+
+      if (params_ptr->ptr_row) {
+        // In this case we are loading from a row vector and broadcasting
+        CUTLASS_PRAGMA_UNROLL
+        for (int i = 0; i < size(src_v); ++i) {
+          bool guard = get<1>(coord_v(i)) < n;
+          cutlass::arch::global_load<VecType, sizeof(VecType)>(dst_v(i), (void const*)&src_v(i), guard);
+        }
+      } else {
+        // In this case we are loading from a scalar and broadcasting
+        VecType filled_vec;
+        CUTLASS_PRAGMA_UNROLL
+        for (int i = 0; i < VecLength; i++) {
+          reinterpret_cast<Element*>(&filled_vec)[i] = params_ptr->null_default;
+        }
+
+        CUTLASS_PRAGMA_UNROLL
+        for (int i = 0; i < size(src_v); ++i) {
+          if(get<1>(coord_v(i)) < n)
+          {
+            dst_v(i) = filled_vec;
+          }
+        }
       }
     }
 
@@ -524,12 +543,27 @@ struct VisitorColBroadcast {
     CUTLASS_DEVICE void
     begin_epilogue() {
       clear(tC_rCol);
+
       Tensor pred = make_tensor<bool>(shape(tC_gCol));
       CUTLASS_PRAGMA_UNROLL
       for (int i = 0; i < size(pred); ++i) {
         pred(i) = get<0>(tC_cCol(i)) < m;
       }
-      copy_if(pred, tC_gCol, tC_rCol);
+
+      if (params_ptr->ptr_col) {
+        // In this case we are loading from a column vector and broadcasting
+        copy_if(pred, tC_gCol, tC_rCol);
+      } else {
+        // In this case we are loading from a scalar and broadcasting
+        auto dst_v = filter(tC_rCol);
+
+        CUTLASS_PRAGMA_UNROLL
+        for (int i = 0; i < size(dst_v); ++i) {
+          if(pred(i)){
+             dst_v(i) = params_ptr->null_default;
+          }
+        }
+      }
     }
 
     template <class ElementAccumulator, int FragmentSize>


tlrmchlsmth:
Very happy to add unit tests and put in the work to get this PR into a landable state. But first hoping to get some high-level feedback on whether this is the right approach or a reasonable thing to do. Thanks!

tlrmchlsmth:
cc @mnicely

Hongbosherlock:
Hi @tlrmchlsmth  thanks for your contribution. I'm working on int8 GEMM with dequant fusion. 
Can the following code work with the original `VisitorRowBroadcast/VisitorColBroadcast` epilogues?
```c++
    // inputs
    //     A           [M, K]    int8
    //     B           [N, K]    int4
    //     alphaCol    [M, 1]    fp32
    //     alphaRow    [1, N]    fp32
    // outputs
    //     mat [M, N]            fp32

    // alphaCol    [M, 1]    fp32
    using V1Broadcast = cutlass::epilogue::threadblock::VisitorColBroadcast<
        OutputTileThreadMap, ElementC,
        cute::Stride<int32_t, _1, _0>  // StrideMNL
    >;

    // alphaRow    [1, N]    fp32
    using V2Broadcast = cutlass::epilogue::threadblock::VisitorRowBroadcast<
        OutputTileThreadMap, ElementC,
        cute::Stride<_0, _1, int32_t>  // StrideMNL
    >;
```

> The inconsistency is that the SM90 epilogues can handle either row/column broadcasting by passing in a nullptr for the first argument, and a float for the second, while the visitor epilogues cannot. This PR adds this functionality to the visitor epilogues.

I don’t quite understand this PR. Regarding this issue, could you please provide some  examples? In what situations won’t it work, and in what situations will it work based on this PR?

tlrmchlsmth:
@Hongbosherlock 
Compare the cutlass 2.0 epilogues in [include/cutlass/epilogue/threadblock/fusion/visitor_load.hpp](https://github.com/NVIDIA/cutlass/blob/main/include/cutlass/epilogue/threadblock/fusion/visitor_load.hpp) and 
the cutlass 3.0 epilogues in [include/cutlass/epilogue/fusion/sm90_visitor_load_tma_warpspecialized.hpp](https://github.com/NVIDIA/cutlass/blob/main/include/cutlass/epilogue/fusion/sm90_visitor_load_tma_warpspecialized.hpp)

In the second, the row and column broadcast epilogues (`Sm90RowBroadcas`t and `Sm90ColBroadcast`) have `null_default` arguments that are used to provide scalar broadcast functionality. In the first file, the similar row and column broadcast epilogues also have `null_default` arguments but they simply aren't used. 

I tried the approach you suggest for cutlass 2.0 but couldn't get it to compile. If you have a full working example, I'd like to see it :)

Anyway, the same approach won't work for cutlass 3.0, as you will fail this static assert `cute::is_same_v<StrideMNL, Stride<_1,_0, _0>>)`, and the problem this PR is addressing is the _inconsistency_ between these two very similar types.

thakkarV:
@hwu36 can we ask Zhaodong to merge this? I don't know his GitHub username

tlrmchlsmth:
JFYI I did end up going in a different direction with these epilogue changes. See https://github.com/vllm-project/vllm/pull/5137 -- I found that it was much nicer for a variety of reasons if both the scalar and the vector broadcast cases take a `float *` that points to device memory as an argument.

include/cutlass/epilogue/threadblock/fusion/visitor_load.hpp
```diff
@@ -403,10 +403,29 @@ struct VisitorRowBroadcast {
...
-        cutlass::arch::global_load<VecType, sizeof(VecType)>(dst_v(i), (void const*)&src_v(i), guard);
+
+      if (params_ptr->ptr_row) {
+        // In this case we are loading from a row vector and broadcasting
+        CUTLASS_PRAGMA_UNROLL
+        for (int i = 0; i < size(src_v); ++i) {
+          bool guard = get<1>(coord_v(i)) < n;
+          cutlass::arch::global_load<VecType, sizeof(VecType)>(dst_v(i), (void const*)&src_v(i), guard);
+        }
+      } else {
```
thakkarV, line 414: Nit: New line after branch close.

include/cutlass/epilogue/threadblock/fusion/visitor_load.hpp
```diff
@@ -403,10 +403,29 @@ struct VisitorRowBroadcast {
...
+        VecType filled_vec;
+        CUTLASS_PRAGMA_UNROLL
+        for (int i = 0; i < VecLength; i++) {
+          reinterpret_cast<Element*>(&filled_vec)[i] = params_ptr->null_default;
+        }
+
+        CUTLASS_PRAGMA_UNROLL
+        for (int i = 0; i < size(src_v); ++i) {
+          if(get<1>(coord_v(i)) < n)
+          {
```
thakkarV, line 425: Not: no new line before brace open

include/cutlass/epilogue/threadblock/fusion/visitor_load.hpp
```diff
@@ -403,10 +403,29 @@ struct VisitorRowBroadcast {
...
+        // In this case we are loading from a scalar and broadcasting
+        VecType filled_vec;
+        CUTLASS_PRAGMA_UNROLL
+        for (int i = 0; i < VecLength; i++) {
+          reinterpret_cast<Element*>(&filled_vec)[i] = params_ptr->null_default;
+        }
+
+        CUTLASS_PRAGMA_UNROLL
+        for (int i = 0; i < size(src_v); ++i) {
+          if(get<1>(coord_v(i)) < n)
```
thakkarV, line 424: Nit: spacing `if (get`

include/cutlass/epilogue/threadblock/fusion/visitor_load.hpp
```diff
@@ -524,12 +543,27 @@ struct VisitorColBroadcast {
...
+
       Tensor pred = make_tensor<bool>(shape(tC_gCol));
       CUTLASS_PRAGMA_UNROLL
       for (int i = 0; i < size(pred); ++i) {
         pred(i) = get<0>(tC_cCol(i)) < m;
       }
-      copy_if(pred, tC_gCol, tC_rCol);
+
+      if (params_ptr->ptr_col) {
+        // In this case we are loading from a column vector and broadcasting
```
thakkarV, line 554: A design question: this isn't really a scalar operation anymore. Does it make sense to extend this visitor, or to add replace this with a vector broadcast instead that then has a broadcasting layout for its data

tlrmchlsmth: >replace this with a vector broadcast instead that then has a broadcasting layout for its data

Could you point me to an example of this?

tlrmchlsmth: thanks for taking a look at the PR BTW :)

hwu36:
@apuaaChen

apuaaChen:
Hi @tlrmchlsmth, thanks for the PR! One question I have is that can we use the `VisitorScalarBroadcast` to achieve the same target? It also takes a scalar (e.g. float) and broadcast to the whole epilogue tile.

tlrmchlsmth:
@apuaaChen We could totally do that, but then in order to have a kernel for every case of fp8 quantized GEMM that we need to support, this is 4x the number of kernels. The activations can have per-tensor or per-token scales and weights can have per-tensor or per-output channel scales. So this PR lets us pick a another point in the binary size/performance tradeoff space.

apuaaChen:
@tlrmchlsmth Got it! Let me merge it. Thanks for the explanation.

ProExpertProg:
@apuaaChen while @tlrmchlsmth ended up using a custom visitor that loads both a row and a scalar from the `float*` argument, I ran into this use-case when the scalar is a known constant (0 bias). Again the benefit is reducing code size by having one kernel handle both cases. Could we get this PR merged?

ProExpertProg:
I guess could you let me know if you plan to merge it, or if there's any cleanup you want me to do before we merge. I also have a version with a boolean `EnableNullptr` parameter (default `false`) that enables the new scalar behavior which is consistent with the c3x epilogues. Let me know if I should push that to this branch.

apuaaChen:
@ProExpertProg Please push your changes to this branch. I will first merge your updates to our internal repo. After the CI is passed, I can get your PR merged, thanks!

ProExpertProg:
Perfect, thank you!!

ProExpertProg:
And please don't hesitate to ask for any changes or improved comments, and feel free to make edits yourself if there are any style/formatting issues.

ProExpertProg:
@apuaaChen were you able to get the PR run on the internal CI?

apuaaChen:
> @apuaaChen were you able to get the PR run on the internal CI?

Yes，It passed the internal CI. I’m combining it with a few other fixes right now

PR #1604, opened 2024-06-27
Add Ampere GEMM example using Cute and CUTLASS 3.x

aacostadiaz:
This pull request adds a GEMM example for the NVIDIA Ampere architecture, using Cute and CUTLASS 3.x. This example demonstrates how to create a GEMM kernel using the Cute components defined for SM80 and the `Collective MMA` and `Collective Epilogue` APIs provided in CUTLASS 3.x.

### DIFF
--- a/examples/14_ampere_tf32_tensorop_gemm/CMakeLists.txt
+++ b/examples/14_ampere_tf32_tensorop_gemm/CMakeLists.txt
@@ -33,3 +33,8 @@ cutlass_example_add_executable(
   ampere_tf32_tensorop_gemm.cu
   )
 
+cutlass_example_add_executable(
+  14_ampere_tf32_tensorop_gemm_cute
+  ampere_tf32_tensorop_gemm_cute.cu
+  )
+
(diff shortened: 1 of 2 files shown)


hwu36:
@thakkarV @ccecka

thakkarV:
this looks like a copy paste of the version we already have in the test dir. may I ask why you want to have this in the examples dir?

aacostadiaz:
> this looks like a copy paste of the version we already have in the test dir. may I ask why you want to have this in the examples dir? we already have example 59 which does this

Hi @thakkarV,

Thank you for your feedback.

This example is meant to complement the version in the test directory and provide more visibility for users who may not explore that directory.

It serves as an entry-level example, similar to example 14 for Ampere with Cutlass 2.x and examples 48 and 49 for Hopper on Cutlass 3.x. In addition, it allows users to easily experiment with different GEMM configurations, enabling them to tune GEMM to their needs. I could not find an easy way to do that without implementing this example.

Example 59 is much more complicated than what I'm adding here. I don't think it targets the same users or use case.

I think it is a valuable contribution that we have found useful for us and could be for others.

aacostadiaz:
Hi @thakkarV, yes. I still would like this to get merged. I accidentally closed it, sorry.

PR #1632, opened 2024-07-15
Make integer_subbyte Fully Compliant with is_integral

osayamenja:
## Context
See #1631 

## Change
Adds integral type trait for `integer_subbyte`

### DIFF
--- a/include/cute/numeric/int.hpp
+++ b/include/cute/numeric/int.hpp
@@ -36,6 +36,8 @@
 #include <cstdint>
 #endif
 
+#include <cute/numeric/integral_constant.hpp>
+#include <cutlass/integer_subbyte.h>
 #include <cutlass/numeric_types.h>
 
 namespace cute
@@ -100,5 +102,8 @@ using uint_byte = uint_bit<8*N>;
 
 template <int N>
 using uint_byte_t = typename uint_byte<N>::type;
+
+template<int Bits, bool Signed>
+struct is_integral<cutlass::integer_subbyte<Bits, Signed>> : true_type {};
 
 } // namespace cute


osayamenja:
@thakkarV Closing the loop on #1631, let me know if this is not actually needed and I will close this instead.

include/cute/numeric/int.hpp
```diff
@@ -101,4 +103,7 @@ using uint_byte = uint_bit<8*N>;
 template <int N>
 using uint_byte_t = typename uint_byte<N>::type;
 
+template<int Bits, bool Signed>
+struct is_integral<cutlass::integer_subbyte<Bits, Signed>> : true_type {};
```
thakkarV, line 107: add one for unsigned too?

thakkarV:
@ccecka for review too

include/cute/numeric/int.hpp
```diff
@@ -36,6 +36,8 @@
 #include <cstdint>
 #endif
 
+#include <cute/numeric/integral_constant.hpp>
+#include <cutlass/integer_subbyte.h>
```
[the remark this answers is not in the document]
osayamenja: Double-checked and the second isn't but the first is, due to `is_integral`. I will remove the first. My IDE (CLion) is very pedantic sometimes, that's why I initially added.

include/cute/numeric/int.hpp
osayamenja: The template covers both since unsigned is `Signed=false`, see [here](https://github.com/NVIDIA/cutlass/blob/56b46e2d13875b46b8f6a03f9f5ac91e2bfdc01a/include/cutlass/integer_subbyte.h#L155C1-L155C44)

osayamenja:
If needed, I can also register float8 and half for `is_floating_point`

PR #1638, opened 2024-07-16
Make mainloop schedule type available as `GemmKernel::Schedule`

manishucsd:
This PR makes mainloop schedule uniformly available for mainloop variants. Our usage of this type involves additionally having it as a part of `GemmDescription` object for our kernels and decide mainloop variant at the dispatch time. 

This was already there in [sm90_gemm_array_tma_warpspecialized_cooperative](https://github.com/NVIDIA/cutlass/blob/main/include/cutlass/gemm/kernel/sm90_gemm_array_tma_warpspecialized_cooperative.hpp#L87)

cc: @IonThruster , @thakkarV , @hwu36 , @mnicely

### DIFF
--- a/tools/library/include/cutlass/library/descriptions.h
+++ b/tools/library/include/cutlass/library/descriptions.h
@@ -34,7 +34,7 @@
 #include <cutlass/library/types.h>
 #include <cutlass/blas3_types.h>
 #include <cutlass/gemm_coord.h>
-
+#include <stdexcept>
 /////////////////////////////////////////////////////////////////////////////////////////////////
 
 namespace cutlass {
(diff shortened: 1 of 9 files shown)


tools/library/include/cutlass/library/descriptions.h
```diff
@@ -34,7 +34,7 @@
 #include <cutlass/library/types.h>
 #include <cutlass/blas3_types.h>
 #include <cutlass/gemm_coord.h>
-
+#include <stdexcept>
```
manishucsd, line 37: Include for `std::runtime_error` used in this file. This will allow us to include this file in other projects without have `#include <stdexcept>` above its include.

PR #1641, opened 2024-07-17
Fix MMA promotion interval assertions

LyricZhao:
For `BLOCK_SIZE_K=256`, `GmmaFP8Accumulation` has `accum_promotion_interval=4` but `mma_count_per_mainloop_iteration=8`, which makes a non-FP8-fast-accum kernel never promote to FP32 accumulators. This PR fixes the wrong assertion by changing `4` into the real number of MMA instructions issued.


LyricZhao:
Anyone replies to this? I do think it's a serious bug, making `BLOCK_SIZE_K=256` made FP8 training loss curve much worse than non-FP8-fast-accum.

thakkarV:
@IonThruster

thakkarV:
@yzhaiustc can we please put it on the list for 3.6?

yzhaiustc:
> @yzhaiustc can we please put it on the list for 3.6?

sure.

hwu36 (approved):

hwu36:
@manishucsd

PR #1647, opened 2024-07-22
Initial implementation of SYCL support for Cutlass 3

aacostadiaz:
This PR is the first step towards enabling Cutlass 3 and Cute to build with SYCL support. This enhancement will significantly expand the compatibility and flexibility of Cutlass, allowing users to build and run Cutlass on a wider range of SYCL-enabled devices.

Key Changes:

- CMake Configuration: Modified the CMake configuration to introduce an alternative build using SYCL. Detailed instructions on how to build with SYCL, including any necessary prerequisites and configuration steps, have been added to the `media/docs` folder.

- Code Migration: Five examples in Cute tutorial have been migrated to SYCL with minimal modifications. Most changes involved adapting CUDA-specific code to be compatible with SYCL, ensuring minimal disruption to the existing codebase while maintaining performance and functionality. You will notice that the examples for SYCL have been added as separate files instead of reusing the existing examples. This is on purpose to reduce disruption in the actual codebase and keep the tutorial example simple and educational. The existing examples contain CUDA-specific code that need to be generalised for SYCL. Other than the CUDA-specific code, the examples are identical.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,8 +26,12 @@
 # OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-cmake_minimum_required(VERSION 3.19 FATAL_ERROR)
-cmake_policy(SET CMP0112 NEW)
+if(CUTLASS_ENABLE_SYCL)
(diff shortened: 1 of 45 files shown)


mnicely:
We have no plans to support 3rd party platforms.

mehdi-goli:
Hi @mnicely,

Thank you for the quick response. We are more than happy to take on the maintenance and support for all non-CUDA-related aspects of the project. Our team is committed to making ongoing contributions to the project.

Please let us know what is the best way to coordinate this effort.

Issue #1633, opened 2024-07-15
[BUG] ElementC=void kernel reads non-void in `GemmDescription`


manishucsd:
I am observing gemm_desc.C.element = `bf16`, when I set it as `void`.

Please use the following [debug_branch](https://github.com/NVIDIA/cutlass/commit/08d11c55797bcd2a1982edf42b04143b3ccb5ead)  

Please check if the below print is expected.
**description_.C.element bf16       // Is this expected?**

We are choosing incorrect kernel because of this, is this a bug or I am messing something up?

```
./tools/profiler/cutlass_profiler --kernels=cutlass3x_sm90_tensorop_s64x128x32gemm_e4m3_e4m3_f32_void_bf16_128x128x128_1x2x1_0_tnn_align16_warpspecialized_epi_nosmem
description_.C.element bf16       // Is this expected?
description_.C.element bf16
description_.C.element bf16
description_.C.element bf16
description_.C.element bf16
description_.C.element bf16
description_.C.element bf16
description_.C.element bf16



=============================
  Problem ID: 1

        Provider: CUTLASS
   OperationKind: gemm
       Operation: cutlass3x_sm90_tensorop_s64x128x32gemm_e4m3_e4m3_f32_void_bf16_128x128x128_1x2x1_0_tnn_align16_warpspecialized_epi_nosmem

          Status: Success
    Verification: ON
     Disposition: Passed

reference_device: Passed
          cuBLAS: Not run
           cuDNN: Not run

       Arguments: --gemm_kind=universal --m=1024 --n=1024 --k=1024 --A=fe4m3:row --B=fe4m3:column --C=bf16:column --D=bf16:column  \
                  --alpha=1 --beta=0 --split_k_mode=serial --split_k_slices=1 --batch_count=1 --raster_order=heuristic  \
                  --op_class=tensorop --accum=f32 --cta_m=128 --cta_n=128 --cta_k=128 --cluster_m=1 --cluster_n=2 --cluster_k=1  \
                  --stages=7 --warps_m=4 --warps_n=1 --warps_k=1 --inst_m=64 --inst_n=128 --inst_k=32 --min_cc=90 --max_cc=90  \
                 

           Bytes: 4194304  bytes
           FLOPs: 2149580800  flops
           FLOPs/Byte: 512

         Runtime: 0.0113731  ms
          Memory: 343.463 GiB/s

            Math: 189005 GFLOP/s


=============================

CSV Results:
```

[...]

PR #1661, opened 2024-07-30
Skip void-C kernels in the profiler when beta is non zero

alihassanijr:
CUTLASS profiler will only skip disposition for void-C kernels when beta is non zero, when it makes more sense to skip running it in the first place.

Not all users are aware of void-C kernels (as far as I know it wasn't a thing in 2.X), and not everyone remembers to filter out voidC kernels when running the profiler with a non zero beta.

The easiest solution (and as far as I can tell correct way of handling this) is that `can_implement` return `false` when beta is non zero (or whatever argument indicates an epilogue source) but we have a void-C kernel.

Profiler already includes functionality to skip running kernels that fail `can_implement`.

CC @thakkarV  @hwu36

### DIFF
--- a/include/cutlass/epilogue/collective/detail.hpp
+++ b/include/cutlass/epilogue/collective/detail.hpp
@@ -392,6 +392,27 @@ public:
   tensormaps_fence_acquire([[maybe_unused]] cute::TmaDescriptor const* tensormap) { }
 };
 
+// SFINAE helpers for detecting beta/beta_ptr in EVT arguments.
+template <class Arguments, class = void>
+struct has_beta {
+  static constexpr bool value = false;
+};
+
+template <class Arguments>
+struct has_beta<Arguments, cute::void_t<decltype(Arguments{}.thread.beta)>> {
+  static constexpr bool value = true;
+};
+
+template <class Arguments, class = void>
+struct has_beta_ptr {
+  static constexpr bool value = false;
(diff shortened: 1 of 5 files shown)


thakkarV:
@manishucsd  CC

manishucsd:
@alihassanijr thanks for the fix. 

Can we verify that the issue #1633 is fixed with this change? I think ElementSource will need to be available and set in GEMM device-level template, then used in struct `GemmDescription` object to set the `gemmm_desc.element_c` correctly.

alihassanijr:
@manishucsd just to clarify, the issue in #1663 was that the profiler still had a non-void element C?
If so, that should be fine because the profiler has to support both voidC and non-voidC kernels (at least I think).

But with these changes, the profiler should skip voidC kernels when beta is non-zero, and any voidC kernel should fail `can_implement` if a beta or beta device pointer has been specified.

Does that resolve your issue?

hwu36 (approved):

manishucsd:
> @manishucsd just to clarify, the issue in #1663 was that the profiler still had a non-void element C?

The gemm_description object's element_c is getting set to wrong value. We expect gemm_description element_c to set to void. It is not fine as we use gemm_description to choose the voidC kernel while it had bf16. Can you repro using the debug_branch I created in the issue #1663

PR #1664, opened 2024-07-30
Build hotfix for SM 5.2 on 3.5.1

eqy:
`hrsqrt` doesn't seem to be defined for SM 5.2

### DIFF
--- a/include/cutlass/functional.h
+++ b/include/cutlass/functional.h
@@ -234,7 +234,7 @@ template <>
 struct inverse_square_root<half_t> {
   CUTLASS_HOST_DEVICE
   half_t operator()(half_t const &lhs) const {
-#if defined(__CUDA_ARCH__)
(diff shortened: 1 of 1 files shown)


thakkarV:
We no longer support Maxwell generation GPUs in cutlass so technically this falls outside of what our CI checks for. But I think this is blocking PyTorch adoption right? @hwu36 @jackkosaian

eqy:
That's correct, this seems a bit faster than convincing PyTorch CI to move off of SM 5.2 CI runners or gating CUTLASS in all compilation units on `__CUDA_ARCH__` CC @skylion007 @drisspg

Skylion007:
Yeah, this was the main blocker for updating CUTLASS to 3.5.0 as well and is blocking updating our FlashAttn implementation.

thakkarV:
@hwu36 @kerrmudgeon can we merge this please?

hwu36 (approved):

PR #1665, opened 2024-07-31
Stamp out 1x1x1 clusters, 128x256 CTA shape

alihassanijr:
Adds 128x256 tile shapes to FP16/BF16 and FP8 generators. Also adds 1x1x1 clusters to all existing FP16/BF16/FP8 generators.

NOTE: it is important to set kernel filter (--kernels / CUTLASS_LIBRARY_KERNELS) to a non empty string and skip pruning to get all of the new configurations.

If profiling exhaustively, they can be set to `*`.

Number of CUTLASS 3.X GEMMs before this commit: 2868
Number of CUTLASS 3.X GEMMs after this commit: 4016

CC @thakkarV @IonThruster @depaulmillz


thakkarV:
@hwu36 this is ready for merge

hwu36 (approved):

Issue #1672, opened 2024-08-01
[FEA] Passing the copy type as a parameter to `copy`


ZelboK:
Kind of a small change.
So I was looking at 
https://github.com/NVIDIA/cutlass/issues/1231

and I was wondering if it made sense to refactor the code so that it will accept the type of copy they want to rely on like so 
`copy(gmem_tiled_copy, SM80_CP_ASYNC_CACHEALWAYS<float>{}, tAgA, tAsA)` so that way you can have the user be aware of what's going on internally. Right now if you use `DefaultCopy` for example it'll dispatch to that instruction which requires the code to have more synchronization which is a little unintuitive. 

You might be able to get the copy type inferred from `gmem_tiled_copy` but passing it as an explicit parameter is probably much easier. 

I'm wondering if there's any flaws/problems with doing this? Perhaps the default should always be `copy(gmem_tiled_copy, DefaultCopy{}, tAgA, tAsA)`. 

I could work on a PR for implementing this if it's beneficial. Would like to know ahead of time if this is even desired or if there are problems with this approach before I commit my time to a PR though.

thakkarV:
You can use `UniversalCopy` as a means of always dispatching to the SIMT load/store rather than cp async copies. A deeper refactor of this automatic dispatch is planned but won't land anytime soon.

thakkarV:
If you would like to work on a PR, that would be lovely! @ccecka CC

thakkarV:
I suspect much of this can be accomplished by just renaming `UniversalCopy` to be the new `DefaultCopy` so that it never goes through automatic LDGSTS route again

ZelboK:
Sure, I'll push up a PR today. I'll see if just replacing the defaultcopy to be universalcopy is enough.

PR #1674, opened 2024-08-01
Avoid LDGSTS routing by changing default copy to be universalcopy

ZelboK:
https://github.com/NVIDIA/cutlass/issues/1672

This PR changes the default copy to be `UniversalCopy` so the LDGSTS instruction is avoided, and downstream users will need to specify the copy type if they want to use it which is more intuitive. 

Note: I imagine since all of the tests and examples that are configured to use `DefaultCopy` will need to now transition over to `AutoVectorizingCopyWithAssumedAlignment<128>` as that was the previous copy type. This way we can preserve behavior across benchmarks and tests. Before I make a bunch of changes across files though I'd like to get feedback now incase I'm missing anything.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
 # PyCache files
 __pycache__/
 cutlass_library.egg-info/
+build/
+.vscode/
--- a/examples/50_hopper_gemm_with_epilogue_swizzle/50_hopper_gemm_with_epilogue_swizzle.cu
+++ b/examples/50_hopper_gemm_with_epilogue_swizzle/50_hopper_gemm_with_epilogue_swizzle.cu
@@ -485,7 +485,7 @@ int main(int argc, char const **args) {
   // Tiled copy from Smem to Registers
   // Note : CuTe will vectorize this copy if the tiling + swizzling above were right
   using TiledCopyS2R = TiledCopy<
-                         Copy_Atom<DefaultCopy, ElementAcc>,
+                         Copy_Atom<AutoVectorizingCopyWithAssumedAlignment<128>, ElementAcc>,
                          Layout< Shape<_128,_16>, 
                                  Stride<_16,_1>>,
                          TileShapeS2R>;
@@ -496,9 +496,9 @@ int main(int argc, char const **args) {
       cutlass::gemm::TagToStrideC_t<LayoutD>,
       cutlass::epilogue::thread::LinearCombination<int32_t, 1, int32_t, int32_t>,
       SmemLayout,
-      Copy_Atom<DefaultCopy, ElementAcc>,
+      Copy_Atom<AutoVectorizingCopyWithAssumedAlignment<128>, ElementAcc>,
       TiledCopyS2R,
-      Copy_Atom<DefaultCopy, ElementOutput>>>;
+      Copy_Atom<AutoVectorizingCopyWithAssumedAlignment<128>, ElementOutput>>>;
 
   //
   // Assembling the GemmKernel
--- a/test/unit/cute/volta/vectorization_auto.cu
+++ b/test/unit/cute/volta/vectorization_auto.cu
@@ -109,7 +109,7 @@ template <class T, class GmemLayout, class RmemTiler>
 void
 test_copy_vectorization(GmemLayout gmem_layout, RmemTiler rmem_tiler)
 {
-  test_copy_vectorization<T>(DefaultCopy{}, gmem_layout, rmem_tiler);
+  test_copy_vectorization<T>(AutoVectorizingCopyWithAssumedAlignment<128>{}, gmem_layout, rmem_tiler);
 }
 
 TEST(SM70_CuTe_Volta, SimpleVec)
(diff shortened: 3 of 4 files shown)


.gitignore
```diff
@@ -1,3 +1,5 @@
 # PyCache files
 __pycache__/
-cutlass_library.egg-info/
\ No newline at end of file
+cutlass_library.egg-info/
+build/
+.vscode/
```
ZelboK, line 5: I'm hoping this is okay to add? I've a habit of just doing `git add .` for small repos and it can get annoying having to deal with me accidently committing my build directory.

examples/50_hopper_gemm_with_epilogue_swizzle/50_hopper_gemm_with_epilogue_swizzle.cu
```diff
@@ -485,7 +485,7 @@ int main(int argc, char const **args) {
   // Tiled copy from Smem to Registers
   // Note : CuTe will vectorize this copy if the tiling + swizzling above were right
   using TiledCopyS2R = TiledCopy<
-                         Copy_Atom<DefaultCopy, ElementAcc>,
+                         Copy_Atom<AutoVectorizingCopyWithAssumedAlignment<128>, ElementAcc>,
```
ZelboK, line 488: Since the DefaultCopy used to be `AutoVectorizingCopyWithAssumedAlignment<128>`, it must now be changed to explicitly specify this to preserve behavior. I believe this is the easiest way to reduce the scope of this PR and keep changes minimal.

test/unit/cute/volta/vectorization_auto.cu
```diff
@@ -109,7 +109,7 @@ template <class T, class GmemLayout, class RmemTiler>
 void
 test_copy_vectorization(GmemLayout gmem_layout, RmemTiler rmem_tiler)
 {
-  test_copy_vectorization<T>(DefaultCopy{}, gmem_layout, rmem_tiler);
+  test_copy_vectorization<T>(AutoVectorizingCopyWithAssumedAlignment<128>{}, gmem_layout, rmem_tiler);
```
ZelboK, line 112: Since the DefaultCopy used to be `AutoVectorizingCopyWithAssumedAlignment<128>`, it must now be changed to explicitly specify this to preserve behavior. I believe this is the easiest way to reduce the scope of this PR and keep changes minimal.

ZelboK:
cc @thakkarV 

Any ideas on other places of CUTLASS that might need help? Perhaps any Ampere related issues?

thakkarV:
@ccecka I like this change. wdyt?

ccecka:
Or (un)setting `CUTE_ARCH_CP_ASYNC_SM80_ENABLED` so that `select_elementwise_copy` always chooses `UniversalCopy`.

All of these changes look very dangerous to me.

ZelboK:
> Or (un)setting `CUTE_ARCH_CP_ASYNC_SM80_ENABLED` so that `select_elementwise_copy` always chooses `UniversalCopy`.

The problem as I understood it was that logically it's a bit unintuitive that `DefaultCopy` dispatches to this instruction because the user doesn't actually know that it's doing this unless they inspect under the hood. Wouldn't unsetting this be equivalent to the user just specifying `UniversalCopy` in the first place? They have to know what's going on in the first place, and it doesn't prevent them from having to discover they need more synchronizations.  

> All of these changes look very dangerous to me.

I'm probably missing some context here. Could you give an example situation of where this change could go wrong? (Might be a stupid question 😅 )

ZelboK:
> how about removing `select_elementwise_copy` and just using `UniversalCopy<SrcType,DstType>{}` by default to avoid LDGSTS?

I think this could work too, actually, IIUC if downstream code specifies that it wants to do a specific async cp it'll go through a different codepath from this anyway. Also won't need to change the default type so less code changes overall.

thakkarV:
@ccecka and I had a conversation about this internally, and have some thoughts. 

1. we likely do not want to disable this auto LDGSTS. This is because it is only enabled when SM80 is set
2. when the source tensor is tagged to have a gmem pointer
3. when the destination tensor is tagged to have an smem pointer

As for the MR itself, you are replacing a copy with a universal copy with assumed alignment of 128b. This is very dangerous because not all tensors may have that alignment.

We are working on a nicer fix of this internally that entails adding a simple assignment copy type that is only used when no atom is specified.

ZelboK:
> As for the MR itself, you are replacing a copy with a universal copy with assumed alignment of 128b. This is very dangerous because not all tensors may have that alignment.

I actually made a mistake I believe and I meant to do `AutoVectorizingCopyWithAssumedAlignment<8>;` and not 128... That is my bad. With that being said I'm curious to know if I'm missing any context - the `DefaultCopy` is set to  `AutoVectorizingCopyWithAssumedAlignment<8>;`, right? If all `DefaultCopy`s in the codebase were replaced with this explicit type why would it be dangerous? Or was that only the case because in the code I mistakenly put 128 bits?

PR #1680, opened 2024-08-02
Add CLayout_64x208

tridao:
Without this I get compilation error when the extended shapes are enabled


thakkarV:
@klevzoff CC

sklevtsov-nvidia:
Slipped through the cracks when porting over to gh :disappointed: 

Thanks for the fix @tridao!

hwu36 (approved):

PR #1695, opened 2024-08-08
Instantiate More CUTLASS 3.x GEMM Kernels

depaulmillz:
- Generator builds with 64x256xK math instruction shapes for fp16, bf16, and fp8 GEMMs.
- Now instantiates 5476 CUTLASS 3.x GEMM kernels when only 4016 were instantiated before.


depaulmillz:
@hwu36 @thakkarV @ANIKET-SHIVAM

hwu36 (approved):

PR #1796, opened 2024-09-09
Prefix a member template name with the template keyword.

shumway:
Fixes llvm buld error.

### DIFF
--- a/include/cutlass/epilogue/collective/sm90_epilogue_tma_warpspecialized.hpp
+++ b/include/cutlass/epilogue/collective/sm90_epilogue_tma_warpspecialized.hpp
@@ -640,7 +640,7 @@ public:
                       tRS_rC,
                       thread_idx
                     );
-    auto cst_callbacks = fusion_callbacks.get_consumer_store_callbacks<RefSrc>(cst_args);
+    auto cst_callbacks = fusion_callbacks.template get_consumer_store_callbacks<RefSrc>(cst_args);
     bool is_producer_load_needed = fusion_callbacks.is_producer_load_needed();
     bool is_C_load_needed = is_source_supported && fusion_callbacks.is_C_load_needed();
 


shumway:
cc @thakkarV @hwu36

thakkarV:
CC @yzhaiustc

hwu36:
Thank you for working on this.  It is important to us too.  May I ask what is your plan of your project?

yzhaiustc:
@shumway can you please paste your error message where you observed the issue?

we have many cst call back entrances. did you see similar issues at other call back functions?

Artem-B:
@shumway It would be useful to add a compilation test for this, or provide a code snippet triggering the problem.

Also, I believe the issue happens w/ clang. It's possible that NVCC may handle it somewhat differently.

shumway:
I only see this compilation error within Google, on our internal LLVM-based compiler. I have not had this problem with the same code with NVCC on Ubuntu.

When compiling a mixed-type GEMM that uses cutlass::arch::Sm90 and cutlass::epilogue::TmaWarpSpecializedCooperative in my experimental directory ([cl/672547324](https://cl.corp.google.com/672547324)) on an LLVM compiler, I got the following build time error:

In file included from experimental/users/shumway/cutlass/mixed_gemm.cu.cc:9:
In file included from ./third_party/gpus/cutlass/include/cutlass/epilogue/collective/collective_builder.hpp:116:
In file included from ./third_party/gpus/cutlass/include/cutlass/epilogue/collective/builders/sm90_builder.inl:41:
In file included from third_party/gpus/cutlass/include/cutlass/epilogue/collective/collective_epilogue.hpp:60:
third_party/gpus/cutlass/include/cutlass/epilogue/collective/sm90_epilogue_tma_warpspecialized.hpp:616:43: error: reference to non-static member function must be called
  616 |     auto cst_callbacks = fusion_callbacks.get_consumer_store_callbacks<RefSrc>(cst_args);

My understanding of this error:

1. The type-alias `FusionCallbacks` is set by a template parameter in CollectiveEpilogue.
2. Thus, the LLVM compiler does not assume that FusionCallbacks::get_consumer_store_callbacks is a templated method.
3. The LLVM compiler misinterprets this line of code to be a less-than comparison between a non-static member function reference `fusion_callbacks.get_consumer_store_callbacks` and `RefSrc`.
4. The LLVM compiler reports an error saying that the non-static member function reference was used without arguments.

The C++ syntactic fix of adding the `template` keyword before the method name tells LLVM to interpret the `<` symbol as part of the template argument `<RefSrc>`, and the line parses correctly.

Because I haven't seen this error in nvcc, I'm not sure how to create a failing test case for this change. This is the only example of this error that I've found, but this is my first attempt to compile v3 code like this with our compiler. This was the only code fix I needed to get a working GEMM example on this LLVM compiler, but we may find other issues as we experiment more.

yzhaiustc:
@mhoemmen how do you think of this issue?

shumway:
My understanding is that GCC tends to be more forgiving of omitting the odd-looking "template" keyword in templated member-function calls, though I remember having to sometimes adding it to heavily-templated GCC code in the past. I think the C++ standard requires the template keyword here, but compilers may infer intent. Adding the template keyword is syntactically correct and shouldn't cause any harm. I generally expect to see some small issues like this when compiling templated C++ libraries on different compilers.

Let me know your preferred process for these kind of compiler issues, since it's likely I'll encounter this other places. I already have a local patch, so this PR is just to improve the compatibility of CUTLASS with Google's internal build tools.

mhoemmen:
@yzhaiustc Thanks for letting me know about this!  I'd like to check carefully whether this is valid C++.  While I think most compilers should accept this, it's a bit weird to have to use the "`template`" disambiguator when the template argument (`RefSrc` here) is not a dependent type (it's always `true` and that's spelled out in the code).  This feels almost like an MSVC kind of build error, where the compiler is excessively conservative about whether a type is dependent.

mhoemmen:
[cppreference](https://en.cppreference.com/w/cpp/language/dependent_name) claims that leaving the template disambiguator in place would not be incorrect.

> As is the case with `typename`, the `template` prefix is allowed even if the name is not dependent or the use does not appear in the scope of a template.

I haven't found the Standardese for that yet.  The rules for when `template` or `typename` are required changed in C++20, so I would need to dig out a C++17 Standard instead of looking at the current Working Draft.  On the other hand, I'm happy to accept cppreference as an authority here.

While I don't think `template` should be *required* here, it should be harmless to add.  Thus, I'm happy to approve this PR.  We don't have a way to test this internal development LLVM version.  Thus, given that this change (though harmless) might indicate an LLVM bug, I'd rather not try to do more than whatever this PR suggests.

Thanks for the contribution!

mhoemmen (approved):

hwu36 (approved):

PR #1820, opened 2024-09-16
Add missing 60_cutlass_import to CMakeLists.txt

joerowell:
It turns out that `60_cutlass_import` has never been built as an example since it was introduced.


d-k-b:
That is correct, @joerowell . That example requires an existing CUTLASS installation for cmake to find, and while it's possible to set that up in CMake using post-install scripting, it was error-prone enough that we decided to just leave it as an exercise to the user to try 😄 as the example seemed simple enough to review.

PR #1864, opened 2024-10-11
Add GMMA shape m64n40k16

tridao:
This GMMA shape is being used in FA3 backward pass for headdim 256 (tile size 64 x 80, split into 2 WGs).
cc @thakkarV

### DIFF
--- a/include/cute/arch/mma_sm90.hpp
+++ b/include/cute/arch/mma_sm90.hpp
@@ -843,6 +843,11 @@ ss_op_selector()
         return SM90::GMMA::MMA_64x48x16_F32F16F16_SS<MajorA, MajorB, Args...>{};
       }
 #endif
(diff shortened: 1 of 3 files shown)


thakkarV:
@klevzoff CC

sklevtsov-nvidia:
Hi @tridao, we are adding all supported GMMA shapes: https://github.com/NVIDIA/cutlass/pull/1890. Out of concern for compile times, we moved all "extended" instruction shapes into their own headers, which are conditionally included into `mma_sm90_gmma.hpp` and `mma_traits_sm90_gmma.hpp`. Nothing needs to change on your side - just compiling with `CUTLASS_ENABLE_SM90_EXTENDED_MMA_SHAPES` is enough.

However it may not be possible to merge both PRs because of conflict. Alternatively, @hwu36 if you merge this one first, I'm happy to rebase mine and resolve the conflict.

hwu36 (approved):

PR #1887, opened 2024-10-20
fix  Copy_Traits<SM80_CP_ASYNC_*_ZFILL> without  definition of with(pred) function

CalebDu:
Definition ```with(pred)``` func of ```Copy_Traits<SM80_CP_ASYNC_*_ZFILL> ``` is placed in wrong class, it causes ```pred ``` passed to SM80_CP_ASYNC_*_ZFILL always is ```false``` and cp_async always fill all zeros.

### DIFF
--- a/include/cute/atom/copy_traits_sm80.hpp
+++ b/include/cute/atom/copy_traits_sm80.hpp
@@ -52,12 +52,7 @@ struct Copy_Traits<SM80_CP_ASYNC_CACHEALWAYS<S,D>>
   // Reference map from (thr,val) to bit
   using RefLayout = SrcLayout;
 
-  // Construct a zfill variant with a given predicate value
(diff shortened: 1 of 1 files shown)


CalebDu:
@ccecka

ccecka:
I have my own MR of fixes for `copy` and this is included in that, including tracking down all of the places where `SM80_CP_ASYNC_*` is chosen instead of the intended `ZFILL` variant.

That will be getting merged soon and pushed out publicly, I hope.

CalebDu:
> I have my own MR of fixes for `copy` and this is included in that, including tracking down all of the places where `SM80_CP_ASYNC_*` is chosen instead of the intended `ZFILL` variant.
> 
> That will be getting merged soon and pushed out publicly, I hope.

I get it. looking forward your fix releasing soon

PR #1891, opened 2024-10-22
Make CuTe and PyCuTe idx2crd Behaviors The Same

leimao:
CuTe C++ has two function overloadings for `idx2crd`:
* [crd2idx(Coord const& coord, Shape const& shape)](https://github.com/NVIDIA/cutlass/blob/f7b19de32c5d1f3cedfc735c2849f12b537522ee/include/cute/stride.hpp#L146)
* [idx2crd(Index  const& idx, Shape  const& shape, Stride const& stride)](https://github.com/NVIDIA/cutlass/blob/f7b19de32c5d1f3cedfc735c2849f12b537522ee/include/cute/stride.hpp#L179)

However, PyCuTe only has one function for `idx2crd`:
* [idx2crd(idx, shape, stride=None)](https://github.com/NVIDIA/cutlass/blob/f7b19de32c5d1f3cedfc735c2849f12b537522ee/python/pycute/int_tuple.py#L147)

If default value `None` if used for the `idx2crd(idx, shape, stride=None)` in PyCuTe, the function behavior is different from the `crd2idx(Coord const& coord, Shape const& shape)` in CuTe C++.

Therefore, a new function `idx2crd(idx, shape)` was added to PyCuTe, and the default value was removed from `idx2crd(idx, shape, stride=None)`.

### DIFF
--- a/python/pycute/int_tuple.py
+++ b/python/pycute/int_tuple.py
@@ -144,7 +144,25 @@ def prefix_product(a, init=1):
       return init
 
 
-def idx2crd(idx, shape, stride=None):
+def idx2crd(idx, shape):
+
+  if is_tuple(idx):
+    if is_tuple(shape):                # tuple tuple tuple
+      assert len(idx) == len(shape)
+      return tuple(idx2crd(i, s) for i, s in zip(idx,shape))
+    else:                              # tuple "int" "int"
+      assert False           # Error
+  else:
+    if is_tuple(shape):                # "int" tuple tuple
+      tuple_idx = []
+      for i in range(len(shape)):
+        tuple_idx.append(idx2crd(idx // product(shape[:i]), shape[i]))
+      return tuple(tuple_idx)
+    else:                              # "int" "int" "int"
+      return idx % shape
+
+
+def idx2crd(idx, shape, stride):
   if stride is None:
     stride = prefix_product(shape)
 


leimao:
@hwu36

hwu36:
@ccecka

ccecka:
I actually recommend removing `idx2crd(idx, shape, stride)` completely, it is dangerous and only works when the strides are compact. Ideally, this would be removed from both python and C++.

This python implementation of `idx2crd(idx, shape)` also appears to be incorrect and inconsistent with the C++ version in the case that `idx >= size(shape)`.

manishucsd:
> I actually recommend removing `idx2crd(idx, shape, stride)` completely, it is dangerous and only works when the strides are compact. Ideally, this would be removed from both python and C++.

In the case of remove of `idx2crd` and it also being incorrect, what would be the best and correct way to move from idx space to crd space, using manual decomposition of col-wise idx into modes or some other way?

leimao:
> I actually recommend removing `idx2crd(idx, shape, stride)` completely, it is dangerous and only works when the strides are compact. Ideally, this would be removed from both python and C++.
> 
> This python implementation of `idx2crd(idx, shape)` also appears to be incorrect and inconsistent with the C++ version in the case that `idx >= size(shape)`.

I did not define `idx2crd(idx, shape)` for the cases that `id >= size(shape)`. I only defined the conversion for the isomorphism in Definition 2.3 in ["A note on the algebra of CuTe Layouts"](https://research.colfax-intl.com/wp-content/uploads/2024/01/layout_algebra.pdf). If you want the definition for the extended isomorphism, i.e., the cases `id >= size(shape)`, I can also accommodate it in the next patch.

ccecka:
> In the case of removal of idx2crd and it also being incorrect, what would be the best and correct way to move from idx space to crd space, using manual decomposition of col-wise idx into modes or some other way?

You simply can't do it in general (without solving certain Diophantine equations...).

> I did not define idx2crd(idx, shape) for the cases that id >= size(shape)

Yes, but those cases need to be consistent as well else you run into problems in defining `composition` and out-of-bounds tiling.

> BTW, why the PyCuTe functions are not Python bindings to the CuTe C++ functions?

It was a playful alternate implementation that got released for some reason. I still use it for prototyping or code golfing once in a while, but that's it.

manishucsd:
> BTW, why the PyCuTe functions are not Python bindings to the CuTe C++ functions?

Having a fully functional PyCuTe support same as CuTe C++ will be extremely useful. 

> You simply can't do it in general (without solving certain Diophantine equations...).

So the layout operator will take me from coordinate space to index space, is there a reliable way in CuTe C++ to go in the reverse direction from index space to coordinate space? Does CuTe C++ solve Diophantine equations?

ccecka:
No, there is no fully reliable way to go from index space to coordinate space. We now use Coordinate Tensors to track and manipulate coordinate spaces. This is how we perform predication generically among other applications. See 
https://github.com/NVIDIA/cutlass/blob/main/media/docs/cute/0y_predication.md
https://github.com/NVIDIA/cutlass/blob/main/media/docs/cute/0z_tma_tensors.md

leimao:
@ccecka So I guess my action item is to define `idx2crd(idx, shape)` for the cases that `id >= size(shape)` and add some unit tests, if you guys don't have a plan to add Python bindings for CuTe C++ in the near future?

PR #1907, opened 2024-10-29
Distributed GEMM

alihassanijr:
Adds experimental support for running tensor parallel GEMMs natively through CUTLASS.

Distributed GEMM (DistGEMM) implements communication-fused GEMMs using point-to-point communication, which allows for better pipelining, and theoretically can hide all communication behind computation. It also makes very few assumptions about the underlying kernel, and only adds a few barriers to the beginning of each GEMM kernel, and attempts to either use the epilogue source as the communication buffer, or a memcopy branch in the cuda graph, leaving SMs free for GEMMs, and the copy engine free for communication.

When benchmarked with Llama 70B and 405B training shapes, DistGEMM can reach 70-80% of peak performance.

A more detailed blog post on DistGEMM will be released soon.

### DIFF
--- a/include/cutlass/experimental/distributed/schedules/dist_gemm_1d_schedules.hpp
+++ b/include/cutlass/experimental/distributed/schedules/dist_gemm_1d_schedules.hpp
@@ -1,0 +1,149 @@
+/***************************************************************************************************
+ * Copyright (c) 2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: BSD-3-Clause
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ * list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ *
+ * 3. Neither the name of the copyright holder nor the names of its
+ * contributors may be used to endorse or promote products derived from
+ * this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ **************************************************************************************************/
+/*!
+  \file 1-D Distributed GEMM Schedules
+
+  Device/iteration mappings are defined with CuTe layouts, 
+  since they are functions from integers to integers as well.
+  
+  Each mapping is defined as a rank-3 layout:
+   First mode is the device mode, second mode is iteration, third mode is a "bias" term.
+   Mappings are defined as: (x * device_idx + y * iteration + z) % TP
+   Their CuTe layout equivalent would be: <<TP, TP, 1>, <x, y, z>>
+  
+  Assumption made in all schedules: TP == NumStages(Iterations)
+*/
+
+#pragma once
+
+#include "cute/layout.hpp"
+#include "cute/tensor.hpp"
+#include "cutlass/cutlass.h"
+
+#include "cutlass/experimental/distributed/schedules/dist_gemm_base_schedule.hpp"
+
+///////////////////////////////////////////////////////////////////////////////
+
+namespace cutlass::distributed::schedules {
+
+template <class TP_>
+struct ReduceScatter1D_TilingA_RotatingC: BaseSchedule<
+    TP_,
+    /* ProcessorTiler_ = */ cute::Shape<_1, _1, TP_, _1>,
+    /* IterationTiler_ = */ cute::Shape<TP_, _1, _1, _1>,
+    /* PeerDeviceMapping_ = */ cute::Layout<cute::Shape<TP_, TP_, _1>, cute::Stride<_1, _0, cute::C<-1>>>,                  // (left neighbor) = device_idx - 1
+    /* ProcessorMappingA_ = */ cute::Layout<cute::Shape<TP_, _1>, cute::Stride<_1, _0>>,                                    // (identity) = device_idx
+    /* ProcessorMappingB_ = */ cute::Layout<cute::Shape<TP_, _1>, cute::Stride<_1, _0>>,                                    // (identity) = device_idx
+    /* ProcessorMappingC_ = */ cute::Layout<cute::Shape<TP_, _1>, cute::Stride<_1, _0>>,                                    // (identity) = device_idx
+    /* ProcessorMappingD_ = */ cute::Layout<cute::Shape<TP_, _1>, cute::Stride<_1, _0>>,                                    // (identity) = device_idx
+    /* IterationMappingA_ = */ cute::Layout<cute::Shape<TP_, TP_, _1>, cute::Stride<_1, cute::C<-1>, cute::C<TP_{} - 1>>>,  // = device_idx - iter + TP - 1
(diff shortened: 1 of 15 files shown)


alihassanijr:
@thakkarV @hwu36

include/cutlass/experimental/distributed/schedules/dist_gemm_base_schedule.hpp
```diff
@@ -0,0 +1,577 @@
...
+
+template <typename ProblemShape>
+constexpr auto check_problem_shape(const ProblemShape & problem_shape) {
+  static_assert(cute::rank(ProblemShape{}) == 3 || cute::rank(ProblemShape{}) == 4,
+      "Expected rank-3 or rank-4 GEMM problem shape.");
+  if constexpr (cute::rank(ProblemShape{}) == 3) {
+    return append<4>(problem_shape, 1);
+  } else {
+    return problem_shape;
+  }
```
ccecka, line 60:
```suggestion
  return append<4>(problem_shape, 1);
```

include/cutlass/experimental/distributed/schedules/dist_gemm_1d_schedules.hpp
```diff
@@ -0,0 +1,149 @@
...
+template <class TP_>
+struct ReduceScatter1D_TilingA_RotatingC: BaseSchedule<
+    TP_,
+    /* ProcessorTiler_ = */ cute::Shape<_1, _1, TP_, _1>,
+    /* IterationTiler_ = */ cute::Shape<TP_, _1, _1, _1>,
+    /* PeerDeviceMapping_ = */ cute::Layout<cute::Shape<TP_, TP_, _1>, cute::Stride<_1, _0, cute::C<-1>>>,                  // (left neighbor) = device_idx - 1
+    /* ProcessorMappingA_ = */ cute::Layout<cute::Shape<TP_, _1>, cute::Stride<_1, _0>>,                                    // (identity) = device_idx
+    /* ProcessorMappingB_ = */ cute::Layout<cute::Shape<TP_, _1>, cute::Stride<_1, _0>>,                                    // (identity) = device_idx
+    /* ProcessorMappingC_ = */ cute::Layout<cute::Shape<TP_, _1>, cute::Stride<_1, _0>>,                                    // (identity) = device_idx
+    /* ProcessorMappingD_ = */ cute::Layout<cute::Shape<TP_, _1>, cute::Stride<_1, _0>>,                                    // (identity) = device_idx
```
ccecka, line 66: The "ProcessorMapping" is the same in all of these schedules and are used in-code very trivially.

(1) What do these parameterize and what other versions of this parameter do you expect to support?
(2) Why do they need to be CuTe Layouts? The sizes, shapes, ranks, strides are never used.
(3) The comments next to these parameters do not explain the domain of this function or the codomain.
(4) The size of the "bias mode" is always `1` and the stride is always `0`, yet these are indexed with a coordinate `1` in that mode always. This seems like a poor parameterization and `Layout`s are not what you actually want.

alihassanijr: You're right, ProcessorMapping is definitely unnecessary at this point. It was originally there in case the first iteration has a remote buffer, in which case they would map device index to the peer device's index.

I'll get rid of this; but just to clarify this is only in reference to ProcessorMapping and not IterationMapping?

ccecka: I'm still picking apart "IterationMapping"... it seems redundant with "PeerDeviceMapping", but that's still unclear to me. It appears that the relationships between MNKL are not respected because each of these layouts refers to A|B|C|D instead...

alihassanijr: PeerDeviceMapping can technically be represented with a boolean with the schedules implemented right now, but I didn't want to assume that it would necessarily stay that way with other schedules that we may add.

And correct, it's a little difficult to solely rely on the relationships of MNKL because operands are sharded and rotated differently in different schedules. Because DistGEMM is also supposed to have a separate buffer space for remote tensors and switch between them, it wasn't possible to maintain that behavior anymore.
I'll try and think of a better way to do this, but at some point, references to ABCD will be inevitable just because schedules shard them differently.

ccecka: And my point is that what is important is how the processors are moving through MNK space, which can then be translated to actions on ABC.

alihassanijr: I see, thanks for explaining. I think we could just replace `IterationMapping{A,B,C,D}` with mappings to MNKL tile coordinates. It would only make sense given that the tile shape (`IterationTiler`) is set up in size-4 tuples corresponding to MNKL anyway. Would moving to `IterationMapping{M,N,K,L}` (or if I can figure out the mapping for it just one `IterationMappingMNKL` layout) be a step in the right direction?

thakkarV: Partitioning is the job of the TV layout that maps the GPU rank to the values it extracts from the global coordinate. Scheduling adds the temporal component on top, which tells each GPU which coordinate to work on at which step of the computation. Separating them out into two makes the most sense. @ccecka agree?

dskhudia:
@alihassanijr : Thanks for writing the detailed writeup on this. https://blog.shi-labs.com/distributed-gemm-88be6a481e2b

I was wondering if it can work with 12.4 toolkit? Is 12.6 toolkit absolutely required?

alihassanijr:
@dskhudia thank you.

Re: CTK version, I'm actually not sure; I know 12.3 and earlier don't work because there's a few cuda graph APIs that DistGEMM relies on. I haven't been able to check exactly which CTK release is the earliest that supports those, I'll try and get that this week.
But it's simple enough to test if you've set it up; if it's incompatible, you'd get compilation errors from those missing cuda graph APIs.

dskhudia:
@alihassanijr : Thanks for the suggestion. I tried it with CTK 12.4 and it worked fine with the following results which roughly match with your results from the blog. 

```
  running DistGEMM...
  running DistGEMM finished without runtime errors

  Disposition (eps: 0): Passed
  Warming up for 10 iterations.
  Profiling for 100 iterations.

  TP: 8
  Problem Size: 16384 x 106496 x 16384 x 1
  Local GEMM Problem Size: 2048 x 13312 x 16384 x 1
  Avg runtime: 9.16037 ms
  TFLOPS: 780.19
  ```
  
  However, TP=4 segfaults (I did change `using TP = _8` to `using TP = _4`) so maybe somewhere TP=8 is assumed.

**Comment by alihassanijr** [comment]
I think CTK 12.4 should be fine then if there are no compile errors.

But the TP=4 issue is weird; we've tested different TP values extensively. I'll try it on my end and get back to you.

**Comment by alihassanijr** [comment]
TP=4 runs fine for me. I'll try and see if I can set up CTK 12.4 somewhere and try.

```
$ ./64_distributed_gemm --m=16384 --n=106496 --k=16384

  running DistGEMM...
  running DistGEMM finished without runtime errors

  Disposition (eps: 0): Passed
  Warming up for 10 iterations.
  Profiling for 100 iterations.

  TP: 4
  Problem Size: 16384 x 106496 x 16384 x 1
  Local GEMM Problem Size: 4096 x 26624 x 16384 x 1
  Avg runtime: 16.8623 ms
  TFLOPS: 847.67
```

**Comment by dskhudia** [comment]
The driver version I have is `Driver Version: 535.183.01`

**Comment by dskhudia** [comment]
I tried on another machine that has `Driver Version: 550.54.14` and it worked fine :-)

**Comment by alihassanijr** [comment]
Yes I was just about to post this: 

I think your driver could also be the problem. The [CTK version corresponding to 535 is 12.2](https://docs.nvidia.com/cuda/cuda-toolkit-release-notes/index.html#id5).

I just tried it with CTK 12.5 (still trying to get it working with 12.4 as well) on a different machine, and this one luckily has driver version 550, which corresponds to CTK 12.4, and it still works fine on my end.

**Comment by dskhudia** [comment]
@alihassanijr : Will this work for fp8 as well ? Anything extra needed?

**Comment by alihassanijr** [comment]
@dskhudia yes it does support FP8, and in theory should support any CUTLASS 3.X dense GEMM.

**Comment by dskhudia** [comment]
Thanks :) Will try with fp8.

**Comment by whatdhack** [comment]
At a superficial level, there appears to be overlap with framework ( e.g. PyTorch )  Distributed Tensor.   Any comment on how  each can benefit from the other ?

**Comment by alihassanijr** [comment]
@whatdhack I don't have too much information about Distributed Tensor, but based on my understanding, it would be a framework-level primitive for model parallelism techniques like Tensor Parallelism (TP). Distributed GEMM is more about how the Tensor Parallel GEMMs are implemented, so it would more serve as a backend, with things like Distributed Tensor being IR-level.

**Comment by christindbose** [comment]
Can this work on a 8xV100 machine assuming I use CUDA 12.4? @alihassanijr

**Comment by alihassanijr** [comment]
@christindbose sorry no, it's only implemented for CUTLASS 3.X kernels right now, which primarily limits it to Hopper.

**Comment by dskhudia** [comment]
@alihassanijr : I was going through the code and where is self_flag_ptr or self_flag_ptr_ used? Is this needed or it's captured by graph edge dependency that you have?

**Comment by christindbose** [comment]
hi @alihassanijr, could you please point me to an example command to build just the distributed GEMM kernel?

**Comment by alihassanijr** [comment]
@christindbose once you clone CUTLASS and as long as you have the requirements (CTK >= 12.4, python, git), you first need to set up the build for SM90a with GDC enabled (recommended for DistGEMM):

```
mkdir build && cd build
cmake .. -DCUTLASS_NVCC_ARCHS="90a" -DCUTLASS_ENABLE_GDC_FOR_SM90=1
```

Then you can simply just build the example without building all the targets:

```
cd examples/64_distributed_gemm     # build/examples/64_distributed_gemm
make
```

Without modification, the example builds a DistGEMM with the following schedule:
 * 1-D, TP=8 (requires 8 Hopper GPUs with any-to-any NVLink)
 * AllGathering A along the M mode (row-wise)
 
and its local GEMMs are based on the following CUTLASS kernel:
* Ping-ponging persistent WS kernel
* FP16 with FP16 accumulate,
* CTA Shape: <_128, _256, _64>
* Cluster Shape: <_1, _2, _1>

Note the above kernel was selected after profiling for the GEMM shapes reported in the blog post, so if you're planning on running a different use case, profiling again for that case is highly recommended to get the best performance.

alihassanijr:
Closing this since DistGEMM is merged with the 3.7 release.

Issue #1915, opened 2024-11-03
[QST] Is cutlass::bfloat16_t x cutlass::int2b_t GEMM possible?


areddy2022:
While looking at example 55 (cutlass/examples/55_hopper_mixed_dtype_gemm/55_hopper_int4_bf16_gemm.cu), I was curious whether this modification would be legal:

From:
`using MmaType = cutlass::bfloat16_t;
using QuantType = cutlass::int4b_t;`

To:
`using MmaType = cutlass::bfloat16_t;
using QuantType = cutlass::int2b_t;`

According to the README.md, for the example, "For 8-bit x 4-bit or 2-bit, both inputs must be K-major." However, the internal comment states, "Only supports INT4 x { FP16, BF16 }." Furthermore, I'm having trouble finding documentation in the library over the use of `int2b_t` datatype for use in GEMM. I apologize if this question needs to be more detailed or if I missed some part of the documentation.

IwakuraRein:
That comment is inaccurate and needs updating. This kernel always converts smaller dtypes to the MMA dtype with CUDA Cores, so as long as such conversion logic exists, the combination is legal. I just checked that INT2 x BF16 compiled successfully and passed the test.

However, for optimal performance, such conversion should be low-cost; otherwise, it may negate the benefits of using mixed dtypes. Please refer to the [PTX doc](https://docs.nvidia.com/cuda/parallel-thread-execution/#type-conversion) to verify if the hardware natively supports this conversion, also [this header](https://github.com/NVIDIA/cutlass/blob/main/include/cutlass/numeric_conversion.h) to see if an optimized software solution is available.

Currently, only naive conversion logic exists for `int2b_t`, i.e., casting first to `int` and then to `bfloat16_t`. This process makes the kernel significantly slower compared to INT4 x BF16.

Please let us know if this combination is important for your use case, and let us see if we can implement an optimized conversion for it.

areddy2022:
This would be a beneficial feature to have for extremely low-bit quantization (e.g. Bitnet or HQQ). 

If I were to try and implement the conversion myself, I would need to add code for numeric conversions in `numeric_conversion.h`?

IwakuraRein:
Yes. Feel free to implement it as a partial specialization of the `NumericArrayConverter`. Please refer to the [INT4 => BF16](https://github.com/NVIDIA/cutlass/blob/main/include/cutlass/numeric_conversion.h#L4112-L4236) as example.

[...]

PR #1926, opened 2024-11-07
add custom conversion from int2b_t -> bfloat16_t

CBepler:
create high speed conversion from int2b_t -> bfloat16_t similiar to that from int4b_t


areddy2022:
This is to directly resolve issue #1915

It should be noted that the scope for the `int2b_t `tensor only works with `scope_min` >= -1, and `scope_max` <= 1. When `scope_min` = 2, disposition fails when compiling the example with the `int2b_t` datatype. We are still debugging why.

PR #1942, opened 2024-11-13
Fix `cutlass` python library with cuda `12.6.2.post1`

danthe3rd:
Previously we had this error upon import:
```
  File "/storage/home/cutlass/python/cutlass/backend/operation.py", line 39, in <listcomp>
    _version_splits = [int(x) for x in __version__.split("rc")[0].split(".")]
                       ^^^^^^
ValueError: invalid literal for int() with base 10: 'post1'
```

### DIFF
--- a/python/cutlass_library/generator.py
+++ b/python/cutlass_library/generator.py
@@ -103,7 +103,7 @@ def CudaToolkitVersionSatisfies(semantic_ver_string, major, minor, patch = 0):
 
   # Update cuda_version based on parsed string
   if semantic_ver_string != '':
-    for i, x in enumerate([int(x) for x in semantic_ver_string.split('.')]):
+    for i, x in enumerate([int(x) for x in semantic_ver_string.split('.')[:3]):
       if i < len(cuda_version):
         cuda_version[i] = x
       else:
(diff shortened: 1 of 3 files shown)


python/cutlass_library/generator.py
```diff
@@ -103,7 +103,7 @@ def CudaToolkitVersionSatisfies(semantic_ver_string, major, minor, patch = 0):
 
   # Update cuda_version based on parsed string
   if semantic_ver_string != '':
-    for i, x in enumerate([int(x) for x in semantic_ver_string.split('.')]):
+    for i, x in enumerate([int(x) for x in semantic_ver_string.split('.')[:3]):
```
jackkosaian, line 106:
```suggestion
    for i, x in enumerate([int(x) for x in semantic_ver_string.split('.')[:3]]):
```

python/cutlass_library/sm90_utils.py
```diff
@@ -61,7 +61,7 @@ def CudaToolkitVersionSatisfies(semantic_ver_string, major, minor, patch = 0):
 
   # Update cuda_version based on parsed string
   if semantic_ver_string != '':
-    for i, x in enumerate([int(x) for x in semantic_ver_string.split('.')]):
+    for i, x in enumerate([int(x) for x in semantic_ver_string.split('.')[:3]):
```
jackkosaian, line 64:
```suggestion
    for i, x in enumerate([int(x) for x in semantic_ver_string.split('.')[:3]]):
```

jackkosaian (commented):
Thanks for making this change. I'll test it out internally.

hwu36 (approved):

Issue #1952, opened 2024-11-19
[BUG] Example 09_turing_tensorop_conv2dfprop does not work


IzanCatalan:
**Describe the bug**
When I compile and execute the example 09_turing_tensorop_conv2dfprop I get an error. My environment is Ubuntu 18.04, cuda 12, GPU Nvidia A100 ampere, make 3.25 and GCC version 8.4.

**Steps/Code to reproduce bug**
`izcagal@cmts10:~/cutlass/build/examples/09_turing_tensorop_conv2dfprop$ ./09_turing_tensorop_conv2dfprop --n=32 --h=224 --w=224 --c=128 --k=256 --r=1 --s=1
09_turing_tensorop_conv2dfprop: /mnt/beegfs/gap/izcagal/cutlass/include/cutlass/integer_subbyte.h:96: cutlass::integer_subbyte<Bits, Signed>::integer_subbyte(int) [with int Bits = 4; bool Signed = true]: Assertion `value < upper_bound' failed.
Aborted (core dumped)`

Any idea why this error happens or what it means?

KenRanmzes:
same error, have you fix it?

foreverlms:
I got the same error. Seems the assertion failed in `integer_subbyte`, but that's weird.

feifei-111:
same error

PR #1968, opened 2024-11-28
Fix upper_bound assert in integer_subbyte

panjd123:
The assertion with a signed upper bound was incorrect. I have corrected it and, to enhance readability, uniformly used closed intervals.

This may fix the problem in #1952.


hwu36:
Thank you for your pr. Your change is very similar to #1961 which was merged.

PR #2035, opened 2025-01-09
Use Compile-Time Constexpr If

leimao:
This runtime if can be removed at compile-time.


ccecka:
If `size<2>(tCrA)` is known at compile-time, this will already be elided by the compiler.

The only thing this does is *require* that it is known at compile-time, which may be attractive but is certainly overly restrictive.

d-k-b:
@leimao, is this fixing a failure you were seeing? If so, please elaborate. If not, the recommendation would be to allow the compiler to do its thing with the code as-is which is slightly more flexible.

PR #2066, opened 2025-01-28
fix cuda 12.6 issues

hwu36:
prior cuda 12.8 compiler is not supported by 3.8 initial upstream.  fix the issue here.


manishucsd:
I tried compiling and running a subset of targets in CUTLASS 3.8 with nvcc 12.4.99. I confirm that it works for me now. Thank you for fixing. We are still on nvcc 12.4.99, because of torch restrictions. 

If you have a note or steps in README.md on how to compile CUTLASS kernels independent of the toolkit used by torch that will be great. It is not the biggest concern at the moment, but I would like to compile CUTLASS kernels and use in production flow with any toolkit. Like most production flows, we also heavily depend on torch. So we have to keep these two very important things running well together. I can post more it when I get to it. Please point me to an existing issue on the same if you know one.

kerrmudgeon (approved):

Issue #2079, opened 2025-02-05
[BUG] `undefined symbol: cudaGetDriverEntryPointByVersion`


danthe3rd:
**Describe the bug**
Compiling CUTLASS with an NVCC version (12.6) newer than the driver version (12.4) causes this error:
```
undefined symbol: cudaGetDriverEntryPointByVersion, version libcudart.so.12
```

This is because of this compile-time check:
https://github.com/NVIDIA/cutlass/blob/affd1b693dfc121c51118cbc8583dfd308227ca6/include/cutlass/cuda_host_adapter.hpp#L107-L115

`cudaGetDriverEntryPointByVersion` has been added in the driver after version 12.5, but the check is on the NVCC version, which can be different from the driver version.
Can we use `cudaGetDriverEntryPoint` for all driver versions instead?

hwu36:
@danthe3rd , we can do that.  Do you want to file a PR?  We don't have your environment.

PR #2086, opened 2025-02-07
Always use cudaGetDriverEntryPoint with CUDA 12

danthe3rd:
`cudaGetDriverEntryPointByVersion` has been added to drivers in 12.5, but we don't know at compile time the driver version. In particular, we can build with nvcc 12.8 for a 12.2 driver for instance, and this was causing the following error:

```
undefined symbol: cudaGetDriverEntryPointByVersion,
```

Closes https://github.com/NVIDIA/cutlass/issues/2079

NOTE: I was not able to test it with CUDA driver 12.5+

### DIFF
--- a/include/cutlass/cuda_host_adapter.hpp
+++ b/include/cutlass/cuda_host_adapter.hpp
@@ -104,8 +104,7 @@ namespace cutlass {
 
 #else // defined(CUTLASS_ENABLE_DIRECT_CUDA_DRIVER_CALL)
 
-#if ((__CUDACC_VER_MAJOR__ >= 13) ||                               \
-    ((__CUDACC_VER_MAJOR__ == 12) && (__CUDACC_VER_MINOR__ >= 5))) \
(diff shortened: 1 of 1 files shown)


tridao:
Can confirm that we ran into this issue as well on older versions of CUDA, and @danthe3rd's fix works

zhyncs:
Here is a workaround https://github.com/sgl-project/sglang/pull/3372

manishucsd:
Did CUTLASS top-level CMake miss these lines?

```cmake
set(CUTLASS_ENABLE_DIRECT_CUDA_DRIVER_CALL OFF CACHE BOOL "Enable direct CUDA driver API calls .")
if (CUTLASS_ENABLE_DIRECT_CUDA_DRIVER_CALL)
  list(APPEND CUTLASS_CUDA_NVCC_FLAGS -DCUTLASS_ENABLE_DIRECT_CUDA_DRIVER_CALL=1)
endif()
```

hwu36 (approved):

PR #2110, opened 2025-02-13
Treat negative zero as equivalent to positive zero in sm90_sparse_gemm_compressor.hpp

tlrmchlsmth:
This PR handles negative zero in `sm90_sparse_gemm_compressor.hpp`, treating it as equivalent to positive zero. This is easy to solve by the caller before calling the compressor however it involves an extra pass of reading and writing the uncompressed sparse matrix, and it would be nice to generally remove this as a potential footgun.


hwu36:
@sklevtsov-nvidia , could you please review it?

include/cutlass/transform/kernel/sm90_sparse_gemm_compressor.hpp
```diff
@@ -401,7 +411,14 @@ class SM90StructuredSparseCompressor {
...
             ElementAMmaRawUnit elem_A = tAsA[elt_log_idx];
-            if ( elem_A != ElementAMmaRawUnit{0} ) {
+
+            // Handle negative 0
+            ElementAMmaRawUnit masked_elem_A = elem_A;
+            if constexpr (has_negative_zero_v<ElementA>) {
+              masked_elem_A = elem_A & sign_mask;
+            }
+
+            if ( masked_elem_A != ElementAMmaRawUnit{0} ) {
```
sklevtsov-nvidia, line 421: Minor nit: remove extra spaces in parens

include/cutlass/numeric_type_traits.h
```diff
@@ -0,0 +1,63 @@
...
+
+/////////////////////////////////////////////////////////////////////////////////////////////////
+
+namespace cutlass {
+
+/////////////////////////////////////////////////////////////////////////////////////////////////
+
+// Default case - no negative zero
+template <typename T>
+struct has_negative_zero : std::false_type {};
```
sklevtsov-nvidia, line 49: @hwu36 to decide if we want a new file for this trait or if it fits in `numeric_types.h` or elsewhere

sklevtsov-nvidia (approved):
LGTM, thank you for the contribution!

include/cutlass/numeric_type_traits.h
```diff
@@ -0,0 +1,63 @@
...
+
+// Float types that support negative zero
+// Note that this is false for float8_e4m3_t and float8_e5m2_t
```
alexsamardzic, line 52: Shouldn't it be true, otherwise how is 0x80 value interpreted for these data types?  According to the [paper](https://arxiv.org/pdf/2209.05433) it's -0.0, see table 1.

sklevtsov-nvidia: @alexsamardzic you are correct, we need to add this for all of fp8/fp6/fp4. @tlrmchlsmth feel free to do this if you can, otherwise we can fix this in https://github.com/NVIDIA/cutlass/pull/2122

alexsamardzic: Here is a patch to apply on top of this PR, with FP8 and MX data types added (please check that all of them are there, and make sense here), and also minor changes from my #2122 that may be worth adding.  For `test/unit/transform/device/sm90_sparse_gemm_compressor_legacy.hpp`, I don't think it would work with sub-byte data types, but at the moment I don't have access to hardware to test - so I've put there simpler check for negative zero, but also a static assert that would fail if a sub-byte data type used.  I hope we can put the patch into this PR, and I'm closing mine now.

[patch.txt](https://github.com/user-attachments/files/18914127/patch.txt)

tlrmchlsmth: ah, you're right - I actually got confused between `float8_e4m3fnuz` and `float8_e4m3fn` on this one. I'll apply the patch, thanks!

tlrmchlsmth: @alexsamardzic thanks for the patch. I've applied it and it looks good to me. PTAL!

tlrmchlsmth:
Any blockers on this @alexsamardzic, @sklevtsov-nvidia?

alexsamardzic (approved):

alexsamardzic:
None, I think it could be merged - just approved.

alexsamardzic:
@hwu36 Could someone please confirm and merge this PR?

hwu36:
We will merge this soon, but we need to change it a bit.  It breaks something.

hwu36 (approved):

PR #2111, opened 2025-02-13
cleanup code in tile iterator example

mayank31398:
as title suggests

### DIFF
--- a/examples/04_tile_iterator/tile_iterator.cu
+++ b/examples/04_tile_iterator/tile_iterator.cu
@@ -95,7 +95,6 @@ __global__ void copy(
     typename Iterator::Fragment fragment;
 
     for(; iterations > 0; --iterations) {
-
       src_iterator.load(fragment);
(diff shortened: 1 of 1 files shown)


examples/04_tile_iterator/tile_iterator.cu
```diff
@@ -94,19 +94,7 @@ __global__ void copy(
...
-    src_iterator.load(fragment);
-    dst_iterator.store(fragment);
-
-
-    ++src_iterator;
-    ++dst_iterator;
-
-    for(; iterations > 1; --iterations) {
-
+    for(; iterations > 0; --iterations) {
```
hwu36, line 97: we still need to keep the initialization part of code.  we want `fragment` is in a known state before the load and store operation.

hwu36:
@kerrmudgeon

mayank31398:
is this relevant or should I close?

PR #2033, opened 2025-01-08
[EVT] Add support for Row/Col broadcast PtrArray

jwfromm:
To enable FP8 grouped gemm with rowwise scaling in the epilogue, we need to be able to provide a list of pointers to the scales for each group. This PR extends Sm90Row/ColBroadcast to support PTRArray to handle this case. Now if the ElementInput type is specified as a pointer, the corresponding input is an array of pointers, enabling grouped gemm. For example in our case we define EVT nodes like this:
```
  using XScale = cutlass::epilogue::fusion::Sm90ColBroadcast<
      0,
      TileShape,
      ElementComputeEpilogue*, // Indicate input is array of pointers.
      ElementComputeEpilogue,
      cute::Stride<cute::Int<1>, cute::Int<0>, cute::Int<0>>>;

  using WScale = cutlass::epilogue::fusion::Sm90RowBroadcast<
      0,
      TileShape,
      ElementComputeEpilogue*, // Indicate input is array of pointers.
      ElementComputeEpilogue,
      cute::Stride<cute::Int<0>, cute::Int<1>, cute::Int<0>>>;
```

PR #2120, opened 2025-02-20
[EVT] Fix Row/Col broadcast with array arguments

jwfromm:
#2033 added the ability for row/col broadcast nodes to accept ptr array arguments, however, during some followup refactoring it was broken. Specifically, there is an invalid type comparison that was added. This small PR fixes the issue by using a constexpr if to avoid the comparison if ptrarray is being used.

### DIFF
--- a/include/cutlass/epilogue/fusion/sm90_visitor_load_tma_warpspecialized.hpp
+++ b/include/cutlass/epilogue/fusion/sm90_visitor_load_tma_warpspecialized.hpp
@@ -1042,8 +1042,10 @@ struct Sm90RowBroadcast {
       is_zero_ = params.null_default == ElementCompute(0);
     }
     // Dynamic non-batched scalar broadcast
-    else if (IsDynamicBroadcast && stride_N == bool(0) && stride_L == repeat_like(stride_L, 0) && !IsArrayOfPointers) {
-      is_zero_ = params.ptr_row[0] == ElementInput(0);
+    if constexpr (!IsArrayOfPointers) {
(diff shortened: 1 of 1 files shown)


jwfromm:
@ANIKET-SHIVAM can you take a look at this small change?

ANIKET-SHIVAM:
@jwfromm Is the invalid type comparison happening bcoz this line right now is not behind `constexpr`:
`is_zero_ = params.ptr_row[0] == ElementInput(0);`
For Grouped Gemm case, there is a type mismatch, where `ptr_row[0]`  is of `ElementInput*` type instead and hence the invalid type conversion issue. Is that correct?
Could you try moving `constexpr` check inside instead like this and see if that works fine for you:
```
    else if (IsDynamicBroadcast && stride_N == bool(0) && stride_L == repeat_like(stride_L, 0)) {
      if constexpr (not IsArrayOfPointers) {
        is_zero_ = params.ptr_row[0] == ElementInput(0);
      }
```

jwfromm:
@ANIKET-SHIVAM Yeah that would also work, as long as we dont compare ptr_row[0] to ElementInput(0) as they are different types it will be fine. If you prefer having the constexpr check inside the `else if` we can do that instead.

ANIKET-SHIVAM:
yes, please. thanks.

jwfromm:
@ANIKET-SHIVAM Done, I also confirmed this still works as expected.

ANIKET-SHIVAM (approved):
@hwu36 this is ready to be merged.

hwu36 (approved):

PR #2122, opened 2025-02-20
Recognize -0.0 as zero in the sparse compressor

alexsamardzic:


alexsamardzic:
The sparse compressor is not treating -0.0 values as zeros.  To reproduce the problem, apply this PR, but then revert change in `include/cutlass/transform/kernel/sm90_sparse_gemm_compressor.hpp` - namely, the rest of the changes are to fix other issues (as follows: the compressor device tests were not built at all, then there was an include file missing, and finally there is a check in the tests for number of non-zero elements per each sparsity block that would fail first, while the point of the reproducer is to show wrong results).  Then, apply following patch, that would replace all zero values with -0.0 values in the tensor to be compressed:
```
diff --git a/test/unit/transform/device/testbed_sparse_gemm_compressor.hpp b/test/unit/transform/device/testbed_sparse_gemm_compressor.hpp
index f4445824..d9c65209 100644
--- a/test/unit/transform/device/testbed_sparse_gemm_compressor.hpp
+++ b/test/unit/transform/device/testbed_sparse_gemm_compressor.hpp
@@ -300,6 +300,20 @@ public:
 
     compressor_utility.structure_sparse_zero_mask_fill(datas.tensor_A.host_data(), seed + 6);
 
+    // Replace all zeros with -0.0.
+    ElementA zero(0);
+    ElementA minus_zero(0);
+    uint8_t* minus_zero_ptr = (uint8_t*)&minus_zero;
+    *minus_zero_ptr = 0x80;
+    for (int m = 0; m < GemmM * GemmL; ++m) {
+      for (int k = 0; k < GemmK; ++k) {
+        cutlass::Coord<2> coord = cutlass::make_Coord(m, k);
+        if (datas.tensor_A.host_view().at(coord) == zero) {
+          datas.tensor_A.host_view().at(coord) = minus_zero;
+        }
+      }
+    }
+
     // Check for failed devide
     CUDA_CHECK_FALSE(cudaGetLastError());
```
and build and run `cutlass_test_unit_sm90_structured_sparse_gemm_compressor_f8` - the test will show mismatched results, as some -0.0 values will be picked into the compressed tensor, instead of values that are actually non-zero.  (By changing `minus_zero_ptr` data type to `uint16_t` and `uint32_t`, and value written into the location pointed by this pointer to `0x8000` and `0x80000000`, respectively, the problem could be reproduced with `cutlass_test_unit_sm90_structured_sparse_gemm_compressor_f16` and `cutlass_test_unit_sm90_structured_sparse_gemm_compressor_f32` tests too).

alexsamardzic:
Huh, I see now there is already [PR 2110](https://github.com/NVIDIA/cutlass/pull/2110) with the fix for the compressor...  Maybe we could combine the PRs, or I can change my PR to keep only the tests-related fixes?

@sklevtsov-nvidia

alexsamardzic:
Closing as #2110 does it right.

PR #2129, opened 2025-02-24
Fix typo in layout algebra md doc

botbw:
According to the definition of product `(A, A* o B)`, I think the doc here describes the second mode `A* o B` instead of `A*`.

### DIFF
--- a/media/docs/cute/02_layout_algebra.md
+++ b/media/docs/cute/02_layout_algebra.md
@@ -491,7 +491,7 @@ This is clearly just a copy of `A`.
 
 > the second mode is the layout `B` but with each element replaced by a "unique replication" of layout `A`.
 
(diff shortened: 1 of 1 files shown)


ccecka:
This sentence is providing the motivation behind using `A*` in `A* o B`. This change appears to make the discussion incorrect.

botbw:
> This sentence is providing the motivation behind using `A*` in `A* o B`. This change appears to make the discussion incorrect.

@ccecka Thanks for pointing out! This still seems a bit confusing to me: If A* represents the repetition of A attempting to complete cosize(B)*size(A), wouldn't A* o B represent those repetition that are available for B? ****

ccecka:
No, `A*` is the repetitions of `A` and `A* o B` is the repetitions of `A` in the shape and order of `B`.

botbw:
> No, `A*` is the repetitions of `A` and `A* o B` is the repetitions of `A` in the shape and order of `B`.

@ccecka I see, thanks for clarification.

PR #2130, opened 2025-02-24
Flash MLA support

myu-guo:
Port the Deepseek's Flash MLA kernel to CUTLASS example. the python code will be translated to c++ code in this PR. 

The kernel can be executed successfully. but It lacks of some reference check and robust tests.

TODO:

- [ ] Add Input&Ouput processing (like reshape() in the python)
- [ ] Add reference check

@hwu36 for viz.


examples/68_hopper_flash_mla/softmax.h
```diff
@@ -0,0 +1,197 @@
+// Adapted from https://github.com/deepseek-ai/FlashMLA/blob/main/csrc/softmax.h
```
tridao, line 1: // Adapted from https://github.com/Dao-AILab/flash-attention/blob/main/csrc/flash_attn/src/softmax.h
as in the DS's comments
Same with other files

myu-guo: Got it, thanks!

PR #2161, opened 2025-03-10
Blockwise Improvement and Programmatic Dependent Launch

depaulmillz:
- Fixes potential OOB access by shrinking copy size for SFA and SFB in blockwise/groupwise kernels
- Programmatic dependent launch on Blackwell
  - Performance testing is enabled in the profiler


thakkarV:
LGTM. @kerrmudgeon please merge.

PR #2167, opened 2025-03-13
Fix sm100 gemm wrong static constexpr that breaks compilation on Windows

SystemPanic:
`constexpr` was incorrectly added to the function definition only for this architecture, breaking the compilation of Cutlass on Windows.

### DIFF
--- a/include/cutlass/platform/platform.h
+++ b/include/cutlass/platform/platform.h
@@ -523,7 +523,7 @@ using std::is_trivially_copyable;
 
 #endif
 
-#if defined(_MSC_VER) || (201703L <=__cplusplus)
+#if (201703L <=__cplusplus)
 
 /// std::is_unsigned_v
 using CUTLASS_STL_NAMESPACE::is_integral_v;
--- a/include/cutlass/platform/platform.h
+++ b/include/cutlass/platform/platform.h
@@ -523,7 +523,7 @@ using std::is_trivially_copyable;
 
 #endif
 
-#if (201703L <=__cplusplus)
+#if defined(_MSC_VER) || (201703L <=__cplusplus)
 
 /// std::is_unsigned_v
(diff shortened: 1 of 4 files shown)


include/cutlass/platform/platform.h
```diff
@@ -523,7 +523,7 @@ using std::is_trivially_copyable;
 
 #endif
 
-#if (201703L <=__cplusplus)
+#if defined(_MSC_VER) || (201703L <=__cplusplus)
```
d-k-b, line 526: This looks like a symptom of another issue. What error did you see here?

d-k-b: The CUTLASS CMake code applies the `__cplusplus` flag to MSVC compilation automatically. Is the error being seen with a build outside of CMake? I'd prefer we revert this change for this PR and then if it's needed for some scenario we handle it in a separate request.

https://github.com/NVIDIA/cutlass/blob/9d165a3b8ef446a7ff3db198413f82bcb83f46fe/CMakeLists.txt#L561

```suggestion
#if (201703L <= __cplusplus)
```

d-k-b (approved):

include/cutlass/platform/platform.h
SystemPanic: Ok, I see where the problem is.

vLLM it's being built with CMake, but it's not adding CUTLASS in the standard way, so the MSVC specific flags [https://github.com/NVIDIA/cutlass/blob/9d165a3b8ef446a7ff3db198413f82bcb83f46fe/CMakeLists.txt#L561](https://github.com/NVIDIA/cutlass/blob/9d165a3b8ef446a7ff3db198413f82bcb83f46fe/CMakeLists.txt#L561) are not being applied at all.

Commit reverted.

d-k-b (approved):

hwu36 (approved):

PR #2195, opened 2025-03-25
suppress compilation warnings

reed-lau:
When updating to the latest version, the cute program compiled with numerous warnings, so I helped fix it.


thakkarV:
@Junkai-Wu @hwu36

hwu36 (approved):

PR #2219, opened 2025-04-04
[SM90] Change register allocation for TileN=208 to avoid spills

tridao:
With the usual register allocation (producer 40, consumer 232) compiling Gemm with tile shape 256 x 208 (cooperative) or 128 x 208 (pingpong) show lots of register spilling (e.g. ~3000 bytes spill). For this case we can change the register allocation to producer 24, consumer 240, which avoids spills.
Cc @thakkarV @hwu36


thakkarV:
This is awesome! Thank you :D

thakkarV:
@hwu36 @Junkai-Wu

thakkarV:
Were you able to AB test perf results by any chance?

tridao:
Perfwise nothing changes for the usual tile shapes since we keep reg allocation the same for those. For tile shape 256 x 208 it's obv a lot faster without the massive spill (5-10x). In some niche cases where 256 x 208 fits just right and there's minimal wave quantization 256 x 208 can be a good choice.

hwu36 (approved):

Issue #2133, opened 2025-02-25
[BUG] EpilogueTileAuto doesn't work when tile shape is (128, 112, 64)


henrylhtsang:
**Describe the bug**
I am using gemm configs from generator.py. op.configuration name is cutlass3x_sm90_tensorop_s64x56x16gemm_f16_f16_f32_void_f16_128x112x64_4x1x1_0_ttn_align8_warpspecialized_pingpong_epi_tma

This is the code https://gist.github.com/henrylhtsang/aa797c7d280b792eef574317111ac215

Question:
1. Should I consider not use EpilogueTileAuto? Or is this fixable from this logic? https://github.com/NVIDIA/cutlass/blob/eefa171318b79cbe2e78514d4cce5cd0fe919d0c/include/cutlass/epilogue/collective/builders/sm90_builder.inl#L115-L121
2. What epilogue tile shape should I use?

**Steps/Code to reproduce bug**
compile https://gist.github.com/henrylhtsang/aa797c7d280b792eef574317111ac215

```
     static_assert(size<1>(CtaTileMNK{}) % size<1>(shape(EpilogueTile{})) == 0, "EPI_TILE_N must divide CTA_N");
```

**Expected behavior**
compilation should be successful

**Environment details (please complete the following information):**
 - Environment location: [Bare-metal, Docker, Cloud(specify cloud provider)] 

**Additional context**

tridao:
You'd want
```
auto tile_n = cute::gcd(cute::min(_32{}, size<1>(TileShape_MNK{})), size<1>(TileShape_MNK{}));
```

thakkarV:
Keeping live.

henrylhtsang:
closed with https://github.com/NVIDIA/cutlass/pull/2220

PR #2220, opened 2025-04-04
Set EpiTile correctly when TileN is not divisible by 32

tridao:
If TileN is not divisible by 32 (e.g, 208), by default EpiTile would be set to 128 x 32, which does not compile as TileN is required to be divisible by EpiTileN. We instead set EpiTileN to be gcd(32, TileN).
Cc @thakkarV @hwu36


richardmcai:
Looks fine to me. What layouts/types/tile sizes was this tested on?

tridao:
I tested on tile shape {128, 256} x {128, 144, 160, 176, 192, 208}.
Yup I can pass the tiler directly, but it might be nicer for `EpilogueTileAuto` to work with these tile shapes too. E.g. this would fix #2133

hwu36 (approved):

PR #2221, opened 2025-04-04
MLA example page cache fix

divchenko:
Current implementation assumes that number of pages in the block table is same as in the kv cache. This is typically not true: page table typically references a subset of pages in the kv cache.
Introduced pages_per_seq vs total_pages parameters. The former one is max pages in the page table while the latter one is total pages in the kv cache. Updated code to generate random page numbers from the cache.

### DIFF
--- a/examples/77_blackwell_fmha/77_blackwell_mla.cu
+++ b/examples/77_blackwell_fmha/77_blackwell_mla.cu
@@ -76,6 +76,7 @@ struct Options {
   bool is_var_split_kv = false;
   int max_split_kv = 16;
   int page = -1;
+  int total_pages = 10000;
   float spread = 0.2f;
   int iterations = 3;
   bool verify = false;
@@ -137,6 +138,7 @@ struct Options {
 
     cmd.get_cmd_line_argument("split_kv", split_kv, defaults.split_kv);
     cmd.get_cmd_line_argument("page", page, defaults.page);
+    cmd.get_cmd_line_argument("total_pages", total_pages, defaults.total_pages);
     cmd.get_cmd_line_argument("spread", spread, defaults.spread);
     cmd.get_cmd_line_argument("is_var_split_kv", is_var_split_kv, false);
     if (page == -1) {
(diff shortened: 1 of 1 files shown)


thakkarV:
Thanks a lot for the contribution!

v0i0:
Hey @divchenko,

Thank you for having a look!

I think you can just set the stride_page_table to account for pages_per_batch being smaller than total_pages, or did you see a failure with that?

In any case, this seems to at the very least make the code more clear, so I am happy to take the change.

divchenko:
> Hey @divchenko,
> 
> Thank you for having a look!
> 
> I think you can just set the stride_page_table to account for pages_per_batch being smaller than total_pages, or did you see a failure with that?
> 
> In any case, this seems to at the very least make the code more clear, so I am happy to take the change.

@v0i0 the issue is in tma impl:
This one needs boundary for page table (paged_BT):
```
    auto mPT_l = make_tensor(make_gmem_ptr(mainloop_args.ptr_page_table), make_shape(paged_BT, B), mainloop_args.stride_page_table);
```
vs this one needs boundary for kv cache (paged_BC):
```
    auto mCL = mainloop_params.tma_load_c_latent.get_tma_tensor(make_shape(paged_K, D_latent, paged_BC));
```

v0i0:
`mPT_l` is not loaded via TMA. it is either a direct load, or a cp.async.

What problem size did you see an issue with? Did it crash or did you see incorrect results?

divchenko:
> `mPT_l` is not loaded via TMA. it is either a direct load, or a cp.async.
> 
> What problem size did you see an issue with? Did it crash or did you see incorrect results?

I mean load_tma function.

Oh, I see what's going on. I was confused what MainloopArguments.page_count means. It actually means total page count in the kv cache (vs max num pages per sequence). I was setting it to the latter and it was giving me wrong results only for TMA. Hence I concluded that TMA impl has a bug.
I reverted the changes to the kernel, but kept the changes to the example invocation code. I think it's much clearer this way (as it's a more typical usage).

mnicely:
@divchenko did you get a resolution here?

Issue #2243, opened 2025-04-16
[BUG] cutlass/python/cutlass has extra dependencies that aren't needed for EVT


mlazos:
**Describe the bug**
When attempting to use the epilogue visitor tree python tracer at cutlass/python/backend/evt, there are a lot of unnecessary dependencies at the top level when importing cutlass/python/cutlass. All of these dependencies could be lazily imported or moved into the functions where they are used, this would make integrating EVT into pytorch as-is feasible. 

**Steps/Code to reproduce bug**
cuda, scipy, pydot are the packages that are imported eagerly. Simply import cutlass/python/cutlass to repro. 

**Expected behavior**
I think these dependencies should be lazily imported, since they are only used in very specific scenarios in cutlass/python/cutlass. 

**Environment details (please complete the following information):**
N/A

**Additional context**

hwu36:
we are going to merge your PRs soon with a little modification.  after merging these, do you still have problems of dependencies?

cc += @jackkosaian

PR #2248, opened 2025-04-18
Import pydot lazily

mlazos:
Towards https://github.com/NVIDIA/cutlass/issues/2243

This PR lazily imports pydot, note that the annotations import is to only import for type annotations during type checking, not on import. 

cc @Henrylhtsang, @eellison, @thakkarV


mlazos:
cc @mnicely

hwu36 (approved):

PR #2250, opened 2025-04-18
Import scipy lazily

mlazos:
Scipy was only used in one piece of code, I only import it if used.


mlazos:
cc @thakkarV, @eellison, @henrylhtsang

mlazos:
cc @mnicely

hwu36 (approved):

PR #2256, opened 2025-04-21
Use cudaMemcpyAsync in gemm grouped with kRequiresPrecomputation sche…

HydraQYH:
# cudaMemcpy may hurt performace in gemm grouped
We use cutlass 2.x api to launch a grouped gemm in some gpus which cc < 90. We find that if we use kHostPrecompute as GroupScheduleMode, cutlass will compute some schedule meta info on host, and then copy it to GPU by a cudaMemcpy api with default stream and sync semantic: https://github.com/NVIDIA/cutlass/blob/81a43e6d92cdd8c20d22392f9579604ed5f710a1/include/cutlass/gemm/device/base_grouped.h#L131

This may hurt performance. So we use cudaMemcpyAsync in grouped gemm.

### DIFF
--- a/include/cutlass/gemm/device/base_grouped.h
+++ b/include/cutlass/gemm/device/base_grouped.h
@@ -127,8 +127,8 @@ private:
   }
 
   /// Copy from `data` to `workspace`
-  Status copy_to_workspace(void* workspace, void* data, size_t bytes) {
-    cudaError_t cuda_error = cudaMemcpy(workspace, data, bytes, cudaMemcpyHostToDevice);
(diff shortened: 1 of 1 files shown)


hwu36:
@jackkosaian , could u please take a look?

jackkosaian:
Should be fine as long as someone passes in the same stream to each. I think that's needed for our general device-level APIs that take in a stream parameter as well.

HydraQYH:
> Should be fine as long as someone passes in the same stream to each. I think that's needed for our general device-level APIs that take in a stream parameter as well.

@jackkosaian @hwu36 So, can we approval this pr?

hwu36 (approved):

PR #2257, opened 2025-04-22
Limit the number of SMs (sm_count) to user-provided value during profiling.

manishucsd:
Provision to limit the number of SMs during CUTLASS profiling runs.


manishucsd:
cc: @hwu36 , @itramble @milesvant

hwu36:
do you see different kernel get picked when changing sm count?

manishucsd:
yes.

Issue #2268, opened 2025-04-29
[BUG] EpilogueTileAuto doesn't work when tile shape N={48, 80, 96, 112, ...} on SM100


Algy:
For non-blockscaled blackwell gemm, theoretically all multiple number of 16 as ctaN = {16, 32, 48, 64, 80, ...} should be supported as the UMMA specification.

However, I get compilation error with those CtaNs other than 16, 32, 64, 96, 128, 192, 256.

For example, when I set the tile shape of the example `70_blackwell_fp16_gemm.cu` to `<_64, _48, _64>`, I encounter some long cryptic error that appears to be related to shape divisibility.

PR #2220 addresses this kind of issue with Hopper kernels, so I will come up with the similar approach to this issue soon.

Algy:
I have come up with the PR #2269 so it can handle multiple of 16 but still have problem with handling mulitple of 8 such as 24, 56, 72. Any idea?

PR #2269, opened 2025-04-29
Support N={48, 80, 96, 112, ...} for SM100 EpilogueTileAuto

Algy:
This partially addresses #2268

I'm still figuring out how I can support ctaNs that are multiples of 8 but not multiple of 16, e.g. N={8, 24, 56, 72, ...}.


Algy:
ok. I'll add unit tests then.

PR #2295, opened 2025-05-13
[Docs] Add clarifying note on PermutationMNK 

imaginary-person:
Added a clarifying note that for PermutationMNK, each mode's shape & strides are listed in the order of `the outermost dim to the innermost dimension`.

`(4, 4, 2):(1, 8, 4)` can perhaps be more easily understood by folks familiar with numpy semantics if they were aware of the order of shape & strides (since it's opposite to the numpy convention).

To elaborate, the visual example in the doc on scatter permutation on M-mode can also be visualized as:

```python
import torch
before = torch.arange(32).view(2, 4, 4).contiguous()
after = a.as_strided(size=(2, 4, 4), stride=(4, 8, 1))
# view indices before permutation 
print(before.flatten())
# view indices after permutation 
# If the value at ith index is n, that means after[i] = before[n]
print(after.flatten())
```

### DIFF
--- a/media/docs/cpp/cute/0t_mma_atom.md
+++ b/media/docs/cpp/cute/0t_mma_atom.md
@@ -504,6 +504,9 @@ That layout `(4,4,2):(1,8,4)` is read like a scatter permutation, telling the m-
 old m-coord:  0  1  2  3  4  5  6  7  8  9 10 11 12 13 14 15 16 17 18 19 20 21 22 23 24 25 26 27 28 29 30 31
 new m-coord:  0  1  2  3  8  9 10 11 16 17 18 19 24 25 26 27  4  5  6  7 12 13 14 15 20 21 22 23 28 29 30 31
 ```
+
+Note that the shape & strides of M are listed in the order of the innermost dimension to the outermost dimension.
+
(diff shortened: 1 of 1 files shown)


imaginary-person:
Hi @thakkarV, can you please help review this change, as per your convenience?
 It'd help me confirm if I understand PermutationMNK correctly. 
Thanks in advance!

ccecka:
I don't see this adding value to the current explanation with the pre- and post- permutation coordinates listed explicitly. Does "innermost" mean fastest changing and on-the-left? That doesn't appear to be your meaning. Point is, you would also need to define these terms.

The description of `Layout`, the colexicographical ordering of coordinates, and the intuition of reading `Layout`s from left to right is given in other sections.

imaginary-person:
Hi @ccecka, thanks a lot for the info!

By innermost dimension, I meant the fastest changing dimension, which is the leftmost in case of cute, but the rightmost in numpy semantics, which may cause some confusion to folks new to cutlass internals.

> The description of Layout, the colexicographical ordering of coordinates, and the intuition of reading Layouts from left to right is given in other sections.

 Thanks, I'll go through the rest of the documentation.
It looks like this change would be redundant.

PR #2315, opened 2025-05-19
[docs] Specify `nvidia-cutlass-dsl` version

justinfargnoli:
By default `pip install nvidia-cutlass-dsl` downloads version `0.0.0a0` instead of `4.0.0.dev1`. However, to the best of my knowledge, version `0.0.0a0` doesn't contain any functional code.

Thus, I'd like to suggest updating the documentation to point to the version of `nvidia-cutlass-dsl` that can be used to reproduce examples in the documentation.

### DIFF
--- a/media/docs/pythonDSL/quick_start.rst
+++ b/media/docs/pythonDSL/quick_start.rst
@@ -12,7 +12,7 @@ To install the CUTLASS DSL, run:
 
 .. code-block:: bash
 
-   pip install nvidia-cutlass-dsl
+   pip install nvidia-cutlass-dsl==4.0.0.dev1
 
 The ``nvidia-cutlass-dsl`` wheel includes everything needed to generate GPU kernels. It requires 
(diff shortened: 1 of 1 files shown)


hwu36:
@IonThruster

IonThruster:
Thanks for reporting - we are working on removing that initial version to avoid confusions.
 
Are you sure you are using Python 3.12 ? 

If so it should pick up the latest (correct) version of the wheel - or alternatively you could force it via 
`pip install -r python/CuTeDSL/requirements.txt`.

IonThruster:
> I tried again on a different machine (macOS)

That `probably` explain why - the CuTe DSL only supports x86 + Python 3.12 at the moment, and I believe it points to the default v0.0 if it doesn't match exactly (which is why plan to remove it from pypi as well - in which case it would return a no-matching version error).

Considering the valid use case you mentioned i.e out of tree usage - I feel it's better to close this PR.

PR #2318, opened 2025-05-20
More generic interface for Group GEMM problem size

nandor:
The problem size array of group gemm is no longer exposed on sm90/sm100 tile schedulers, allowing custom problem sizes to be plugged in, as long as they respect a generic interface.

This is done in order to be able to better integrate CUTLASS into MoE GEMM problems.


nandor:
For reference, this is the problem shape we can define for MoE:

```
struct MoEProblemShape {
  using UnderlyingProblemShape = Shape<int, int, int>;

  int32_t num_groups = 1;
  const int32_t *num_routed;
  int32_t n;
  int32_t k;

  CUTLASS_HOST_DEVICE
  int32_t groups() const { return num_groups; }

  CUTLASS_HOST_DEVICE
  UnderlyingProblemShape const
  get_problem_shape(int32_t group_idx) const {
    return UnderlyingProblemShape{num_routed[group_idx], n, k};
  }

  CUTLASS_HOST_DEVICE
  UnderlyingProblemShape const
  get_host_problem_shape(int32_t group_idx) const {
    return UnderlyingProblemShape{};
  }

  CUTLASS_HOST_DEVICE
  bool
  is_host_problem_shape_available() const {
    return false;
  }
};
```

ANIKET-SHIVAM:
These changes are now in main branch. Closing the PR.

PR #2324, opened 2025-05-22
Ensure compatibility with "-Wimplicit-fallthrough" when compiling with Clang

wenxin0319:
This commit makes its files safe for use with "-Wimplicit-fallthrough".

### DIFF
--- a/include/cute/arch/copy_sm90_desc.hpp
+++ b/include/cute/arch/copy_sm90_desc.hpp
@@ -239,7 +239,7 @@ to_CUtensorMapDataType() {
 inline CUtensorMapSwizzle
 to_CUtensorMapSwizzle(SmemSwizzleBits const& t, SmemSwizzleBase const& b) {
   switch (t) {
-    default: assert(false && "Unsupported pair of SmemSwizzleBits and SmemSwizzleBase!");
+    default: throw std::runtime_error("Unsupported pair of SmemSwizzleBits and SmemSwizzleBase!");
     case SmemSwizzleBits::DISABLE: 
       assert((b == SmemSwizzleBase::SWIZZLE_BASE_16B) && "Expected 16B swizzle base for 0B swizzle bits.");
       return CU_TENSOR_MAP_SWIZZLE_NONE;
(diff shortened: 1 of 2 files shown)


ccecka:
Good point, though we should prefer a solution that doesn't use exceptions.

r-barnes:
@ccecka - What solution would suggest?

Elsewhere in the project I see `std::cerr` followed by `abort()`:
https://github.com/NVIDIA/cutlass/blob/b995f933179c22d3fe0d871c3a53d11e4681950f/tools/util/include/cutlass/util/device_rmsnorm.h#L181-L182

and `std::cerr` followed by `std::exit`:
https://github.com/NVIDIA/cutlass/blob/b995f933179c22d3fe0d871c3a53d11e4681950f/examples/41_fused_multi_head_attention/fused_multi_head_attention_backward.cu#L89-L90

and a number of uses of `throw` ([link](https://github.com/search?q=repo%3ANVIDIA%2Fcutlass+throw&type=code)):
https://github.com/NVIDIA/cutlass/blob/b995f933179c22d3fe0d871c3a53d11e4681950f/tools/library/src/handle.cu#L62

https://github.com/NVIDIA/cutlass/blob/b995f933179c22d3fe0d871c3a53d11e4681950f/tools/util/include/cutlass/util/device_memory.h#L69

As contributors @wenxin0319 and I would like to match the existing project style, and it seems to us that `throw std::runtime_error` is pretty reasonable given its existing use within the project.

If that's not the case, what should we do instead?

thakkarV:
keeping live. will get back to you

ccecka:
We would prefer `static_assert` and `assert` when possible, but since these are host-only I suppose it's fine.

Approving now.

thakkarV:
@Junkai-Wu CC

Junkai-Wu:

hwu36 (approved):

r-barnes:
I've tried to enable `-Wimplicit-fallthrough` in #2661, but I'm not sure what I'm doing.

PR #2383, opened 2025-06-08
[CuteDSL] Wrap nvvm.barrier_arrive

tridao:
Wrap nvvm.barrier_arrive. This function is useful in FA3 for inter-warpgroup overlap.

### DIFF
--- a/python/CuTeDSL/cutlass/cute/arch/nvvm_wrappers.py
+++ b/python/CuTeDSL/cutlass/cute/arch/nvvm_wrappers.py
@@ -225,6 +225,25 @@ def barrier(*, barrier_id=None, number_of_threads=None, loc=None, ip=None) -> No
         barrier_id=barrier_id, number_of_threads=number_of_threads, loc=loc, ip=ip
     )
 
+
+@dsl_user_op
+def barrier_arrive(
+    barrier_id: int | Int32,
+    number_of_threads: int | Int32,
+    *,
+    loc=None,
+    ip=None
+) -> None:
+    """
+    Arrive at a named barrier.
+    """
+    barrier_id = Int32(barrier_id).ir_value(loc=loc, ip=ip)
+    number_of_threads = Int32(number_of_threads).ir_value(loc=loc, ip=ip)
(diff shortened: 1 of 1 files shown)


python/CuTeDSL/cutlass/cute/arch/nvvm_wrappers.py
```diff
@@ -225,6 +225,25 @@ def barrier(*, barrier_id=None, number_of_threads=None, loc=None, ip=None) -> No
         barrier_id=barrier_id, number_of_threads=number_of_threads, loc=loc, ip=ip
     )
 
+
+@dsl_user_op
+def barrier_arrive(
```
thakkarV, line 230: Yep, totally worth adding. we may have to change this to match CUTLASS C++ and make this `named_barrier_arrive` instead.

Perhaps it is also a good time to start adding classes here 

```python
class NamedBarrier():
    barrier_id = #number set during __init__
    num_threads = # number set during __init__

    def barrier_arrive():
        pass # this method here.

    barrier_wait():
      pass

    barrier_arrive_and_wait():
      pass
```

that would make using this in actual kernels super easy since you can instantiate a barrier object and then just arrive and wait on it like a pipeline without having to match the barrier ID and thread count in every separate control flow path.

thakkarV: CC @IonThruster

thakkarV: NOTE: Not saying this class should be in the NVVM wrappers file, this addition is good as it is. but adding a nambed barrier object to our own `barriers.py` file and adding a separate class like this for `MBarrier` would be good too.

IonThruster: ++ @albertxu1994

thakkarV:
shipping in 4.1

thakkarV:
Closing as done, thanks a lot for the idea and contribution

PR #2402, opened 2025-06-14
Make unittest compilation faster

andralex:
FILE_TO_C_STRING creates gigantic files consisting mainly of sequences char(0xNN). Just including such a file takes a long time (e.g. 27 seconds for build/test/unit/nvrts/nvrtc/cute/arch/mma_sm90_gmma_sparse_ext.hpp).

This PR generates files as raw delimited C++ strings. This reduces the cost of compilation by roughly 10x. There are over 500 such files.

### DIFF
--- a/bin2hex.cmake
+++ b/bin2hex.cmake
@@ -28,17 +28,26 @@
 
 # A small utility function which generates a C-header from an input file
 function(FILE_TO_C_STRING FILENAME VARIABLE_NAME OUTPUT_STRING ZERO_TERMINATED)
-  FILE(READ "${FILENAME}" HEX_INPUT HEX)
   if (${ZERO_TERMINATED})
-    string(APPEND HEX_INPUT "00")
+    # Read file as plain text and create output as a raw string literal
+    FILE(READ "${FILENAME}" RAW_CONTENT)
+
+    set(RAW_DELIM_END "abc1ae6e1e82e3de")
+    set(RESULT "static char const ${VARIABLE_NAME}[] = R\"${RAW_DELIM_END}(\n${RAW_CONTENT})${RAW_DELIM_END}\";\n")
+
+    set(${OUTPUT_STRING} "${RESULT}" PARENT_SCOPE)
+  else()
+    # Read file as hex and create output as a C-style brace-delimited array
+    # N.B.: this is much slower to compile than the raw string literal above
+    FILE(READ "${FILENAME}" HEX_INPUT HEX)
+
+    string(REGEX REPLACE "(....)" "\\1\n" HEX_OUTPUT ${HEX_INPUT})
+    string(REGEX REPLACE "([0-9a-f][0-9a-f])" "char(0x\\1)," HEX_OUTPUT ${HEX_OUTPUT})
+
+    set(HEX_OUTPUT "static char const ${VARIABLE_NAME}[] = {\n  ${HEX_OUTPUT}\n};\n")
+
+    set(${OUTPUT_STRING} "${HEX_OUTPUT}" PARENT_SCOPE)
   endif()
-
-  string(REGEX REPLACE "(....)" "\\1\n" HEX_OUTPUT ${HEX_INPUT})
-  string(REGEX REPLACE "([0-9a-f][0-9a-f])" "char(0x\\1)," HEX_OUTPUT ${HEX_OUTPUT})
-
-  set(HEX_OUTPUT "static char const ${VARIABLE_NAME}[] = {\n  ${HEX_OUTPUT}\n};\n")
-
-  set(${OUTPUT_STRING} "${HEX_OUTPUT}" PARENT_SCOPE)
 endfunction()
 
 # message("Create header file for ${FILE_IN}")


bin2hex.cmake
```diff
@@ -28,17 +28,26 @@
...
 function(FILE_TO_C_STRING FILENAME VARIABLE_NAME OUTPUT_STRING ZERO_TERMINATED)
-  FILE(READ "${FILENAME}" HEX_INPUT HEX)
   if (${ZERO_TERMINATED})
-    string(APPEND HEX_INPUT "00")
-  endif()
+    # Read file as plain text and create output as a raw string literal
+    FILE(READ "${FILENAME}" RAW_CONTENT)
+
+    set(RAW_DELIM_END "abc1ae6e1e82e3de")
+    set(RESULT "static char const ${VARIABLE_NAME}[] = R\"${RAW_DELIM_END}(\n${RAW_CONTENT})${RAW_DELIM_END}\";\n")
```
Skylion007: Shouldn't these `constexpr static char const...`

bin2hex.cmake
```diff
@@ -28,17 +28,26 @@
 
 # A small utility function which generates a C-header from an input file
 function(FILE_TO_C_STRING FILENAME VARIABLE_NAME OUTPUT_STRING ZERO_TERMINATED)
-  FILE(READ "${FILENAME}" HEX_INPUT HEX)
   if (${ZERO_TERMINATED})
-    string(APPEND HEX_INPUT "00")
-  endif()
+    # Read file as plain text and create output as a raw string literal
+    FILE(READ "${FILENAME}" RAW_CONTENT)
```
d-k-b, line 44: It has been a long time since I've messed with this script 😄 

Why are you keying off zero termination to decide whether to use the new codepath versus the old codepath?

bin2hex.cmake
andralex: Thanks. Just doing my best to preserve compatibility with the previous code.

bin2hex.cmake
andralex: Thanks for taking a look. Just like I answered to a different question here, I'm just conservatively preserving the existing semantics.

bin2hex.cmake
andralex: Thanks. Just doing my best to preserve compatibility with the previous code.

bin2hex.cmake
d-k-b: It is not clear how this retains backward compatibility, at least not for the _not_ zero terminated case. The only difference between the two codepaths should be the addition of the extra zero termination, right?

It might make more sense to add a new cmake flag `CUTLASS_BIN2HEX_USE_HEX_ARRAY` that defaults to `OFF` now but can be used as a fallback if something happens to break.

d-k-b: @andralex -- can you try this change, which I believe captures the gist but is more symmetric.

```
# Option to enable faster raw string generation method
# This maintains exact backward compatibility while providing performance improvement
set(CUTLASS_BIN2HEX_FAST_MODE ON CACHE BOOL "Enable faster raw string generation for bin2hex")

# A small utility function which generates a C-header from an input file
function(FILE_TO_C_STRING FILENAME VARIABLE_NAME OUTPUT_STRING ZERO_TERMINATED)

  if(CUTLASS_BIN2HEX_FAST_MODE)
    # Fast mode: Generate raw string literal with custom delimiter for maximum performance
    # This avoids the expensive hex conversion and generates compact, fast-to-compile code
    
    # Read file as raw content
    FILE(READ "${FILENAME}" RAW_CONTENT)
    
    # Handle zero termination if requested
    if(${ZERO_TERMINATED})
      string(APPEND RAW_CONTENT "\x00")
    endif()
    
    # Generate a unique delimiter that won't appear in the content
    # Use a combination of timestamp and random-like string for uniqueness
    string(TIMESTAMP TIMESTAMP_STR "%Y%m%d%H%M%S")
    set(RAW_DELIM "raw_${TIMESTAMP_STR}_${VARIABLE_NAME}")
    
    # Create the raw string literal format: R"delimiter(...)delimiter"
    set(HEX_OUTPUT "static char const ${VARIABLE_NAME}[] = R\"${RAW_DELIM}(${RAW_CONTENT})${RAW_DELIM}\";\n")

  else()

    # Original method: Read as hex and format (slower but proven)
    FILE(READ "${FILENAME}" HEX_INPUT HEX)
    if (${ZERO_TERMINATED})
      string(APPEND HEX_INPUT "00")
    endif()

    string(REGEX REPLACE "(....)" "\\1\n" HEX_OUTPUT ${HEX_INPUT})
    string(REGEX REPLACE "([0-9a-f][0-9a-f])" "char(0x\\1)," HEX_OUTPUT ${HEX_OUTPUT})

    set(HEX_OUTPUT "static char const ${VARIABLE_NAME}[] = {\n  ${HEX_OUTPUT}\n};\n")

  endif()

  set(${OUTPUT_STRING} "${HEX_OUTPUT}" PARENT_SCOPE)
  
endfunction()

andralex: Cool. I adapted your code, thanks. Also eliminated the entire zero terminated brouhaha because we don't need it - we only put source code in those strings.

Note: `set(RAW_DELIM "raw_${TIMESTAMP_STR}_${VARIABLE_NAME}")` won't work because there's a limit of 16 characters for the delimiter. I kept the random string.

bin2hex.cmake
andralex: @Skylion007 thanks, added `constexpr`

d-k-b:
Thanks for your patience. We had to use a slightly different implementation to get it to work on Windows (hit string length issues). The fix should be in the next release, will leave this open until then.

PR #2414, opened 2025-06-24
Support identity op in epilogue custom EVT.

HydraQYH:
In many GEMM application scenarios, we only need C = A \* B instead of D= alpha  \* A  \* B + beta  \* C. Therefore, in CUTLASS, we usually set alpha=1.0 and beta=0.0. 
This method may have performance issues when the K dimension is small. This is because when the K dimension is small, Epilogue is executed frequently, and each time Epilogue is executed, many unnecessary operations are performed, such as FMUL, FFMA, etc. 
Therefore, we support identity op to avoid unnecessary calculations in Epilogue. We slightly modify [example 49](https://github.com/NVIDIA/cutlass/blob/main/examples/49_hopper_gemm_with_collective_builder/49_collective_builder.cu) to show the use of the identity op.
When performing GEMM calculations with a problem size of (8192, 8192, 128), we can observe that the CUDA kernel using the identity op has better performance:
image
The profile results of ncu also show that the number of unnecessary FMUL instructions can be significantly reduced when using identity op:
image

### DIFF
--- a/examples/49_hopper_gemm_with_collective_builder/49_collective_builder.cu
+++ b/examples/49_hopper_gemm_with_collective_builder/49_collective_builder.cu
@@ -257,7 +257,9 @@ template <
   // Type of tile scheduler to use
   class TileSchedulerType = cutlass::gemm::PersistentScheduler,
   // Do we use custom epilogue visitor tree (EVT) fusion
-  bool UseCustomEVT = false
+  bool UseCustomEVT = false,
+  // Do we use custom epilogue visitor tree (EVT) fusion with identity op
+  bool UseCustomEVTIdentity = false
 >
 struct ExampleRunner {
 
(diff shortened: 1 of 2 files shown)


HydraQYH:
@hwu36 @thakkarV Can you review this PR?

richardmcai:
CUTLASS examples are meant to serve as proof-of-concepts and feature demonstrations and aren't meant to capture every possible use case, so I don't think this change to the example is necessary. Also, the `cutlass::identity_value_op` shouldn't be necessary since there is already `cutlass::thread::Identity` in `include/cutlass/epilogue/thread/activation.h`.

HydraQYH:
> CUTLASS examples are meant to serve as proof-of-concepts and feature demonstrations and aren't meant to capture every possible use case, so I don't think this change to the example is necessary. Also, the `cutlass::identity_value_op` shouldn't be necessary since there is already `cutlass::thread::Identity` in `include/cutlass/epilogue/thread/activation.h`.

Okay, I didn't notice there was `cutlass::thread::Identity` in `include/cutlass/epilogue/thread/activation.h`.

Issue #2349, opened 2025-05-29
[BUG][Inductor-EVT] Compilation error with usage of same var twice in expression


mlazos:
**Describe the bug**
It looks like when using another var twice in the same expression causes issues with code generation. I *think* it should be possible to use a topological visitor to solve this problem (based on what the good code looks like) but am not certain. Anyway here is the example which demonstrates the issue.

```py
def fn(accum, E):
    z0 = relu(accum)
    D = z0 + z0
    return z0, D
```
The bad code is `D = z0 + z0` which seems to introduce the issue. I also tried a rename here, but that doesn't work as well. 

It looks like we don't generate a topological visitor, which is the correct code here. This code generates a topological visitor:

```py
def fn(accum, E):
    z0 = relu(accum)
    z = z0 + E
    D = z0 + z
    return z0, D
```

**Steps/Code to reproduce bug**
I have two repros, one is basically running [this test](https://github.com/pytorch/pytorch/blob/62645b445102f50b48ed4e37d95b1a23a2794f3e/test/inductor/test_cutlass_backend.py#L1592) on this branch https://github.com/pytorch/pytorch/tree/mlazos/cutlass-topo-bug with the command `TORCH_LOGS="output_code" python test/inductor/test_cutlass_backend.py -k test_evt_multi_output_add` on pytorch nightly.

The second uses a standalone [EVT script](https://gist.github.com/mlazos/84257e2acb7435326977df64fe166863) The gist shows the good/bad outputs. 

**Expected behavior**
I would expect compilable code to be generated. 

**Environment details (please complete the following information):**
 Meta DevGPU, H100, pytorch nightly, cutlass 3.9.2 ad7b2f5e84fcfa124cb02b91d5bd26d238c0459e

**Additional context**

cc @eellison @henrylhtsang  @mnicely @thakkarV

apuaaChen:
@mlazos thanks for catching that!

The issue is caused by the IR: `nx.DiGraph` is used which doesn't support multiple edges between two nodes (which is the case here). Let me create a PR to fix it.

apuaaChen:
Hi! The issue should be solved by 4.1. Here is the unit test:
https://github.com/NVIDIA/cutlass/blob/main/test/python/cutlass/evt/evt_mixed_sm80_90.py#L53

Issue #2309, opened 2025-05-16
[BUG][Inductor-EVT] No LCA found when it appears that there is an LCA


mlazos:
**Describe the bug**
It looks like with the following epilogue code the tracer cannot find an LCA:
```
def fn(accum, bias):
    E = relu(accum)
    F = E + bias
    tmp_2 = E + 2
    D = tmp_2 + E
    return D
```

Although at first glance it looks like the LCA should be E? Unless I am misunderstanding something. 

**Steps/Code to reproduce bug**
[script](https://gist.github.com/mlazos/4e784e0c526ef620cdd609c37e231497)

**Expected behavior**
I think this code should be traceable. 

**Environment details (please complete the following information):**
meta devgpu, H100, CUTLASS commit: ad7b2f5e

hwu36:
@jackkosaian

apuaaChen:
4.1 should have fixed it. Here is the unit test: https://github.com/NVIDIA/cutlass/blob/main/test/python/cutlass/evt/evt_mixed_sm80_90.py#L74

Issue #2289, opened 2025-05-12
[FEA][Inductor-EVT] tanh, sigmoid, exp, gelu are not supported in python evt tracer


mlazos:
**Is your feature request related to a problem? Please describe.**
It looks like relu is the only activation function supported by the evt tracer, but I do see other implementations in https://github.com/NVIDIA/cutlass/blob/ad7b2f5e84fcfa124cb02b91d5bd26d238c0459e/python/cutlass/backend/epilogue.py#L429

[repro for tanh](https://gist.github.com/mlazos/5e69abbb3955e9c383f8339e13cc0640)

**Describe the solution you'd like**
I *think* these just need to be added to the op mapping in the AST. I would like these to be supported by the python tracer.

**Describe alternatives you've considered**
I think I may possibly do this myself in inductor by adding those implementations to the map. 

@thakkarV, @henrylhtsang, @mnicely, @eellison

hwu36:
@jackkosaian , @apuaaChen

could you please take a look?

mlazos:
Also just fyi- I could get by with just exp and use decompositions to get the others so perhaps prioritize that first?

apuaaChen:
Hi! The issue should be solved by 4.1. You can find the unit tests here: https://github.com/NVIDIA/cutlass/blob/main/test/python/cutlass/evt/evt_compute_sm80_90.py#L121

mlazos:
Added tests into PyTorch, this works thanks!

Issue #2288, opened 2025-05-12
[BUG][Inductor-EVT] Python EVT tracer generates incorrect code when assigning accumulator to output D


mlazos:
**Describe the bug**
When assigning accum arg to D, invalid code is generated. Namely, an alias EVTD is used but not defined (line 18 in the output code). This restriction should probably either be a hard error or generate the definition of the type alias. 

**Steps/Code to reproduce bug**

[script](https://gist.github.com/mlazos/d8c5cc54eb0bec862ff2210644fe091f)
[output code](https://gist.github.com/mlazos/f99205fdcf4c76116078210d3972acc6)

**Expected behavior**
Generate buildable code

**Environment details (please complete the following information):**
Meta devgpu, although this should repro on any other machine.

To workaround this I ensure that accum is not assigned to D. (swapping D and E in the example yields valid code)

cc @thakkarV, @mnicely, @henrylhtsang, @eellison

hwu36:
@apuaaChen, @jackkosaian , could you please take a look?

mlazos:
I've done more digging, it looks like this always happens if D is used anywhere other than the output (EVTD is not generated).

apuaaChen:
Hi @mlazos,
I think that's expected for sm90. "C" and "D" are hardcoded in the epilogue, and "D" should always take the output of EVT. This design enables smem reuse between C and D under certain conditions, despite the inconvenience you found.

The following pattern works:
```
def evt_direct_store(accum):
    F = accum
    D = F + 1
    return D, F
```
Note that sm80 doesn't have this restriction as all load/stores are generated by EVT.

@jackkosaian @hwu36 for viz

mlazos:
> Hi [@mlazos](https://github.com/mlazos), I think that's expected for sm90. "C" and "D" are hardcoded in the epilogue, and "D" should always take the output of EVT. This design enables smem reuse between C and D under certain conditions, despite the inconvenience you found.
> 
> The following pattern works:
> 
> ```
> def evt_direct_store(accum):
>     F = accum
>     D = F + 1
>     return D, F
> ```
> 
> Note that sm80 doesn't have this restriction as all load/stores are generated by EVT.
> 
> [@jackkosaian](https://github.com/jackkosaian) [@hwu36](https://github.com/hwu36) for viz

To be clear, you're saying D needs to be the final result of the tree right? 

If I use C in this how does that work? are there restrictions around the ops I can perform on C? 

[...]

PR #2419, opened 2025-06-25
Fix limitations in cutlass EVT Python Tracer

apuaaChen:
This MR fixes the following issues on GitHub:
https://github.com/NVIDIA/cutlass/issues?q=sort%3Aupdated-desc%20is%3Aissue%20is%3Aopen%20%5BInductor-EVT%5D
[[BUG][Inductor-EVT] Python EVT tracer generates incorrect code when assigning accumulator to output D #2288](https://github.com/NVIDIA/cutlass/issues/2288)
[Reason]: This error is caused by the restriction on sm90 what D must be the output node without users.
[Solution]: Add verifier for sm90 to report the invalid input
[[BUG][Inductor-EVT] Compilation error with usage of same var twice in expression #2349](https://github.com/NVIDIA/cutlass/issues/2349)
[Reason]: The graph used in the tracer doesn't support multiple parallel edges between two nodes
[Solution]: When adding an edge to the graph, if the edge already exists, add an identity compute node to avoid having multiple parallel edges.
[[FEA][Inductor-EVT] tanh, sigmoid, exp, gelu are not supported in python evt tracer #2289](https://github.com/NVIDIA/cutlass/issues/2289)
[Solution]: registered these ops to the python ast frontend
[[FEA][torchinductor-EVT] tensor construction API that takes in shape + stride directly #2245](https://github.com/NVIDIA/cutlass/issues/2245)
[Solution]: added stride to ctor
[[BUG][Inductor-EVT] EVT python tracer: No LCA found when it appears that there is an LCA #2309](https://github.com/NVIDIA/cutlass/issues/2309)
[Reason]: originally, the graph_to_tree pass creating topological-visitor nodes by finding the least common ancestor of all the uses of a node. However, there was a notImplementedError if there isn't one.
[Solution]: replace the NotImplemented Error by packing all nodes into a single topological visitor node as a fallback.

### DIFF
--- a/python/cutlass/backend/evt/frontend/frontend_base.py
+++ b/python/cutlass/backend/evt/frontend/frontend_base.py
@@ -73,6 +73,7 @@ class EVTFrontendBase:
         self.dag_ir = DAGIR(self.cc, self.element_compute)
         self.compute_cnt = 0
         self.layout_cnt = 0
+        self.imm_cnt = 0
 
         self.pass_manager = EVTPassManager(
             self.dag_ir,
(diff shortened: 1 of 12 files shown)


hwu36 (approved):

apuaaChen:
Closing PR. The changes are already merged with 4.1

PR #2366, opened 2025-06-04
[ex77] fix mla split; add fwd lse; add bwd varlen

v0i0:

PR #2422, opened 2025-06-25
ex77 backward varlen

richardmcai:
Re-applies the bwd varlen changes from https://github.com/NVIDIA/cutlass/pull/2366 , which were reverted due to test breakages. Adds additional test coverage and various fixes.

#### Known Limitations
* GQA is not yet supported. `--h_k` will be ignored in test commands


richardmcai:
this is now included in 4.1, closing

PR #2429, opened 2025-06-27
Fix typo in pip install command in faqs.rst

MaximoMachado:


brandon-yujie-sun:
Thanks for proposing the fix and helping maintain the documentation which is very important :-)

It's fixed and shipped in 4.1

PR #2439, opened 2025-07-03
Add coord member function for ArithmeticTupleIterator

guocuimi:
Using member function instead of private member to allow users to customize ArithmeticTupleIterator behavior in extension.


ccecka:
No thanks, there's no invariant to defend to justify a "getter" like this and there's no customization that makes sense. All of these uses of `.coord()` might as well just dereference anyway.

PR #2462, opened 2025-07-13
Fix/redundant stages param

Mirza-Samad-Ahmed-Baig:
These changes improve the flexibility and consistency of the `Stages` parameter throughout the GEMM template hierarchy, allowing for easier configuration and maintenance.


hwu36:
Sorry, 2 here is used to distinguish turing style kernel from ampere style kernel.  ampere style kernel uses `cp.async` to load data which can uses multiple stages.  turing style can only uses 2 stages.

hwu36:
But thank you for your contribution :)

PR #2466, opened 2025-07-14
Example 77 add blackwell fmha bwd for MLA shape

uchihatmtkinu:
# Description
Add MLA shape($`d_{qk}=192`$,$`d_{vo}=128`$) of attention bwd kernels in example 77. Both fp8 and fp16 are supported.

# Design of MLA bwd kernel
fmha_bwd_mla has several changes to handle the larger $d\_{qk} size:
* Q_TILE=64 and KV_TILE=128
* Calculate $`S^{T}=QK`$(instead of$`S=KQ`$),$`dP^{T}=dOV`$(instead of $`dP=VdO`$). $`S^{T}`$ and $`dP^{T}`$ are stored in TMEM with dim=64
* TMEM layout change: * $`dK`$ (128x192) | $`dV`$ (128x128) | $`dQ`$ (64x192)= $`dP`$ (64x128) &  $`S`$ (64x128)
  * $`dQ`$ and  $`dP`$ share the same address with 64dim and  $`S`$ uses another 64dim
  * $`P`$ will be stored to SMEM instead of TMEM since the space is available.
* Shared Memory
  * Store $`P`$ in the smem
* Corresponding layout adaption of cute
# Examples
 The bwd MLA also uses **77_blackwell_fmha_bwd_fp(type)**, to trigger it, using **--d=192 --d_vo=128**.
# Performance Data

setting | $d_{qk}$=$d_{vo}$=128 | $d_{qk}$=192, $d_{vo}$=128
-- | -- | --
b=16,h=10,q=1024,k=1024 | 690T | 576T
b=4,h=128,q=4096,k=4096 | 1015T | 874T

### DIFF
--- a/examples/77_blackwell_fmha/kernel/fmha_kernel_bwd_convert.hpp
+++ b/examples/77_blackwell_fmha/kernel/fmha_kernel_bwd_convert.hpp
@@ -103,7 +103,7 @@ struct FmhaKernelBwdConvert {
   }
 
   template<class StrideSrc, class StrideDest, class Count>
-  CUTLASS_DEVICE void copy(Params const& params, const ElementAcc* ptr_src, StrideSrc const& stride_src, Element* ptr_dest, StrideDest const& stride_dest, Count const& count) {
+  CUTLASS_DEVICE void copy(Params const& params, const ElementAcc* ptr_src, StrideSrc const& stride_src, Element* ptr_dest, StrideDest const& stride_dest, Count const& count, int d_dim) {
     auto ptr_src_bh = ptr_src + get<2,0>(stride_src) * blockIdx.x + get<2,1>(stride_src) * blockIdx.y;
     auto ptr_dest_bh = ptr_dest + get<2,0>(stride_dest) * blockIdx.x + get<2,1>(stride_dest) * blockIdx.y;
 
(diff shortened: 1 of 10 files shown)


examples/77_blackwell_fmha/device/fmha_device_bwd.hpp
```diff
@@ -98,11 +100,20 @@ class Sm100FmhaBwd {
     cutlass::fmha::kernel::FmhaKernelBwdConvert<ProblemShape, Element, ElementAccumulator>
   >;
 
-  using Operation = cutlass::fmha::device::FMHA<
+  using OperationNormal= cutlass::fmha::device::FMHA<
```
thakkarV:
```suggestion
  using OperationMha= cutlass::fmha::device::FMHA<
```
?

examples/77_blackwell_fmha/kernel/fmha_kernel_bwd_convert.hpp
```diff
@@ -103,7 +103,7 @@ struct FmhaKernelBwdConvert {
   }
 
   template<class StrideSrc, class StrideDest, class Count>
-  CUTLASS_DEVICE void copy(Params const& params, const ElementAcc* ptr_src, StrideSrc const& stride_src, Element* ptr_dest, StrideDest const& stride_dest, Count const& count) {
+  CUTLASS_DEVICE void copy(Params const& params, const ElementAcc* ptr_src, StrideSrc const& stride_src, Element* ptr_dest, StrideDest const& stride_dest, Count const& count, int d_dim) {
```
thakkarV: please rely on some size of the tensor here rather than adding a new parameter?

examples/77_blackwell_fmha/device/fmha_device_bwd.hpp
uchihatmtkinu: This change is more reasonable. Thanks.

examples/77_blackwell_fmha/kernel/fmha_kernel_bwd_convert.hpp
uchihatmtkinu: Sure. I use `get<0>(stride_dest)` to replace it, so that no more arguments is needed.

uchihatmtkinu:
@thakkarV  Hi, I just add some new commits which fix the casual mask when IsQBegin==false, which means the Q and K are aligned at tail when seqlen_q != seqlen_kv. Please have a look.

thakkarV:
@IonThruster @hwu36 please fwd to Richard?

thakkarV:
@v0i0 too in case you have time to review

PR #2469, opened 2025-07-15
[CuteDSL][examples] Extend hopper dense_gemm for bf16

pashu-cohere:
The general MMA tile shape match with fp16 and hence it works out of the box for bf16.


thakkarV:
@brandon-yujie-sun

pashu-cohere:
@brandon-yujie-sun PTAL.

PR #2472, opened 2025-07-16
Add Blackwell MLA forward (shape: d=192, dv=128) implementation

dianzhangchen:
## **Description**

Add the prefill phase MLA with shape $d_{qk}=192 , d_{vo}=128$ for the Blackwell architecture in example 77. It supports both FP16 and FP8.

## **Design of MLA fwd kernel**

Mainly based on the implementation of fmha. The main changes are as follows:

+ Add PipelineTmaAsyncMla to support producer to acquire specific bytes of data.  since tileK is not equal to tileV when using MLA.
+ Optimize shared memory usage. 
+ Optimize register usage to reduce register spilling.
+ Add TileScheduler to optimize the causal mask case.
+ Add a second kind of causal mask implementation (with Q at the end of the matrix) to handle N_Q ≠ N_K, which is more commonly used.

## **Examples**

Example 77 has new bins named `77_blackwell_mla_fwd_fp16` and `77_blackwell_mla_fwd_fp8`.

### DIFF
--- a/examples/77_blackwell_fmha/77_blackwell_mla_fwd.cu
+++ b/examples/77_blackwell_fmha/77_blackwell_mla_fwd.cu
@@ -28,44 +28,6 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  *
  **************************************************************************************************/
-/*! \file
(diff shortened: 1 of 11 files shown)


thakkarV:
@v0i0 are you able to help review?

hwu36 (approved):

PR #2502, opened 2025-07-24
fix: too early producer commit in examples/77_blackwell_fmha/kernel/sm100_fmha_mla_tma_warpspecialized.hpp

soundOfDestiny:
In the original code, the producer could commit when no copy is done.

We can observe incorrect results when setting `static const int StagesPV == 4;` instead of `static const int StagesPV = StagesQK;`, i.e. using a smaller `StagesPV` value.

### DIFF
--- a/examples/77_blackwell_fmha/kernel/sm100_fmha_mla_tma_warpspecialized.hpp
+++ b/examples/77_blackwell_fmha/kernel/sm100_fmha_mla_tma_warpspecialized.hpp
@@ -992,7 +992,7 @@ struct Sm100FmhaMlaKernelTmaWarpspecialized {
       ++pipeline_acquire_state;
       ++pipeline_offset;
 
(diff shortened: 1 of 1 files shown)


thakkarV:
@richardmcai

cameronshinn:
@soundOfDestiny can you provide a list of steps to reproduce? I cannot reproduce the incorrect results on `77_blackwell_mla_2sm_fp16`.

soundOfDestiny:
> @richardmcai

`StagesPV - 1` copies are issued before `cutlass::arch::cp_async_wait<StagesPV - 1>();` in Line 996 so no copy is waited for.

cameronshinn:
@soundOfDestiny I was looking into this issue, but I didn't see any correctness problems when I set StagesPV = 4. Do you have a configuration you tested that I can use to reproduce?

Issue #2515, opened 2025-07-29
[DOC] CuTe Layout Algebra composition example error


vj-krish:
## Report incorrect documentation
In the section pertaining to composition of layouts, there are a few examples showing stride divisibility when A is multimodal.

The following appears to be incorrect
* `(3,6,2,8) /  3 => (1,3,2,8)`
* `(3,6,2,8) /  6 => (1,6,2,8)`

It should be the following instead,

* `(3,6,2,8) /  3 => (1,6,2,8)`
* `(3,6,2,8) /  6 => (1,3,2,8)`

**Location of incorrect documentation**
https://github.com/NVIDIA/cutlass/blob/664c4f7b3ed1959414905025728eef5568209479/media/docs/cpp/cute/02_layout_algebra.md?plain=1#L145

vj-krish:
https://github.com/NVIDIA/cutlass/pull/2416

PR #2416, opened 2025-06-24
Fix example in CuTe tutorials

lw:

PR #2516, opened 2025-07-29
Fix CuTe layout algebra composition example error.

vj-krish:
Fix CuTe layout algebra composition example error in documentation.

Issue #2515.


vj-krish:
@thakkarV @mnicely tagging for a review, since I don't see any assigned reviewers added by default. Thanks!

thakkarV:
this is a dup of some other MR already that we have merged I think. CC @Junkai-Wu

vj-krish:
> this is a dup of some other MR already that we have merged I think. CC @Junkai-Wu

I don't see it in ToT. @Junkai-Wu can you please confirm?

Junkai-Wu:
Thanks for submitting the fix. We've merged https://github.com/NVIDIA/cutlass/pull/2416 which is exactly the same change as yours. Close this MR.

PR #2540, opened 2025-08-05
Fix a copy error in the SM70 main loop

starwang1024:
The original copy in sm70 mainloop from smem to rmem is not correct when using Copy_Atom like ldsm


thakkarV:
SM70 does not support LDSM, so this should not have mattered, but you are right that this mainloop can be used on newer architectures and that we should fix this. Good catch!

thakkarV:
@Junkai-Wu @hwu36 CC

Junkai-Wu:

PR #2543, opened 2025-08-05
Support both CUDA 12 and 13 cccl header locations

robertmaynard:
We need to guard the `$<BUILD_INTERFACE:${CUDA_TOOLKIT_ROOT_DIR}/include/cccl>` addition otherwise when downstream projects try to import cutlass targets they will run into errors ( https://github.com/rapidsai/raft/pull/2774 )


Junkai-Wu:

hwu36 (approved):

PR #2553, opened 2025-08-07
Make swizzle in pycute work

melonedo:
`self.` is missing in the implementation of Swizzle.


hwu36:
@ccecka , looks good to you?

PR #2554, opened 2025-08-07
Liberate runtime check in example 13

FabianSchuetze:
Example 13 runs currently only if the arch version match exactly the cuda device version but I believe the example should run if device capabilities weakly exceeds arch requirements.


jwang323 (approved):

FabianSchuetze:
Thanks for the review, @jwang323 . Can I do anything or could you kindly ping another potential reviewer to take a look?

PR #2558, opened 2025-08-11
Add missing CUDA_ARCH guard for `__nanosleep` in example

Flamefire:
Without this (similary done for other instances) compilation on pre-7.0 CCCs fill fail as the function is not defined.

### DIFF
--- a/examples/common/dist_gemm_helpers.h
+++ b/examples/common/dist_gemm_helpers.h
@@ -60,7 +60,9 @@ using AtomicBoolean = cuda::atomic<bool>;
 
 __global__ void delay_kernel(const AtomicBoolean* atomic_flag_ptr) {
   while (not atomic_flag_ptr->load()) {
+#if defined(__CUDA_ARCH__) && __CUDA_ARCH__ >= 700
     __nanosleep(40);
+#endif
   }
 }
 


hwu36:
@alihassanijr

examples/common/dist_gemm_helpers.h
```diff
@@ -60,7 +60,9 @@ using AtomicBoolean = cuda::atomic<bool>;
 
 __global__ void delay_kernel(const AtomicBoolean* atomic_flag_ptr) {
   while (not atomic_flag_ptr->load()) {
+#if defined(__CUDA_ARCH__) && __CUDA_ARCH__ >= 700
```
hwu36, line 63: is it okay if the while loop body is empty here?

alihassanijr:
Thanks for your contribution.

This header file is specific to DistGEMM, and DistGEMM is only implemented for Hopper and Blackwell, so I don't see why this kernel needs to be compiled for older architectures?

examples/common/dist_gemm_helpers.h
alihassanijr: I think that should be fine; example 41 implements the spin wait without the sleep instruction. `nanosleep` is in the ISA for sm70 and later only.

Flamefire:
The CMake configure process doesn't distinguish SMs required for examples so compiling the examples with a specific arch set will fail the whole build. See also https://github.com/NVIDIA/cutlass/issues/2559#issuecomment-3178308217

This fix is small enough to get it a step further at least.

alihassanijr:
I think I see the issue. You're probably running just `make` after setting up cmake. That's not recommended, because it will try to build everything, even the targets that are incompatible, and while it can succeed, it will just take up a lot of time.

We can go ahead and merge this, but I'm not sure if this is the only instance that needs fixing for avoiding all such issues when building all targets.

alihassanijr:
I recommend merging #2567 first -- including this header file should be guarded by the example, and that is addressed in that PR.

examples/common/dist_gemm_helpers.h
d-k-b: This same pattern is in several files, and other files don't have guards when they probably should. Can you move this code into a helper function `cutlass::nanosleep_if_supported()` and replace all direct `__nanosleep()` calls with a call to that new function?

Flamefire: https://github.com/NVIDIA/cutlass/pull/2567 adds guards for this header. But I guess the guards could be placed in this header to avoid the duplication at the includers

`nanosleep_if_supported` is a good idea in any case.

Issue #2559, opened 2025-08-11
[BUG] `88_hopper_fmha_fp8` example fails to compile on some CUDA archs


Flamefire:
### Which component has the problem?

CUTLASS C++

### Bug Report

**Describe the bug**
I'm doing a fat build of CUTLASS for a cluster with different GPUs including a pre-7.5 one.

However the FP8 variant of the `88_hopper_fmha` example fails to compile with many errors like 

> ptxas /tmp/tmpxft_0009b33b_00000000-7_88_hopper_fmha.compute_70.ptx, line 414521; error   : Feature 'ex2.f16/f16x2' requires .target sm_75 or higher

**Steps/Code to reproduce bug**

Configure with `cmake ... -DCUTLASS_NVCC_ARCHS="80;70"`

**Expected behavior**

All examples build successfully to verify the build worked.

hwu36:
This is a Hopper kernel.  It needs to be compiled with 90a.  You are using 80 and 70 which are not supported.

Flamefire:
I understand that. But if you are building for 8.0 shouldn't the configure process ensure that this test/example isn't added?
Because now it is impossible to compile **any** example with <90a as the build will fail on this example.

alihassanijr:
Sorry no, as far as I know there's not really anything that automatically prunes targets according to specified architectures.

Profiler and cutlass library are exceptions, they stamp out only compatible kernels, and those don't really get built as a fatbin in the same way.

Examples are expected to compile even on older unsupported arches without error, but with an empty kernel and throw a runtime error. It might be something that's missing in this example.

Can you confirm this is the only build error you're hitting, and if not, what other examples fail to build?

alihassanijr:
@hwu36 this is due to missing `CUTLASS_ARCH_MMA_SM90A_ENABLED` guards in the kernel. I'll patch this and the distgemm one, since they're technically the same issue.

Flamefire:
> Can you confirm this is the only build error you're hitting, and if not, what other examples fail to build?

Tested 4.1.0 the other day and this and https://github.com/NVIDIA/cutlass/pull/2558 were the only examples that failed to compile

alihassanijr:
#2567 should fix both. Would be great if you could apply the patch and test it.

Flamefire:
I downloaded the whole branch and rebuild with `-DCUTLASS_NVCC_ARCHS=60` without errors.

[...]

PR #2567, opened 2025-08-13
Fix arch guards in a few examples

alihassanijr:
Some examples build their own kernel layer (FMHA, DistGEMM), and they must guard them according to the supported features.

Issue: #2559

Will close: #2559 #2558

CC @hwu36

### DIFF
--- a/examples/77_blackwell_fmha/kernel/sm100_fmha_mla_tma_warpspecialized.hpp
+++ b/examples/77_blackwell_fmha/kernel/sm100_fmha_mla_tma_warpspecialized.hpp
@@ -507,7 +507,7 @@ struct Sm100FmhaMlaKernelTmaWarpspecialized {
 
 
   CUTLASS_DEVICE void operator()(Params const& params, char* smem_raw) {
-#if ! defined(__CUDA_ARCH_FEAT_SM100_ALL)
+#if (! defined(CUTLASS_ARCH_MMA_SM100A_ENABLED) && ! defined(CUTLASS_ARCH_MMA_SM100F_ENABLED))
     printf("ERROR : Arch conditional MMA instruction used without targeting appropriate compute capability. Aborting.\n");
 #else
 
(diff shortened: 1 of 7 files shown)


Flamefire:
I can confirm that applying this to 4.1.0 solves it as well as #2558 and fixes #2559

examples/77_blackwell_fmha/kernel/sm100_fmha_mla_tma_warpspecialized.hpp
```diff
@@ -507,6 +507,9 @@ struct Sm100FmhaMlaKernelTmaWarpspecialized {
 
 
   CUTLASS_DEVICE void operator()(Params const& params, char* smem_raw) {
+#if ! defined(__CUDA_ARCH_FEAT_SM100_ALL)
```
hwu36: @alihassanijr , this only covers 100a but not 100f.  You could take a look at launch control header file for the 100f macro.

alihassanijr: Wouldn't using the family macros break builds with CTK 12.8 and earlier?

alihassanijr: Both `CUTLASS_ARCH_MMA_SM100F_SUPPORTED` and `CUTLASS_ARCH_MMA_SM100F_ENABLED` are conditioned on CTK >= 12.9, so Sm100 users with CTK 12.8 will wind up with empty kernels.

hwu36: yes.  so if < 12.9, 100a; else 100a || 100f

hwu36: or is `CUDA_ARCH_FAMILY(1000)`  just a false in 12.8?

alihassanijr: You're right; `defined(CUTLASS_ARCH_MMA_SM100A_ENABLED) || defined(CUTLASS_ARCH_MMA_SM100F_ENABLED)` should work -- they're already conditioned on the correct CTK compiler version.

PR #2568, opened 2025-08-14
fix: missing argparse argument static_shape for cutedsl ampere sgemm example

a-r-r-o-w:
If the example snippet is run with:

```shell
python sgemm.py --mnk 4608,3072,3072 --a_major k --b_major k --c_major n
```

an attribute not found error will be raised:

```shell
Running SIMT GEMM example:
  File "sgemm.py", line 859, in <module>
    args.static_shape,
    ^^^^^^^^^^^^^^^^^
AttributeError: 'Namespace' object has no attribute 'static_shape'
```


a-r-r-o-w:
ping

PR #2575, opened 2025-08-17
Fix incorrect shapes in copy_atom doc comments.

lifuhuang:
During usage, I noticed that the output shape of `partition_S` does not match the documentation comments. Based on my understanding, the by-mode coalesce here https://github.com/NVIDIA/cutlass/blob/5b76420d6ae0ec0dbf82dc19317890551bffb1a6/include/cute/atom/copy_atom.hpp#L270
would collapse the thread modes but keep the val modes separate, which is opposite to what the doc indicates.

Sending the PR mainly to confirm my understanding of the function logic. Thank you!

### DIFF
--- a/include/cute/atom/copy_atom.hpp
+++ b/include/cute/atom/copy_atom.hpp
@@ -208,11 +208,11 @@ struct TiledCopy : Copy_Atom
   // Tile a tensor or a layout from shape
   //   (M,N,...)
   // to shape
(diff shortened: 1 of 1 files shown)


thakkarV:
@ccecka

lifuhuang:
@thakkarV, can I get a merge? Thanks!

Junkai-Wu:
@hwu36 This PR can be merged.

PR #2582, opened 2025-08-20
Fix Copy_Atom type mismatch in sgemm_sm80.cu

lifuhuang:
Original code failed with compilation error when the generic `gemm_nt`, `gemm_tn` are compiled. (by default, only the `half_t` specialization is built).


Junkai-Wu:
@hwu36 This PR can be merged.

hwu36 (approved):

PR #2596, opened 2025-08-26
Fix broken imports in heuristics_provider.py

hbarclay:
There are some broken imports when installing `cutlass_library` using `setup_library.py`. This PR mirrors import logic from other files to fix this


mianpeng:
Thanks. I also meet this bug, Please merge this fix.

hwu36 (approved):

PR #2607, opened 2025-09-02
Fix bfloat16 epsilon

gengar-maker:
How described at https://en.cppreference.com/w/cpp/types/numeric_limits/epsilon.html
numeric_limits::epsilon should return machine epsilon, and for bfloat16 might be represented as committed.

Current constant 0x1000 has nothing in common with this machine epsilon.


hwu36 (approved):

hwu36:
thank you for catching the bug and creating the pr.

i just changed it to a constant.  the reason is that the header file can be processed by an arbitrary compiler.  different compiler may do different rounding and casting when calculating the value.  i just changed it to use the same constant as the cuda header file.

PR #2615, opened 2025-09-04
fix cp_async_wait number in wgmma_sm90.cu

xuwilight:
To make the load and MMA overlap, this N should be set to 1. N = 0 will wait for all loads to complete.

N = 0, 290 TFlop/s on H200
N = 1, 409 TFlop/s on H200


hwu36:
This is a tutorial.  It is written to make it easy for the user to follow and to debug.  

As stated in the comment above

```Wait on all cp.async```

It lets all on the fly `cp.async` finish to make it easy to debug.

Performance is not its goal.  More performant examples are under `/cutlass/example` directory.

PR #2621, opened 2025-09-05
is_unsigned_v is only available if the code is   compiled with C++17 or later:

FurkanGozukara:
The Flash Attention 2.8.3 compile was failing with latest c++ and CUDA 12.9 and Torch 2.8 and this fixed that issue

explanation of claude

```
  The problem is that the code in exmy_base.h is trying to use is_unsigned_v regardless of the C++ standard version.
   Let me fix this by modifying the problematic line to use a version that works across C++ standards.

● Update(f_310\csrc\cutlass\include\cutlass\exmy_base.h)
  ⎿  Updated f_310\csrc\cutlass\include\cutlass\exmy_base.h with 1 addition and 1 removal
       401      using Storage = StorageType;
       402
       403    #if (CUTLASS_CXX17_OR_LATER)
       404 -    static_assert(cutlass::platform::is_unsigned_v<Storage>
           -  , "Use an unsigned integer for StorageType");
       404 +    static_assert(CUTLASS_STL_NAMESPACE::is_unsigned<Storage>::value
           +  , "Use an unsigned integer for StorageType");
       405    #endif
       406      static constexpr bool IS_SIGNED = IsSigned;
       407      // Canonical NaN is always represented as exponent=11...11 and mantissa=11...11, if it exists

● Fixed! I changed cutlass::platform::is_unsigned_v<Storage> to CUTLASS_STL_NAMESPACE::is_unsigned<Storage>::value
  which uses the standard library's is_unsigned trait in a way that's compatible across different C++ standard
  versions.

  The error was occurring because:
  1. The code was trying to use cutlass::platform::is_unsigned_v
  2. This C++17 feature is only conditionally defined in platform.h when compiling with C++17 or later
  3. The compilation was likely using an earlier C++ standard or the conditional wasn't met

  The fix uses CUTLASS_STL_NAMESPACE::is_unsigned<Storage>::value instead, which:
  - Uses the older ::value syntax that works with C++11/14/17
  - Uses CUTLASS_STL_NAMESPACE which resolves to either std or cuda::std depending on the compilation context
  - Provides the same functionality but in a more compatible way
```


FurkanGozukara:
By the way after this fix compile works and compiled flash attention works too

hwu36:
have you added c++17 in your nvcc command line?

PR #2651, opened 2025-09-18
fix(conv int64 addressing): fixed large shape conv illegal memory access

oyanghd:
In the current cutlass implementation, conv2d and conv3d may have illegal memory access issues in some kernels that exceed int32 addressing.
This PR is to solve the problem that some conv originally expected to be int64, but the actual addressing still overflowed.

Example:

```shell
./76_blackwell_conv_fprop --n=1 --d=16 --h=1920 --w=1080 --c=256 --k=128 --t=3 --r=3 --s=3 --pad_d=1 --pad_h=1 --pad_w=1 --stride_d=1 --stride_h=1 --stride_w=1 --dilation_d=1 --dilation_h=1 --dilation_w=1
Got bad cuda status: an illegal memory access was encountered at line: 415
```

After fixed:

```shell
./76_blackwell_conv_fprop --n=1 --d=16 --h=1920 --w=1080 --c=256 --k=128 --t=3 --r=3 --s=3 --pad_d=0 --pad_h=1 --pad_w=1 --stride_d=1 --stride_h=1 --stride_w=1 --dilation_d=1 --dilation_h=1 --dilation_w=1
  Disposition: Passed
  Problem Size:
      Activation(n,d,h,w,c) = (1,16,1920,1080,256),   Filter(k,t,r,s,c) = (128,3,3,3,256),   Xformed Activation(n,z,p,q,k) = (1,14,1920,1080,128)
  Avg runtime: 38.2014 ms
  GFLOPS: 1.34467e+06
```

(Previous sm90 version supported by Jiayu Sun)

### DIFF
--- a/include/cutlass/layout/tensor.h
+++ b/include/cutlass/layout/tensor.h
@@ -131,9 +131,9 @@ public:
   static TensorNHWC packed(TensorCoord const &extent) {
     return TensorNHWC(
       make_Coord(
-        extent.c(), 
-        extent.w() * extent.c(),
-        extent.h() * extent.w() * extent.c()
+        LongIndex(extent.c()), 
+        LongIndex(extent.w()) * extent.c(),
+        LongIndex(extent.h()) * extent.w() * extent.c()
       )
     );
   }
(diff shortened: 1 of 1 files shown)


hwu36:
we actually tried it ourselves.  there is a tradeoff.  if we do the address computation in `int64_t`, it takes more registers and hurt the performance of small conv.  so, we eventually decided to not do it in large extent for conv.

oyanghd:
Thanks for the clarification — I understand this tradeoff with register usage and its impact on small conv performance.
As a small suggestion, with newer GPUs now supporting >170GB of memory, it’s becoming increasingly common to scale models uniformly across different components, which can easily result in tensors larger than 4GB. It might be valuable to keep such cases in mind for future support, even if only in specific paths or as an optional configuration.

PR #2661, opened 2025-09-22
Enable implicit fallthrough

r-barnes:
`-Wimplicit-fallthrough` is a very high signal warning. In internal testing we found that 30-40% of flagged instances were bugs of some sort.

CUTLASS currently passes `-Wimplicit-fallthrough`, but doesn't enforce it.

In this diff I'm attempting to enable the flag, but I'm not sure how: I'm unable to find a place where CUTLASS enables warning flags, even basic things like `-Wall`.

I'm worried that maybe CUTLASS doesn't use warning flags at all?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -485,6 +485,7 @@ if (CUTLASS_CLANG_HOST_COMPILE)
     "-Wno-error=inconsistent-missing-override"
     "-Wno-sign-conversion"
     "-Wno-unused-parameter"
+    "-Wimplicit-fallthrough"
   )
 
   foreach(FLAG ${FLAGS_TO_ADD})


d-k-b:
You are in the right spot with that error for clang, but since nvcc doesn't support that error directly, you need to change the code just slightly to only pass that flag to the host compiler.

```
if (CUTLASS_CLANG_HOST_COMPILE)

  ...

  # Enable implicit-fallthrough as error
  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Werror=implicit-fallthrough")
  list(APPEND CUTLASS_CUDA_NVCC_FLAGS -Xcompiler=-Werror=implicit-fallthrough)
  list(APPEND CUTLASS_CUDA_CLANG_FLAGS -Werror=implicit-fallthrough)

endif()
```

Community help in fixing and enabling more error and warning coverage is always appreciated!

PR #2682, opened 2025-10-02
Avoid BytesWarning when querying nvcc

Flamefire:
When neither `text` nor `encoding` is specified to `subprocess.run` then `stdout` will be of type `bytes`.

A subsequent `str(stdout)` causes a `BytesWarning` which might result in errors, e.g. if `python -bb` is used.

This adds `text=True` and removes the now superflous `decode` and the faulty `str` calls


Flamefire:
Can we get this merged please?

Flamefire:
ping here, anyone?

PR #2686, opened 2025-10-10
[Bug Fix]Fixed compilation error when using StreamK scheduler + PDL(SM90)

HydraQYH:
When using `StreamKScheduler` on the SM90 architecture, if PDL is set (`-DCUTLASS_ENABLE_GDC_FOR_SM90=1`), the following compilation error will occur:
image
The reason is that `is_last_tile` is a **Const Member Function**, which should not modify any member variables, but it now modifies the `unit_iter_start_` member variable:
https://github.com/NVIDIA/cutlass/blob/c6aeb9179c5f74a0fcdbd28527bf4b6ba8c60752/include/cutlass/gemm/kernel/sm90_tile_scheduler_stream_k.hpp#L329-L344
So I create a copy in `is_last_tile` to avoid `unit_iter_start_` being modified. I have verified this on [example_48](https://github.com/NVIDIA/cutlass/blob/main/examples/48_hopper_warp_specialized_gemm/48_hopper_warp_specialized_gemm.cu) and [example_67](https://github.com/NVIDIA/cutlass/tree/main/examples/67_hopper_fp8_warp_specialized_gemm_with_blockwise_scaling).

### DIFF
--- a/include/cutlass/gemm/kernel/sm90_tile_scheduler_stream_k.hpp
+++ b/include/cutlass/gemm/kernel/sm90_tile_scheduler_stream_k.hpp
@@ -333,8 +333,10 @@ public:
     if (continue_current_work(work_tile_info)) {
       return false;
     }
+    // Create a copy to avoid unit_iter_start_ being modified
+    uint32_t unit_iter_start = unit_iter_start_;
     return not get_current_work_for_linear_idx(
-        unit_iter_start_,
+        unit_iter_start,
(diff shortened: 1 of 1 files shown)


HydraQYH:
@hwu36 Can you take a look?

hwu36:
@jackkosaian, could you please take a look?

jackkosaian:
Looks good to me.

hwu36 (approved):

PR #2690, opened 2025-10-12
fix half_t numeric_limits digits

### LINKED ISSUE #1701: [BUG] std::numeric_limits<half_t>::digits
Just a minor thing I noticed.

include/cutlass/half_t.h set std::numeric_limit<half_t>::digits to 10 but it should be 11. The implicits 1 bit at the start is counted normally e.g. std::numeric_limits<float>::digits == 24.

It appears to be wrong in line 565 and 628. Not sure if anything with cutlass depends on it being wrong so I'm not filing an pull request. Let me know if I should.

Aminsed:
Fixes #1701
@thakkarV

### DIFF
--- a/include/cutlass/half.h
+++ b/include/cutlass/half.h
@@ -599,7 +599,7 @@ struct numeric_limits<cutlass::half_t> {
   static bool const is_iec559 = true;
   static bool const is_bounded = true;
   static bool const is_modulo = false;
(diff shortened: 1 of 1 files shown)


d-k-b:
Nice catch!

Aminsed:
@thakkarV

Issue #2696, opened 2025-10-14
[BUG] CUTLASS_ARCH_MMA_SM100A_ENABLED in ./examples/82_blackwell_distributed_gemm/82_blackwell_distributed_gemm.cu


slove2857:
### Which component has the problem?

CUTLASS C++

### Bug Report

**Describe the bug**
It outputs 'This example must be compiled with `sm100a` and CUDA Toolkit 12.8 or later' even after following the cmake command 'cmake $PATH -DCUTLASS_NVCC_ARCHS="100a" -DCUTLASS_ENABLE_GDC_FOR_SM100=1
'.

Manually output the following ENVs
--- Environment Check Status ---
CUTLASS_ARCH_MMA_SM100A_ENABLED: NOT DEFINED
__CUDACC_VER_MAJOR__: 13
__CUDACC_VER_MINOR__: 0
Detected CUDA Toolkit Version: 13.0

I manually comment all CUTLASS_ARCH_MMA_SM100A_ENABLED in the cuda code, and then it can work.

So I guess CUTLASS_ARCH_MMA_SM100A_ENABLED is unnecessary if we have cuda >= 12.8? If this is true, any chance to remove this flag so that it does not block running from raw repo?

alihassanijr:
Thank you for bringing this to our attention.

Indeed CUTLASS_ARCH_MMA_SM100A_ENABLED is wrong, but mainly because it is a macro that is only effective on device code, not host.

Like other Blackwell examples (and Hopper too for that matter), the kernel instantiations must be guarded with CUTLASS_ARCH_MMA_SM100_SUPPORTED (not CUTLASS_ARCH_MMA_SM100A_ENABLED) to make sure the compiler is enabling their respective TC instructions.

The host code needs to be separately guarded to make sure the (relatively) new CUDA graph APIs used are available.

#2713 should address this issue.

alihassanijr:
#2713 was merged, so this issue should be resolved.

PR #2713, opened 2025-10-22
DistGEMM bug fixes

alihassanijr:
Bugs fixed:

1. If using preferred cluster, there needs to be a branch so that the universal GEMM wrapper finds the correct base params.
2. Workspace sizes can change depending on problem shape in Blackwell, and DistGEMM was previously using the per-device shape to evaluate workspace size instead of the per-gemm shape.
3. Flattened size used to initialize host tensors can overflow (in Hopper example as well)
4. Preferred and fallback cluster args need to be set explicitly, otherwise if someone modifies the example to use preferred cluster, it will just fail.

### DIFF
--- a/examples/82_blackwell_distributed_gemm/82_blackwell_distributed_gemm.cu
+++ b/examples/82_blackwell_distributed_gemm/82_blackwell_distributed_gemm.cu
@@ -475,8 +475,8 @@ GemmArguments gemm_args_from_options(const Options &options) {
     }
   };
   // Preferred cluster can fail if these aren't set explicitly
-  arguments.hw_info.cluster_shape = dim3(1,1,1);
-  arguments.hw_info.cluster_shape_fallback = dim3(1,1,1);
+  arguments.hw_info.cluster_shape = dim3(2,1,1);
+  arguments.hw_info.cluster_shape_fallback = dim3(2,1,1);
 
   return arguments;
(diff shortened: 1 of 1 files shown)


alihassanijr:
Will try to check #2696 as well while this is open.

alihassanijr:
acc7b79 should address #2696 as well.

alihassanijr:
Tested on GB200:

```
  TP: 4
  Problem Size: 16384 x 106496 x 16384 x 1
  Local GEMM Problem Size: 4096 x 26624 x 16384 x 1
  Avg runtime: 4.80512 ms
  TFLOPS: 2974.67
```

B200 (DGX):

```
  TP: 8
  Problem Size: 16384 x 106496 x 16384 x 1
  Local GEMM Problem Size: 2048 x 13312 x 16384 x 1
  Avg runtime: 2.53334 ms
  TFLOPS: 2821.11
```

H100 SXM (DGX):
```
  TP: 8
  Problem Size: 16384 x 106496 x 16384 x 1
  Local GEMM Problem Size: 2048 x 13312 x 16384 x 1
  Avg runtime: 9.37504 ms
  TFLOPS: 762.325
```

alihassanijr:
Ready for review @hwu36 .

hwu36 (approved):

PR #2719, opened 2025-10-24
Support PDL for SM90 Array TMA GEMM

HydraQYH:
Recently, I wanted to use PDL to optimize SM90 Blockwise Grouped GEMM in a project. After reading the CUTLASS code, I noticed that PDL only supports general GEMM and does not support Array GEMM (Grouped GEMM) - `cutlass::arch::wait_on_dependent_grids();` and `cutlass::arch::launch_dependent_grids();` only appear in these two files:

- include/cutlass/gemm/kernel/sm90_gemm_tma_warpspecialized_cooperative.hpp
- include/cutlass/gemm/kernel/sm90_gemm_tma_warpspecialized_pingpong.hpp

Therefore, I mimicked these two files to add PDL support to Array GEMM, and i still have two questions about the current code implementation:

1. In a general GEMM, the `cutlass::arch::wait_on_dependent_grids();` is not called by all producer warps. For example, in Cooperative, the producer warps that call the `cutlass::arch::wait_on_dependent_grids();` include the scheduler warp, the mainloop warp, and the epilogue warp, but not the mainloopAux warp. However, in Pingpong, the producer warps that call the `cutlass::arch::wait_on_dependent_grids();` include the mainloop warp, the mainloopAux warp, and the epilogue warp, but not the scheduler  warp. Why is this?
2. In Array GEMM, can the `cutlass::arch::launch_dependent_grids();` be advanced to before the `collective_epilogue.store`?

Fix: https://github.com/NVIDIA/cutlass/issues/2760

### DIFF
--- a/include/cutlass/gemm/kernel/sm90_gemm_tma_warpspecialized_cooperative.hpp
+++ b/include/cutlass/gemm/kernel/sm90_gemm_tma_warpspecialized_cooperative.hpp
@@ -804,7 +804,7 @@ public:
           // Update starting mainloop pipeline state for the next tile
           mainloop_pipe_consumer_state.advance(work_k_tile_count);
         }
-
+        #ifdef CUTLASS_ENABLE_GDC_FOR_SM90
         if (scheduler.is_last_tile(work_tile_info)) {
           // Hint on an early release of global memory resources.
           // The timing of calling this function only influences performance,
@@ -812,6 +812,7 @@ public:
           cutlass::arch::launch_dependent_grids();
 
         }
+        #endif
 
         // Index of warp group within consumer warp groups
         int consumer_warp_group_idx = canonical_warp_group_idx() - NumLoadWarpGroups;
--- a/include/cutlass/gemm/kernel/sm90_gemm_tma_warpspecialized_pingpong.hpp
+++ b/include/cutlass/gemm/kernel/sm90_gemm_tma_warpspecialized_pingpong.hpp
@@ -800,6 +800,7 @@ public:
     else if (warp_group_role == WarpGroupRole::Consumer0 || warp_group_role == WarpGroupRole::Consumer1) {
       cutlass::arch::warpgroup_reg_alloc<MmaRegisterRequirement>();
 
+      #ifdef CUTLASS_ENABLE_GDC_FOR_SM90
       // It is possible to have work tiles start off invalid,
       // so we have to check that first.
       if (not work_tile_info.is_valid()) {
@@ -810,6 +811,7 @@ public:
 
         return;
       }
+      #endif
       
       if constexpr (IsSchedDynamicPersistent) {
         // Consumer0's initial tile is static. It starts consuming the 2nd tile.
@@ -866,6 +868,7 @@ public:
         // Update starting mainloop pipeline state for the next tile
         mainloop_pipe_consumer_state.advance(k_tile_count * NumMmaWarpGroups);
 
+        #ifdef CUTLASS_ENABLE_GDC_FOR_SM90
         if (scheduler.is_last_tile(work_tile_info, NumMmaWarpGroups)) {
           // Hint on an early release of global memory resources.
           // The timing of calling this function only influences performance,
@@ -873,6 +876,7 @@ public:
           cutlass::arch::launch_dependent_grids();
 
         }
+        #endif
 
         // Order two Math WG's Epilogue one after the other
         math_wg_order_barrier.wait();


include/cutlass/gemm/kernel/sm90_gemm_array_tma_warpspecialized_cooperative.hpp
```diff
@@ -610,6 +610,9 @@ class GemmUniversal<
       if (producer_warp_role == ProducerWarpRole::Scheduler) {
         // GroupScheduler requires a producer warp to iterate over the group infos and push
         // the work tile infos to the downstream pipelines.
+        #ifdef CUTLASS_ENABLE_GDC_FOR_SM90
```
IonThruster: It might be safer / simpler to hoist these waits above the warp specialized region. Further optimization / pulling it into specialized regions - should be performance data driven - do you happen to have any details you can share ?. 

++ @ANIKET-SHIVAM  , @depaulmillz for review.

I see initial work-tile info & tensormap updates inside consumer regions not guarded / waiting for prior grid to complete, is that fine ?. I would have thought if problem shape is on device, before reading it - you'd need for the prior / dependent grid to complete ?

HydraQYH: Thank you for your reply. After analysis and testing, i hoist these waits above the warp specialized region. There are two main reasons for this:

1. Placing waits within or outside the WS region yields almost identical performance in my scenario.
2. I think your second point is correct, work-tile info & tensormap updates inside consumer regions should be guarded.

I rebase the code and adjust the position of the waits, and it's ready for review.

include/cutlass/gemm/kernel/sm90_gemm_array_tma_warpspecialized_pingpong.hpp
```diff
@@ -1034,6 +1037,11 @@ class GemmUniversal<
 
         // Get next work tile
         auto [next_work_tile_info, increment_pipe] = scheduler.fetch_next_work(work_tile_info, tile_scheduler_pipeline, tile_scheduler_pipe_consumer_state);
+        #ifdef CUTLASS_ENABLE_GDC_FOR_SM90
```
Algy: Nit: You don't have to wrap these stubs with `#ifdef - #endif`. `wait_on_dependent_grids()` and `launch_dependent_grids()` already do this for you.

HydraQYH: Thank you for pointing out the problem. I checked the code and found that this was indeed the case. I have removed unnecessary `#ifdef` and `#endif`.

include/cutlass/gemm/kernel/sm90_gemm_array_tma_warpspecialized_cooperative.hpp
```diff
@@ -990,6 +994,11 @@ class GemmUniversal<
 
         // Get next work tile
         auto [next_work_tile_info, increment_pipe] = scheduler.fetch_next_work(work_tile_info, tile_scheduler_pipeline, tile_scheduler_pipe_consumer_state);
+        #ifdef CUTLASS_ENABLE_GDC_FOR_SM90
```
d-k-b: Consider removing the `ifdef` since the function `launch_dependent_grids` internally checks this and the compiler should be able to remove the if statement if the function has an empty body on not supported.

HydraQYH: Thanks for the reminder. You are right, the compiler should be able to remove the if statement if the function has an empty body.

HydraQYH:
@Algy @d-k-b I've noticed that even the general gemm contains unnecessary macro definitions:
https://github.com/NVIDIA/cutlass/blob/a2439551c765c5393aebe557ee75d3a0412d2211/include/cutlass/gemm/kernel/sm90_gemm_tma_warpspecialized_pingpong.hpp#L871-L879
https://github.com/NVIDIA/cutlass/blob/a2439551c765c5393aebe557ee75d3a0412d2211/include/cutlass/gemm/kernel/sm90_gemm_tma_warpspecialized_cooperative.hpp#L793-L801

Even in TMA WS Pingpong GEMM, the early check will be skipped if `CUTLASS_ENABLE_GDC_FOR_SM90` is not defined:
https://github.com/NVIDIA/cutlass/blob/a2439551c765c5393aebe557ee75d3a0412d2211/include/cutlass/gemm/kernel/sm90_gemm_tma_warpspecialized_pingpong.hpp#L803-L814
Should we fix it?

d-k-b:
@hwu36 -- what are thoughts on removing `ifdef` checks from other locations?

hwu36:
They can be removed. May not be in this pr.

IonThruster:
We noticed there are more kernels that needs this fix - so we are pushing some basic / safe version into 4.3 release (coming out very soon). Would it be possible to wait for that and rebase this PR on top or see if that is enough ?

HydraQYH:
> We noticed there are more kernels that needs this fix - so we are pushing some basic / safe version into 4.3 release (coming out very soon). Would it be possible to wait for that and rebase this PR on top or see if that is enough ?

@IonThruster OK.

Algy:
@HydraQYH I've found out some more SM90 kernels need to be fixed in the same way. See the issue I posted before.

HydraQYH:
> @HydraQYH I've found out some more SM90 kernels need to be fixed in the same way. See the issue I posted before.

@Algy NVIDIA engineers have also discovered this issue and will fix it in version 4.3. cc @IonThruster @hwu36

IonThruster:
@Algy , @HydraQYH - could you check main branch if the changes look good (the changes are merged now), or if you'd like to rebase this PR ?

HydraQYH:
> @Algy , @HydraQYH - could you check main branch if the changes look good (the changes are merged now), or if you'd like to rebase this PR ?

@IonThruster It seems that the main branch only addresses the race condition raised in https://github.com/NVIDIA/cutlass/issues/2760.

There are still two problems that need to be solved:

1. Enable `cutlass::arch::launch_dependent_grids();` in array gemm.
2. Remove unnecessary `#ifdef` / `#endif`

I rebase code and fix the two issues mentioned above, and it's ready for review. cc @hwu36

Algy:
@HydraQYH @IonThruster  

I find the correctness issue fixed in the main branch, though no `cutlass::arch::launch_dependent_grids()` s are found in those kernels. I'm ok with the fix, but it appears @HydraQYH needs dep grid launch for the sake of performance, right?

HydraQYH:
> @HydraQYH @IonThruster
> 
> I find the correctness issue fixed in the main branch, though no `cutlass::arch::launch_dependent_grids()` s are found in those kernels. I'm ok with the fix, but it appears @HydraQYH needs dep grid launch for the sake of performance, right?

Yes!

IonThruster:
++ @Junkai-Wu , @hwu36

HydraQYH:
@hwu36 It looks like nvidia's engineers have approved this PR. Can you merge this?

include/cutlass/gemm/kernel/sm90_gemm_tma_warpspecialized_cooperative.hpp
```diff
@@ -804,15 +804,14 @@ class GemmUniversal<
           // Update starting mainloop pipeline state for the next tile
           mainloop_pipe_consumer_state.advance(work_k_tile_count);
         }
-        #ifdef CUTLASS_ENABLE_GDC_FOR_SM90
```
Junkai-Wu, line 807: This deletion will cause other kernels fail in build:
```
... has no member "is_last_tile"
```
It can't pass our internal pipeline. Is there other alternative approach?

HydraQYH: How can I reproduce this error?

HydraQYH:
@Junkai-Wu I reverted the deletion. I believe the error is caused by another bug—it seems a `TileScheduler` instance is missing `is_last_tile`, but without error messages, I can't determine which `TileScheduler` instance it is. This PR currently only adds PDL support to Array GEMM. Regarding the other bug, if you can provide more error information and a method to reproduce it, I can help fix it.

Junkai-Wu:
@HydraQYH Thanks for the quick action. You can reproduce it through running L0 functional test in profiler. Detailed commands are:
```
cmake .. -DCUTLASS_NVCC_ARCHS=120a -DCUTLASS_BUILD_FOR_PROFILER_REGRESSIONS=ON
make cutlass_profiler -j16
```
I saw sm120 kernels building failure in our internal pipeline. Not sure whether only sm120 kernels have this issue.

HydraQYH:
> @HydraQYH Thanks for the quick action. You can reproduce it through running L0 functional test in profiler. Detailed commands are:
> 
> ```
> cmake .. -DCUTLASS_NVCC_ARCHS=120a -DCUTLASS_BUILD_FOR_PROFILER_REGRESSIONS=ON
> make cutlass_profiler -j16
> ```
> 
> I saw sm120 kernels building failure in our internal pipeline. Not sure whether only sm120 kernels have this issue.

@Junkai-Wu Okay. I will try it later. I will be starting a new PR to track this issue. Could you please merge this first?

Junkai-Wu:
@HydraQYH I'm rerunning the internal pipeline after reverting the deletion. Will merge this PR after internal pipeline passes.

Junkai-Wu:
Seems internal pipeline is passing. I'll merge this PR.

HydraQYH:
@Junkai-Wu I discovered the cause of the bug. Please refer to: https://github.com/NVIDIA/cutlass/pull/2865

PR #2740, opened 2025-11-01
Fix register index bug in mma.sync.aligned.m16n8k16

neilkichler:
In `include/cutlass/arch/mma_sm90.h` the ptx instruction `mma.sync.aligned.m16n8k16` has a typo in the variable for `%5` which should be `A[1]` but is currently `A[2]` (and hence using `A[2]` twice and `A[1]` not at all).


neilkichler:
@hwu36 this should be a very quick review (single typo).

Junkai-Wu (approved):

PR #2741, opened 2025-11-01
Minor fix cute dsl example paths

Edenzzzz:


fengxie:
@brandon-yujie-sun have we fixed those pathes?

PR #2742, opened 2025-11-01
[Docs] Add the exact math definition of complement

Edenzzzz:
Helps understanding a bit, as the docs don't really say how complement is computed :)


ccecka:
I'll be releasing and referencing a whitepaper with correct and sufficiently general post conditions shortly.

Edenzzzz:
Sure, looking forward to it!

PR #2758, opened 2025-11-10
[cute-dsl][fix]remove cute dsl pdl example.

limin2021:
As `griddepcontrol` ops are available only from 13.0, so we remove the PDL example in cute dsl 4.3 release which is based on 12.9


limin2021:
Please help review the changes. @fengxie @brandon-yujie-sun

Junkai-Wu (approved):

PR #2789, opened 2025-11-21
v4.3 tag release update.

Junkai-Wu:


hwu36 (approved):

bbrowning:
Just a note that the tag was not actually pushed to GitHub - the latest shown here is still v4.2.1.

PR #2798, opened 2025-11-22
allow swizzled store for fill

drisspg:
Still trying to grok if this guard is needed when storing to a smem view if not just allowing that case feels better


fengxie:
I believe just removing the check is not enough. We need to update lower level compiler support. It's done internally but need to propagate to release. @brandon-yujie-sun for visibility and track.

PR #2803, opened 2025-11-23
DOC: Fixing fundamental types codeblock

SwayamInSync:
`cutlass/core_io.h` is required to use the IO overloads on `cutlass::half_t`

### DIFF
--- a/media/docs/cpp/fundamental_types.md
+++ b/media/docs/cpp/fundamental_types.md
@@ -224,7 +224,7 @@ int const kN = 16;
 Array<int8_t, kN> destination;
 Array<int,    kN> source;
 
-NumericConverter<descltype(destination), decltype(source)> convert;
+NumericArrayConverter<int8_t, int, kN> convert;
 
 destination = convert(source);
 ```


SwayamInSync:
Later in the same file demonstrating Array container for `int4b_t` it seems user-defined literals `_s4` isn't available even after including `subbyte_reference.h`

Is this expected? or those literals are removed, we might need to fix that as well

media/docs/cpp/fundamental_types.md
```diff
@@ -223,7 +224,7 @@ int const kN = 16;
 Array<int8_t, kN> destination;
 Array<int,    kN> source;
 
-NumericConverter<descltype(destination), decltype(source)> convert;
+NumericArrayConverter<int8_t, int, kN> convert;
```
SwayamInSync, line 227: Originally it was a typo here (`descltype`) but also can't use `decltype` as it expands to `const cutlass::NumericConverter<cutlass::Array<int8_t, 16, false>, cutlass::Array<int, 16, true>`

media/docs/cpp/fundamental_types.md
SwayamInSync, line 352: This seems to be the duplicate section (it exists inside Containers section) and the code block is empty as well. If this wasn't intended then we can remove this as well

PR #2818, opened 2025-11-27
v4.3.1 update

Junkai-Wu:


tridao:
awesome, should the PyPI wheel be updated as well?

Junkai-Wu:
@tridao Just marked the wheel as ready which will be updated soon on PyPI.

tqchen:
Wheel is now on PyPI

PR #2822, opened 2025-11-28
[CuTeDSL] Feature/fp8e4m3 to fp16 conversion

arseniivanov:
Support for type-converting to FP16 from FP8e4m3 tensors using .to(cute.Float16).

Very similar to existing FP4->FP16, but with chunks of 4 instead of 8 to fit a INT32 source.

Questions:

Got a test, but cannot find a test folder for CuteDSL.
I saw that the integer conversions had supported SM's listed at the bottom, but FP4 did not. Not sure if it should be added for this, perhaps different PR to add for both?

### DIFF
--- a/python/CuTeDSL/cutlass/cute/tensor.py
+++ b/python/CuTeDSL/cutlass/cute/tensor.py
@@ -14,7 +14,6 @@ from typing_extensions import deprecated
 from inspect import isclass
 import operator
 
-from cutlass.base_dsl.typing import Float8E4M3
 from cutlass.cutlass_dsl import (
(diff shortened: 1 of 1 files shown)


fengxie:
@brandon-yujie-sun

fengxie (approved):

arseniivanov:
It seems like this logic was moved from CuteDSL to nvgpu cvt MLIR arith in 4.4 v2(https://github.com/NVIDIA/cutlass/pull/2999), I have tested and things work with the existing main, this PR is redundant. Thank you for your time.

PR #2824, opened 2025-11-29
[DOCS] Update docs to precisely describe env stream scenario

tqchen:
Since the term synchronize may cause confusion that user thought it means stream sync, it simply means we pass the right current stream as env stream


fengxie:
@Junkai-Wu to merge

Issue #2845, opened 2025-12-05
[FEA] distributed_helper used to allow mem_order="acquire" in spin_lock_wait, but in 4.3.1 pip package only "relaxed" is exposed via


aleozlx:
### Which component requires the feature?

CuTe DSL

### Feature Request

**Is your feature request related to a problem? Please describe.**

distributed_helper used to allow mem_order="acquire" in spin_lock_wait, but in 4.3.1 pip package only "relaxed" is exposed via spin_lock_atom_cas_relaxed_wait. wish there is an "acquire" version exposed somehow as well

**Describe the solution you'd like**
one possibility : spin_lock_atom_cas_acquire_wait

**Describe alternatives you've considered**
keep mem_order string arg
may be some of our source code is ported from examples? i'm not sure, need to check ...

**Additional context**
used by https://github.com/aleozlx/flashinfer/blob/442dec9bea569f53e01b799a2e0328c2ea30bbca/flashinfer/cute_dsl/gemm_allreduce_two_shot.py#L1399-L1403
https://github.com/NVIDIA/cutlass/blob/v4.3.1/python/CuTeDSL/cutlass/utils/distributed_helpers.py#L136

XiaoSongXS:
close this issue, no acquire needed

XiaoSongXS:
if u want acquire, u should use relax spin-lock + acquire fence. this help avoid putting acquire related CCTL.IVAL in spin-lock while loop, which is important for perf.

PR #2846, opened 2025-12-05
Add spin_lock_atom_cas_acquire_wait function

aleozlx:
For https://github.com/NVIDIA/cutlass/issues/2845

Added spin_lock_atom_cas_acquire_wait function to handle spin lock acquisition with atomic compare-and-swap.

### DIFF
--- a/python/CuTeDSL/cutlass/utils/distributed.py
+++ b/python/CuTeDSL/cutlass/utils/distributed.py
@@ -266,6 +266,47 @@ def red_add1(
         elif order == "relaxed":
             red_relaxed_sys_add1(lock_ptr=lock_ptr, loc=loc, ip=ip)
 
+@cute.jit
+def spin_lock_atom_cas_acquire_wait(
+    lock_ptr: Pointer,
+    *,
+    expected_val: Int32,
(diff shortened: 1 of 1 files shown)


aleozlx:
This is functional. https://github.com/flashinfer-ai/flashinfer/pull/2171

Raising it as a proposed solution for what we needed when upgrading to nvidia-cutlass-dsl 4.3.1 https://github.com/NVIDIA/cutlass/issues/2845 

Kind regards from FlashInfer & cuDNN :)

XiaoSongXS:
acquire wait is not needed. slack Xiao Song and we can schedule a meeting to explain this

XiaoSongXS:
the two shot all redue.py fail is related to something else, let's discuss this in the meeting

shubaoyu2:
you can use the new two-shot gemm+ar kernel in cutedsl examples. The one in flashinfer should be an old version. 

adding something to CuTeDSL wheel package will take some time, so I would recommend you use the new kernel.

aleozlx:
sounds good will discuss with you over slack. will learn about the new kernel example and bring action item back to FI

PR #2865, opened 2025-12-09
[Bug Fix]Bypass launch grids for SM120 Kernel with SM90 Mainloop & SM100 TileScheduler

HydraQYH:
Follow: https://github.com/NVIDIA/cutlass/pull/2719. Some SM120 kernels will use SM90 [Cooperative](https://github.com/NVIDIA/cutlass/blob/main/include/cutlass/gemm/kernel/sm90_gemm_tma_warpspecialized_cooperative.hpp)/[Pingpong](https://github.com/NVIDIA/cutlass/blob/main/include/cutlass/gemm/kernel/sm90_gemm_tma_warpspecialized_pingpong.hpp) Mainloop with SM100 TileScheduler. SM100 TileScheduler do not have `is_last_tile` function. This can cause compilation errors when we remove unnecessary conditional compilation.
In this PR, If SM120 blockscaled is detected at compile time, then bypass `launch_dependent_grids`.

### DIFF
--- a/include/cutlass/gemm/kernel/sm90_gemm_tma_warpspecialized_pingpong.hpp
+++ b/include/cutlass/gemm/kernel/sm90_gemm_tma_warpspecialized_pingpong.hpp
@@ -802,9 +802,9 @@ public:
     else if (warp_group_role == WarpGroupRole::Consumer0 || warp_group_role == WarpGroupRole::Consumer1) {
       cutlass::arch::warpgroup_reg_alloc<MmaRegisterRequirement>();
 
-      // It is possible to have work tiles start off invalid,
-      // so we have to check that first.
       if constexpr (!IsSm120Family) {
+        // It is possible to have work tiles start off invalid,
+        // so we have to check that first.
         if (not work_tile_info.is_valid()) {
           // Hint on an early release of global memory resources.
           // The timing of calling this function only influences performance,


HydraQYH:
@Junkai-Wu I believe the current solution is only temporary, and in the future we should add `is_last_tile` to the SM100 TileScheduler. cc @hwu36 @IonThruster @d-k-b

include/cutlass/gemm/kernel/sm90_gemm_tma_warpspecialized_pingpong.hpp
```diff
@@ -147,6 +147,17 @@ class GemmUniversal<
   static constexpr uint32_t LoadRegisterRequirement = !HeavyRegisterPressure ? 40 : 24;
   static constexpr uint32_t MmaRegisterRequirement = !HeavyRegisterPressure ? 232 : 240;
 
+  // Detect if this is SM120 blockscaled kernel which hits high register pressure
+  // on smaller tiles (e.g. 256x128 registers per thread)
+  template <typename T>
+  struct is_blockscaled : cute::false_type {};
```
Junkai-Wu: The struct name here is not aligned with the coding style of cutlass. Please use camel case naming convention.

include/cutlass/gemm/kernel/sm90_gemm_tma_warpspecialized_pingpong.hpp
```diff
@@ -147,6 +147,17 @@ class GemmUniversal<
...
+
+  template <int Stages, int SchedStages, class ClusterShape, class KernelSchedule>
+  struct is_blockscaled<MainloopSm120TmaWarpSpecializedBlockScaled<Stages, SchedStages, ClusterShape, KernelSchedule>>
```
Junkai-Wu: I'm worried that `MainloopSm120TmaWarpSpecializedBlockScaled` is not the only dispatch policy that needs to be excluded in above `is_last_tile`. For example, there is another dispatch policy:`MainloopSm120ArrayTmaWarpSpecializedBlockScaled` which may have similar issue.

include/cutlass/gemm/kernel/sm90_gemm_tma_warpspecialized_pingpong.hpp
Junkai-Wu: Also the meaning of the struct name is not very clear. If it needs to bypass sm120 blockscaled kernel, it should be named at least `isSm120BlockScaled`.

Junkai-Wu:
> @Junkai-Wu I believe the current solution is only temporary, and in the future we should add `is_last_tile` to the SM100 TileScheduler. cc @hwu36 @IonThruster @d-k-b

Agreed. @HydraQYH Do you think if you can do the corresponding changes in this PR? I think just an empty const function with false return should be OK.

include/cutlass/gemm/kernel/sm90_gemm_tma_warpspecialized_pingpong.hpp
HydraQYH: This code is copied from: https://github.com/NVIDIA/cutlass/blob/49bd6bf1ba80abd588a56a3f3af2f1bcd41d215f/include/cutlass/gemm/kernel/sm90_gemm_tma_warpspecialized_cooperative.hpp#L136-L145
So the code of the main branch also has this typo, and I think you need to fix it internally. Also, it's not important, I have chosen another way to fix it.

include/cutlass/gemm/kernel/sm90_gemm_tma_warpspecialized_pingpong.hpp
HydraQYH: `MainloopSm120ArrayTmaWarpSpecializedBlockScaled` will use the following two Mainloops:

- https://github.com/NVIDIA/cutlass/blob/main/include/cutlass/gemm/kernel/sm90_gemm_array_tma_warpspecialized_cooperative.hpp
- https://github.com/NVIDIA/cutlass/blob/main/include/cutlass/gemm/kernel/sm90_gemm_array_tma_warpspecialized_pingpong.hpp

These two Mainloops use GroupScheduler and do not call `is_last_tile`. So it doesn't seem to have any risk. However, this is no longer important, I will fix it in another way.

HydraQYH:
> > @Junkai-Wu I believe the current solution is only temporary, and in the future we should add `is_last_tile` to the SM100 TileScheduler. cc @hwu36 @IonThruster @d-k-b
> 
> Agreed. @HydraQYH Do you think if you can do the corresponding changes in this PR? I think just an empty const function with false return should be OK.

@Junkai-Wu Thank you for your suggestion, I will do it as soon as possible.

HydraQYH:
@Junkai-Wu I rebase code and just add `is_last_tile` for SM100 TileScheduler. After testing with following command:
```
cmake .. -DCUTLASS_NVCC_ARCHS=120a -DCUTLASS_BUILD_FOR_PROFILER_REGRESSIONS=ON
make VERBOSE=1 cutlass_profiler -j16
```
There are no compilation errors anymore. And this PR is ready for review.

include/cutlass/gemm/kernel/sm90_gemm_tma_warpspecialized_pingpong.hpp
Junkai-Wu: Thanks for catching this. Will try to refactor it in future.

include/cutlass/gemm/kernel/sm100_tile_scheduler.hpp
```diff
@@ -449,6 +449,12 @@ class PersistentTileSchedulerSm100 {
     return clc_pipe_producer_state;
   }
 
+  CUTLASS_DEVICE
+  bool is_last_tile(WorkTileInfo& work_tile_info, uint32_t advance_count = 1) const {
+    // TODO: Complete this part
```
Junkai-Wu: Please remove this `TODO` comment. We are trying to avoid adding TODO comment in the public codes. Besides, there is nothing else to do in this function currently.

include/cutlass/gemm/kernel/sm100_tile_scheduler_stream_k.hpp
```diff
@@ -226,7 +226,13 @@ class PersistentTileSchedulerSm100StreamK {
   PipelineState<Stages> 
   advance_to_next_work(Pipeline& clc_pipeline, PipelineState<Stages> clc_pipe_producer_state) const {
     return sm100_scheduler_.advance_to_next_work(clc_pipeline, clc_pipe_producer_state);
- }
+  }
+
+  CUTLASS_DEVICE
+  bool is_last_tile(WorkTileInfo work_tile_info, uint32_t advance_count = 1) const {
+    // TODO: Complete this part
```
Junkai-Wu: Same as above.

Junkai-Wu:
@HydraQYH The changes look much cleaner now. Thanks for the quick action. Leave minor comments above.

include/cutlass/gemm/kernel/sm100_tile_scheduler.hpp
HydraQYH: Done

include/cutlass/gemm/kernel/sm100_tile_scheduler_stream_k.hpp
HydraQYH: Done

Junkai-Wu:
@HydraQYH I ran the internal pipeline with these changes and got a timeout issue of this unit test: https://github.com/NVIDIA/cutlass/blob/main/test/unit/gemm/device/sm120_tensorop_gemm/CMakeLists.txt#L51

The issue disappeared when I added the macro back on the `is_last_tile` function. I've asked corresponding developer to help identify the issue. You can also investigate it if possible.

HydraQYH:
@Junkai-Wu Sorry, I don't have an SM120 device, so all I can do is compile. However, I noticed that removing this macro in Pingpong causes the program to **return immediately** sometime: https://github.com/NVIDIA/cutlass/blob/d4e16f5d4e70cd95049e3708cbee01205abe43c0/include/cutlass/gemm/kernel/sm90_gemm_tma_warpspecialized_pingpong.hpp#L803-L814
I think this radical approach may be risky, and may result in some semaphores not being released yet. In contrast, other changes will not affect the program's execution. I've reverted the changes to the early stop part. Could you try this?

Junkai-Wu:
@HydraQYH After investigation, the timeout issue is caused by sm120 kernel calling sm90 scheduler where sm120 kernel didn't need to call `is_last_tile` before and after the removal of the macro, it will call this function which may cause some schedule issue. Maybe revert to previously implementation which using `IsBlockScaled` but with a proper name would be a safe option.

HydraQYH:
@Junkai-Wu Okay. I rebaseed the code and reverted to previously implementation. I changed `is_blockscaled` to `IsBlockScaledDispatchPolicy`, which I believe is reasonable because the template parameter is just a `DispatchPolicy`. I also fixed the typo in [Cooperative](https://github.com/NVIDIA/cutlass/blob/main/include/cutlass/gemm/kernel/sm90_gemm_tma_warpspecialized_cooperative.hpp). And it's ready for review.

Junkai-Wu:
@HydraQYH The internal pipeline still fails. After checking, I found all sm120 kernels should not call the `is_last_tile` function, not just sm120 blockscaled kernels. I refactored the change and reran the internal pipeline. I'll leave comments in this PR after the pipeline passes.

include/cutlass/gemm/kernel/sm90_gemm_tma_warpspecialized_cooperative.hpp
```diff
@@ -136,13 +136,13 @@ class GemmUniversal<
   // Detect if this is SM120 blockscaled kernel which hits high register pressure
   // on smaller tiles (e.g. 256x128 registers per thread)
   template <typename T>
-  struct is_blockscaled : cute::false_type {};
+  struct IsBlockScaledDispatchPolicy : cute::false_type {};
```
Junkai-Wu:
```suggestion
  struct IsSm120BlockScaled : cute::false_type {};
```

include/cutlass/gemm/kernel/sm90_gemm_tma_warpspecialized_cooperative.hpp
```diff
@@ -136,13 +136,13 @@ class GemmUniversal<
...
   
   template <int Stages, int SchedStages, class ClusterShape, class KernelSchedule>
-  struct is_blockscaled<MainloopSm120TmaWarpSpecializedBlockScaled<Stages, SchedStages, ClusterShape, KernelSchedule>> 
+  struct IsBlockScaledDispatchPolicy<MainloopSm120TmaWarpSpecializedBlockScaled<Stages, SchedStages, ClusterShape, KernelSchedule>>
```
Junkai-Wu:
```suggestion
  struct IsSm120BlockScaled<MainloopSm120TmaWarpSpecializedBlockScaled<Stages, SchedStages, ClusterShape, KernelSchedule>> 
```

include/cutlass/gemm/kernel/sm90_gemm_tma_warpspecialized_cooperative.hpp
```diff
@@ -136,13 +136,13 @@ class GemmUniversal<
...
-  struct is_blockscaled : cute::false_type {};
+  struct IsBlockScaledDispatchPolicy : cute::false_type {};
   
   template <int Stages, int SchedStages, class ClusterShape, class KernelSchedule>
-  struct is_blockscaled<MainloopSm120TmaWarpSpecializedBlockScaled<Stages, SchedStages, ClusterShape, KernelSchedule>> 
+  struct IsBlockScaledDispatchPolicy<MainloopSm120TmaWarpSpecializedBlockScaled<Stages, SchedStages, ClusterShape, KernelSchedule>> 
     : cute::true_type {};
   
-  static constexpr bool IsBlockScaled = is_blockscaled<DispatchPolicy>::value;
+  static constexpr bool IsBlockScaled = IsBlockScaledDispatchPolicy<DispatchPolicy>::value;
```
Junkai-Wu:
```suggestion
  static constexpr bool IsSm120Family = cute::is_same_v<typename DispatchPolicy::ArchTag, arch::Sm120>;
```

include/cutlass/gemm/kernel/sm90_gemm_tma_warpspecialized_cooperative.hpp
```diff
@@ -136,13 +136,13 @@ class GemmUniversal<
...
   
   static constexpr bool HeavyRegisterPressure = 
     IsBlockScaled ? (RegsPerThread >= 128) : (RegsPerThread >= 208);
```
Junkai-Wu:
```suggestion
    IsSm120BlockScaled<DispatchPolicy>::value ? (RegsPerThread >= 128) : (RegsPerThread >= 208);
```

include/cutlass/gemm/kernel/sm90_gemm_tma_warpspecialized_cooperative.hpp
```diff
@@ -804,15 +804,16 @@ class GemmUniversal<
...
           mainloop_pipe_consumer_state.advance(work_k_tile_count);
         }
-        #ifdef CUTLASS_ENABLE_GDC_FOR_SM90
-        if (scheduler.is_last_tile(work_tile_info)) {
-          // Hint on an early release of global memory resources.
-          // The timing of calling this function only influences performance,
-          // not functional correctness.
-          cutlass::arch::launch_dependent_grids();
 
+        if constexpr (!IsBlockScaled) {
```
Junkai-Wu:
```suggestion
        if constexpr (!IsSm120Family) {
```

include/cutlass/gemm/kernel/sm90_gemm_tma_warpspecialized_pingpong.hpp
```diff
@@ -147,6 +147,17 @@ class GemmUniversal<
...
+  // Detect if this is SM120 blockscaled kernel which hits high register pressure
+  // on smaller tiles (e.g. 256x128 registers per thread)
+  template <typename T>
+  struct IsBlockScaledDispatchPolicy : cute::false_type {};
+
+  template <int Stages, int SchedStages, class ClusterShape, class KernelSchedule>
+  struct IsBlockScaledDispatchPolicy<MainloopSm120TmaWarpSpecializedBlockScaled<Stages, SchedStages, ClusterShape, KernelSchedule>> 
+    : cute::true_type {};
+
+  static constexpr bool IsBlockScaled = IsBlockScaledDispatchPolicy<DispatchPolicy>::value;
```
Junkai-Wu:
```suggestion
  static constexpr bool IsSm120Family = cute::is_same_v<typename DispatchPolicy::ArchTag, arch::Sm120>;
```

include/cutlass/gemm/kernel/sm90_gemm_tma_warpspecialized_pingpong.hpp
```diff
@@ -800,7 +811,6 @@ class GemmUniversal<
     else if (warp_group_role == WarpGroupRole::Consumer0 || warp_group_role == WarpGroupRole::Consumer1) {
       cutlass::arch::warpgroup_reg_alloc<MmaRegisterRequirement>();
 
-      #ifdef CUTLASS_ENABLE_GDC_FOR_SM90
```
Junkai-Wu, line 803: Add `if constexpr (!IsSm120Family)` condition here.

include/cutlass/gemm/kernel/sm90_gemm_tma_warpspecialized_pingpong.hpp
```diff
@@ -868,15 +877,15 @@ class GemmUniversal<
         // Update starting mainloop pipeline state for the next tile
         mainloop_pipe_consumer_state.advance(k_tile_count * NumMmaWarpGroups);
 
-        #ifdef CUTLASS_ENABLE_GDC_FOR_SM90
-        if (scheduler.is_last_tile(work_tile_info, NumMmaWarpGroups)) {
-          // Hint on an early release of global memory resources.
-          // The timing of calling this function only influences performance,
-          // not functional correctness.
-          cutlass::arch::launch_dependent_grids();
+        if constexpr (!IsBlockScaled) {
```
Junkai-Wu:
```suggestion
        if constexpr (!IsSm120Family) {
```

Junkai-Wu:
@HydraQYH The internal pipeline passes. Leave suggested implementations in the comments.

HydraQYH:
@Junkai-Wu Thank you very much for your help. I have picked all your suggestions.

include/cutlass/gemm/kernel/sm90_gemm_tma_warpspecialized_pingpong.hpp
HydraQYH: Done.

Junkai-Wu (approved):

PR #2870, opened 2025-12-10
docs: Add FP16 GEMM documentation to sgemm_sm80.cu - Fixes #1686

### LINKED ISSUE #1686: [QST] Is there any fp16xfp16 GEMM sample using CUTE with a performance comparable to cublas?
**What is your question?**
I want to write my own fused fp16xfp16 gemm kernel with CUTE, but I can not find a tutorial/sample code with a performance comparable to cublas. 

I noticed there are some tutorials in https://github.com/NVIDIA/cutlass/tree/main/examples/cute/tutorial, which has fp32xfp32 and int8xint8 gemm. But the performance of int8xint8 gemm is not good enough. I also noticed a 3rd party of fp16xfp16 gemm with CUTE https://github.com/leimao/CUDA-GEMM-Optimization?tab=readme-ov-file, but as shown in the readme, the performance is yet not comparable to cublas. So I wonder whether CUTE can give an official fp16xfp16 gemm kernel with good performance, so that I can develop based on that?

blueberrycongee:
## Summary
Add descriptive documentation to [sgemm_sm80.cu](cci:7://file:///d:/Desktop/cutlass-1/examples/cute/tutorial/sgemm_sm80.cu:0:0-0:0) explaining that it is actually an FP16xFP16 GEMM (HGEMM) tutorial using CuTe.

## Problem
Issue #1686 reports that users cannot find an fp16 GEMM tutorial in the CuTe examples.

## Solution
The existing [sgemm_sm80.cu](cci:7://file:///d:/Desktop/cutlass-1/examples/cute/tutorial/sgemm_sm80.cu:0:0-0:0) already implements FP16 GEMM using `cute::half_t`, but this was not documented. This PR adds a documentation block clarifying:
- This example uses FP16 data types despite the "sgemm" filename
- Key features: Tensor Cores, cp.async, pipelining, swizzled shared memory
- Usage examples

Fixes #1686

### DIFF
--- a/examples/cute/tutorial/sgemm_sm80.cu
+++ b/examples/cute/tutorial/sgemm_sm80.cu
@@ -28,6 +28,28 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  *
  **************************************************************************************************/
+
+/*
+  This example demonstrates a high-performance FP16xFP16 GEMM (HGEMM) using CuTe
+  for SM80 (Ampere) and later architectures.
+
+  Key features:
+  - Half-precision (FP16) inputs and outputs using cute::half_t
+  - Tensor Core operations via SM80_16x8x16_F16F16F16F16_TN MMA atom
+  - Asynchronous global-to-shared memory copy using cp.async (SM80_CP_ASYNC)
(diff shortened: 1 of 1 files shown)


hwu36:
if the accumulation type is fp32, it is sgemm.  if the accumulation type is fp16, it is hgemm.

PR #2881, opened 2025-12-16
new example with TMA prefetch feature targeting for DRAM latency boun…

questa-wang:
The newly added examples enable TMA prefetch during TMA load issue to increase the number of inflight TMA requests and hide latency, resulting in significant performance speedups up to 1.12x for cases limited by DRAM latency.

### DIFF
--- a/examples/python/CuTeDSL/blackwell/dense_blockscaled_gemm_persistent_prefetch.py
+++ b/examples/python/CuTeDSL/blackwell/dense_blockscaled_gemm_persistent_prefetch.py
@@ -1,0 +1,2657 @@
+# Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: BSD-3-Clause
+
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+
+# 1. Redistributions of source code must retain the above copyright notice, this
+# list of conditions and the following disclaimer.
+
+# 2. Redistributions in binary form must reproduce the above copyright notice,
+# this list of conditions and the following disclaimer in the documentation
+# and/or other materials provided with the distribution.
+
+# 3. Neither the name of the copyright holder nor the names of its
+# contributors may be used to endorse or promote products derived from
+# this software without specific prior written permission.
+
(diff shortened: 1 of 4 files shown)


examples/python/CuTeDSL/blackwell/dense_blockscaled_gemm_persistent_prefetch.py
```diff
@@ -0,0 +1,2657 @@
...
+        @cute.struct
+        class SharedStorage:
+            ab_full_mbar_ptr: cute.struct.MemRange[cutlass.Int64, self.num_ab_stage]
+            ab_empty_mbar_ptr: cute.struct.MemRange[cutlass.Int64, self.num_ab_stage]
+            acc_full_mbar_ptr: cute.struct.MemRange[cutlass.Int64, self.num_acc_stage]
+            acc_empty_mbar_ptr: cute.struct.MemRange[cutlass.Int64, self.num_acc_stage]
+            tmem_dealloc_mbar_ptr: cutlass.Int64
+            tmem_holding_buf: cutlass.Int32
+            # (EPI_TILE_M, EPI_TILE_N, STAGE)
+            sC: cute.struct.Align[
```
fengxie, line 646: I think we can use `get_tensor` from `SmemAllocator` to simplify this part?

fengxie (approved):

examples/python/CuTeDSL/blackwell/dense_blockscaled_gemm_persistent_prefetch.py
questa-wang: Those newly added examples are copied from the existing ones, but with TMA loading changes applied before the mainloop and inside the mainloop.
.
```
dense_gemm_persistent.py
dense_blockscaled_gemm_persistent.py
```
For the epilogue part, the current code keeps the same style as in dense_blockscaled_gemm_persistent.py([code](https://github.com/NVIDIA/cutlass/blob/main/examples/python/CuTeDSL/blackwell/dense_blockscaled_gemm_persistent.py#L589-#592)). Since dense_blockscaled_gemm_persistent.py has not been updated yet, it might be better to start by refactoring that original example first. We can discuss opening a separate PR to gradually apply the same changes, including to the prefetch version later on. Does this make sense to you?

PR #2890, opened 2025-12-19
Fix finding cuDNN

### LINKED ISSUE #2885: [BUG] cuDNN not found (but it is installed)

CUTLASS C++

### Bug Report

**Describe the bug**
When configuring cutlass with CMake, I get the error `cuDNN not found`. However, I do have cuDNN installed, using the [installer from NVIDIA](https://developer.nvidia.com/cudnn-downloads?target_os=Windows&target_arch=x86_64&target_version=11). When installed that way, *cudnn is NOT embedded inside CUDA directories*.

1. Install CUDNN via [the official installer](https://developer.nvidia.com/cudnn-downloads?target_os=Windows&target_arch=x86_64&target_version=11). As a result, the include directory is `C:\Program Files\NVIDIA\CUDNN\v9.13\include\12.9` (for CUDA 12.9) and `C:\Program Files\NVIDIA\CUDNN\v9.13\include\13.0` (for CUDA 13.0). **Please note that these paths do not end with `/include`**
2. In this repo, launch cmake: `cmake -B build -DCUTLASS_ENABLE_CUDNN=ON .`
3. I get the error `cuDNN not found.`

**Expected behavior**
I expect CMake configure to go well, and CUDNN to be correctly found.

 - Regular Win11 PC

**Additional context**
It seems like [`cuDNN.cmake`](https://github.com/NVIDIA/cutlass/blob/d4e16f5d4e70cd95049e3708cbee01205abe43c0/cuDNN.cmake#L45) does a `find_path` but force `cudnn.h` to be under a path ending with `/include`. This is just not the case with the official installer...

I can propose a patch, if you want - mainly adding naked `$ENV{CUDNN_PATH}` in `find_path`/`find_library`

TLescoatTFX:
Fix #2885

This PR is exactly [the same patch as on vcpkg](https://github.com/microsoft/vcpkg/pull/48953)

### DIFF
--- a/cuDNN.cmake
+++ b/cuDNN.cmake
@@ -39,16 +39,17 @@ message(STATUS "Configuring cuDNN ...")
 find_path(
     _CUDNN_INCLUDE_DIR cudnn.h
     PATHS
+    $ENV{CUDNN_PATH}
     ${CUDA_TOOLKIT_ROOT_DIR}/include
     $ENV{CUDNN_PATH}/include
     $ENV{CUDA_PATH}/include
     ${CUDNN_PATH}/include
-    /usr/include
-    $ENV{CUDNN_PATH})
+    /usr/include)
 
 find_library(
     _CUDNN_LIBRARY cudnn
     HINTS
+    $ENV{CUDNN_PATH}
     ${CUDA_TOOLKIT_ROOT_DIR}/lib64
     ${CUDA_TOOLKIT_ROOT_DIR}/lib/x64
     ${CUDA_TOOLKIT_ROOT_DIR}/lib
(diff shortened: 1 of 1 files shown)


hwu36:
@d-k-b

cuDNN.cmake
```diff
@@ -43,7 +43,8 @@ find_path(
     $ENV{CUDNN_PATH}/include
     $ENV{CUDA_PATH}/include
     ${CUDNN_PATH}/include
-    /usr/include)
+    /usr/include
```
d-k-b: Overall this seems reasonable. However, I would move this to be the first location searched instead of the last since we want env variables to have precedence.

TLescoatTFX: Thanks for the comment, I updated the PR to search `$ENV{CUDNN_PATH}` first

TLescoatTFX:
@d-k-b @hwu36 This patch would still help a lot, it would be greatly appreciated if you could review/accept it. Please tell me if you want me to rebase it. Thanks !

d-k-b:
@hwu36 -- changes look good.

PR #2891, opened 2025-12-20
docs: note when DSL dumps are populated

ColinPeppler:
Added a disclaimer that __ptx__, __cubin__, and __mlir__ attributes are only populated when their respective CUTE_DSL_KEEP_* envvar are set.

```
print(f"PTX: {compiled_foo.__ptx__}")    # Otherwise, this prints None without CUTE_DSL_KEEP_PTX=1
```

### DIFF
--- a/media/docs/pythonDSL/cute_dsl_general/debugging.rst
+++ b/media/docs/pythonDSL/cute_dsl_general/debugging.rst
@@ -112,6 +112,9 @@ For compiled kernels, the generated PTX/CUBIN/IR can be accessed programmaticall
 - ``__cubin__``: The generated CUBIN data of the compiled kernel.
 - ``__mlir__``: The generated IR code of the compiled kernel.
 
+These attributes are populated only when the corresponding ``CUTE_DSL_KEEP_*`` environment variable is enabled;
+otherwise they return ``None``.
+
 .. code:: python
     
     compiled_foo = cute.compile(foo, ...)


ColinPeppler:
Hi @Junkai-Wu @hwu36, I was initially confused why `compiled.__ptx__` wasn't showing anything. I'm wondering if we're required to set `CUTE_DSL_KEEP_PTX=1` for `__ptx__` to be populated.

media/docs/pythonDSL/cute_dsl_general/debugging.rst
```diff
@@ -112,6 +112,9 @@ For compiled kernels, the generated PTX/CUBIN/IR can be accessed programmaticall
 - ``__cubin__``: The generated CUBIN data of the compiled kernel.
 - ``__mlir__``: The generated IR code of the compiled kernel.
 
+These attributes are populated only when the corresponding ``CUTE_DSL_KEEP_*`` environment variable is enabled;
```
brandon-yujie-sun, line 115: Yes, these attributes are meaningful only when the corresponding dumping is enabled, either by setting ENV VAR or through compile options for cute.compile, see https://github.com/NVIDIA/cutlass/blob/main/media/docs/pythonDSL/cute_dsl_general/dsl_jit_compilation_options.rst

Junkai-Wu:
@fengxie @brandon-yujie-sun This change is not in internal yet. We should port it to internal before merging this PR.

PR #2916, opened 2025-12-31
feat(examples/test_run): use runtime sm arch

tpoisonooo:
## Issue Summary
Example `13_two_tensor_op_fusion` uses hardcoded SM architecture, causing confusion and extra work for new comers.

## Background
I'm using H800 GPU (SM90) and initially misunderstood the SM compatibility (perhaps due to TensorRT 1.0 and cudnn7). 

To run the `13_two_tensor_op_fusion` example, I have spent hours reading template source code, write new version and fix compile error for SM90 (like [these code](https://github.com/tpoisonooo/cutlass/blob/h800-compatible/examples/13_two_tensor_op_fusion/fused_two_gemms_f16_sm90_rf.cu)) .

## Proposal
Modify `testRun` to use actual SM architecture of the current GPU, rather than using a hardcoded value.

## Benefits
- **Improves onboarding experience** by eliminating manual code modification for new users 
- **Avoids confusion** about SM compatibility requirements

### DIFF
--- a/examples/45_dual_gemm/test_run.h
+++ b/examples/45_dual_gemm/test_run.h
@@ -92,3 +92,4 @@ int testRun(int arch, std::vector<bool (*)()> & test_funcs, const std::string & 
     return -1;
 
 }
+
--- a/examples/45_dual_gemm/test_run.h
+++ b/examples/45_dual_gemm/test_run.h
@@ -31,16 +31,9 @@
 
 
 #include <iostream>
-#include <cuda_runtime_api.h>
-
-int getDefaultSmArch() {
-    cudaSetDevice(0);
-    cudaDeviceProp prop;
-    cudaGetDeviceProperties(&prop, 0);
-    return prop.major * 10 + prop.minor;
-}
 
 // Run tests on GPUs 
+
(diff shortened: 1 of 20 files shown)


examples/45_dual_gemm/test_run.h
```diff
@@ -31,9 +31,16 @@
 
 
 #include <iostream>
+#include <cuda_runtime_api.h>
 
-// Run tests on GPUs
```
hwu36, line 35: how is this used in this example?

tpoisonooo: reverted for `examples/45_dual_gemm`

tpoisonooo:
Now the fix only applies to `examples/13_two_tensor_op_fusion`.
I tested it with

```bash
cd build/examples/13_two_tensor_op_fusion

for f in 13_fused_*; do [ -f "$f" ] && ./"$f" >> run.out; done

# manually check the output file
```

@hwu36

hwu36:
@jwang323

tpoisonooo:
@jwu323 please review.

PR #2924, opened 2026-01-03
[doc] N / scale N granularity typo?

aidando73:
- Pretty sure this should be N / scale granularity N right? cc @depaulmillz


depaulmillz:
That is right. Thanks for catching this.

hwu36 (approved):

PR #2940, opened 2026-01-09
[docs] Add additional tip for generating less kernels in blockwise

aidando73:
- Running without this generates a lot of kernels - takes a long time


aidando73:
cc @depaulmillz @hwu36

depaulmillz:
Thanks

PR #2965, opened 2026-01-19
[Bug Fix]Set NumSplitsM to 1 when TileShapeM < 128 in sm90 fp8 blockwise scaling CollectiveMma

HydraQYH:
Simple modifications to [Example 68](https://github.com/NVIDIA/cutlass/blob/main/examples/68_hopper_fp8_warp_specialized_grouped_gemm_with_blockwise_scaling/68_hopper_fp8_warp_specialized_grouped_gemm_with_blockwise_scaling.cu):
```
using TileShape     = Shape<_64,_128,_128>;                        // Threadblock-level tile size
// ...
using KernelSchedule    = cutlass::gemm::KernelPtrArrayTmaWarpSpecializedPingpongFP8Blockwise;
using EpilogueSchedule  = cutlass::epilogue::PtrArrayTmaWarpSpecializedPingpong;
```
This configuration is commonly used with H20 GPUs, but it triggered a compilation error since v4.3:
CUTLASS Bug
The reason is that when `get<0>(TileShape_{})<_128{}`, `NumSplitsM` is set to `_0`, causing the static assert to fail: https://github.com/NVIDIA/cutlass/blob/8debf77437753beca676eb3c6bf97b56a5f9fd68/include/cutlass/gemm/collective/sm90_mma_array_tma_gmma_ss_warpspecialized_fp8_blockwise_scaling.hpp#L215-L216
In this case, we should not split M, therefore `NumSplitsM` should be `_1`.

### DIFF
--- a/include/cutlass/gemm/collective/sm90_mma_array_tma_gmma_ss_warpspecialized_fp8_blockwise_scaling.hpp
+++ b/include/cutlass/gemm/collective/sm90_mma_array_tma_gmma_ss_warpspecialized_fp8_blockwise_scaling.hpp
@@ -213,7 +213,7 @@ struct CollectiveMma<
   static_assert(cute::is_same_v<ElementAccumulator, ElementBlockScale>,
              "ElementAccumulator and ElementBlockScale should be same datatype");
   // For TileShapeM < 128, NumSplitsM should be 1
-  using NumSplitsM = std::conditional_t<get<0>(TileShape_{}) < _128{}, _1, cute::C<get<0>(TileShape_{}) / 128>>;
+  using NumSplitsM = cute::conditional_t<get<0>(TileShape_{}) < _128{}, _1, cute::C<get<0>(TileShape_{}) / 128>>;
   static_assert(NumSplitsM{} == 1 || NumSplitsM{} == 2);
 
   struct SharedStorage {


HydraQYH:
@hwu36 @Junkai-Wu Please take a look, thanks. cc @johnnynunez

Junkai-Wu:
Make sense to me. Thanks for the fix.

Junkai-Wu (approved):

include/cutlass/gemm/collective/sm90_mma_array_tma_gmma_ss_warpspecialized_fp8_blockwise_scaling.hpp
```diff
@@ -212,7 +212,8 @@ struct CollectiveMma<
 
   static_assert(cute::is_same_v<ElementAccumulator, ElementBlockScale>,
              "ElementAccumulator and ElementBlockScale should be same datatype");
-  using NumSplitsM = cute::C<get<0>(TileShape_{}) / 128>;
+  // For TileShapeM < 128, NumSplitsM should be 1
+  using NumSplitsM = std::conditional_t<get<0>(TileShape_{}) < _128{}, _1, cute::C<get<0>(TileShape_{}) / 128>>;
```
Junkai-Wu: Please use `cute::conditional_t` here since using `std::` here will break one of our internal pipelines.

HydraQYH: Done.

PR #2969, opened 2026-01-21
[NVVM API] Update cutedsl nvvm api change 

XiaoSongXS:


shubaoyu2:
LGTM，we shouldn't allow direct nvvm api calls in example as they are unstable

brandon-yujie-sun:
Thanks for the proposed change. LGTM,  +@fengxie

shubaoyu2:
just a reminder, we should merge it after 4.4 release, now there is no cute.arch.load in public wheel

PR #3008, opened 2026-02-06
use compiler macro to imporve the compatibility

reed-lau:
Switching from the CMake-defined CUDA_VERSION macro to the compiler-provided one ensures correct behavior for all users, including those who include the headers directly, and improves overall compatibility.


reed-lau:
@hwu36 could you help review it?

reed-lau:
@thakkarV could you help review it?

PR #3022, opened 2026-02-11
[Cute-DSL] Add cute.arch.fmin by calling nvvm

tridao:
There's already cute.arch.fmax but no fmin yet


fengxie:
@brandon-yujie-sun to track

PR #3053, opened 2026-02-22
Use unrounded inputs for the profiler by default

saagarjha:
This behavior leads to confusing performance numbers on modern hardware; one example is https://hazyresearch.stanford.edu/blog/2026-02-19-tk-2. Based on a conversation with @thakkarV this seems unintended so this changes the default to be more representative.


saagarjha:
Nothing has changed with regards to how relevant this PR is.

saagarjha:
There are no updates, it's just that nobody is reviewing it

PR #3055, opened 2026-02-23
Replace std::min with cute::min in sm120 blockwise scaling device functions

### LINKED ISSUE #3051: [BUG] std::min (host only) is called from device function

CUTLASS C++

### Bug Report

https://github.com/NVIDIA/cutlass/blob/main/include/cutlass/gemm/collective/sm120_mma_tma_blockwise_scaling.hpp#L475

The std::min is not device tagged and cannot be called from this host function. You'll see that it breaks with nvcc when building examples/87_blackwell_geforce_gemm_blockwise. It only passes when expressing --expt-relaxed-constexpr, which is kind of a quirk, because this is not a constexpr initializer.

vruga:
Fixes #3051.

 ## Fix

  Replaced `std::min` with `cute::min`, which is tagged `CUTE_HOST_DEVICE constexpr`
  and is safe to call from device code. It accepts plain arithmetic types (both
  arguments here are runtime integers), and is already the standard choice for this
  pattern across the rest of the CUTLASS collective headers.

  The same bug was also present in the `load()` function of the sister file
  `sm120_mma_array_tma_blockwise_scaling.hpp`, so that has been fixed here as well.

### DIFF
--- a/include/cutlass/gemm/collective/sm120_mma_array_tma_blockwise_scaling.hpp
+++ b/include/cutlass/gemm/collective/sm120_mma_array_tma_blockwise_scaling.hpp
@@ -540,8 +540,8 @@ struct CollectiveMma<
       Tensor tApA_SFA = make_tensor<bool>(shape(tAsA_SFA(_,_,0)));
       Tensor tBpB_SFB = make_tensor<bool>(shape(tBsB_SFB(_,_,0)));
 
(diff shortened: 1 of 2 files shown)


johnnynunez:
cc @depaulmillz

hwu36 (approved):

PR #3059, opened 2026-02-23
[CuTeDSL] Add BF16 Grouped GEMM example for Hopper (SM90)

vruga:
## Summary

Adds `examples/python/CuTeDSL/hopper/grouped_gemm.py`, a Python/CuTeDSL grouped GEMM kernel for the NVIDIA Hopper SM90 architecture with BF16 (and Float16) input support.

This is the CuTeDSL equivalent of `examples/57_hopper_grouped_gemm`, which:
- exists only in C++, and
- only supports FP8.

This PR adds a Python implementation that also covers the BF16 use case — the precision most commonly required for accuracy-sensitive workloads like Mixture-of-Experts (MoE) serving on Hopper.


jiuzhengWang:
what happended to this PR? is there anything wrong with the implementation?

vruga:
i put the same pr after verifying

PR #3075, opened 2026-02-26
[Cute-DSL] Fix runtime shared library loading for TVM FFI Module Loading

Alkaid-Benetnash:
When loading a pre-compiled CuTe DSL shared library via `cute.runtime.load_module()` with TVM FFI, the dynamic linker fails to resolve `_cudaLibraryLoadData`:

```
m = cute.runtime.load_module(path, enable_tvm_ffi=True)

#Stack trace:
cutlass/cute/runtime.py:966: in load_module
    return tvm_ffi.load_module(file_path, keep_module_alive=False)
tvm_ffi/module.py:472: in load_module
    mod = _ffi_api.ModuleLoadFromFile(path)
tvm_ffi/cython/function.pxi:923: in tvm_ffi.core.Function.__call__
    ...
# Error: _cudaLibraryLoadData not defined
```

`libcute_dsl_runtime.so` exports `_cudaLibraryLoadData` and other CUDA Runtime API wrappers that the MLIR-generated code in compiled `.so` files references. The library should be loaded with `RTLD_GLOBAL` so its symbols are visible to subsequent `dlopen()` calls.

`ctypes.CDLL(path)` defaults to `mode=0`, which is `RTLD_LOCAL | RTLD_LAZY` on Linux. `RTLD_LOCAL` keeps the loaded symbols **private**.

Without this PR, the workaround would be "always run cute.compile once before running pre-compiled binaries".

### DIFF
--- a/python/CuTeDSL/cutlass/cute/runtime.py
+++ b/python/CuTeDSL/cutlass/cute/runtime.py
@@ -951,7 +951,7 @@ def load_module(file_path: str, *, enable_tvm_ffi: bool = False):
         # no need to load tvm_ffi library here since it will be loaded by tvm_ffi package.
         for path in find_runtime_libraries(enable_tvm_ffi=False):
             if Path(path).exists():
-                _LOAD_MODULE_LIBS_CACHE.append(ctypes.CDLL(path))
+                _LOAD_MODULE_LIBS_CACHE.append(ctypes.CDLL(path, mode=ctypes.RTLD_GLOBAL))
 
     if enable_tvm_ffi:
         import tvm_ffi


Alkaid-Benetnash:
cc @tqchen  ?

tqchen:
Thanks @Alkaid-Benetnash seems this is a requirement for related symbols being available in the global space. this is a reasonable change. alternatively i think we could have the .o link to libcute_dsl_runtime.so explicitly in linking step, which should get this module

There should some some further updates that already loads the cute.runtime.load_module to use cuteDSL's binaryeExecutor which might resolve the issue, needs to confirm if it lands in release cc @fengxie

fengxie:
@brandon-yujie-sun can you confirm this will be included in latest release?

brandon-yujie-sun:
We have been using BinaryExecutionEngine and this is already fixed. Closing

Issue #3086, opened 2026-03-01
[BUG] Boolean.__dsl_and__ emits unnecessary i1→i32→i1 round-trip instead of arith.andi for two Boolean operands


Johnsonms:
CuTe DSL

### Bug Report

**Describe the bug**
  In CuTe DSL, when Python and is used to combine two Boolean (i1) values in a while
  condition, the AST preprocessor rewrites it through __dsl_and__, which promotes both
   operands from i1 to i32 via arith.extui, performs an arith.select, and converts
  back to i1 via arith.cmpi ne. This emits 6 unnecessary MLIR operations for what
  should be a single arith.andi %a, %b : i1.

  The bitwise & operator calls __and__ directly and already emits the correct
  single-op IR. Both operators produce semantically equivalent i1 results, but and
  generates verbose, unnecessary IR that is harder to read and verify.

  This affects any DSL while loop that combines two Boolean values — a common pattern
  in tile schedulers and other control-flow-heavy kernels.

**Steps/Code to reproduce bug**
  ```
from cutlass._mlir import ir
  from cutlass.base_dsl.typing import Boolean, Int32
  from cutlass._mlir.dialects import arith as arith_d
  from cutlass.cutlass_dsl.cutlass import and_

  with ir.Context(), ir.Location.unknown():
      m = ir.Module.create()
      with ir.InsertionPoint(m.body):
          i1 = ir.IntegerType.get_signless(1)
          a = Boolean(arith_d.constant(i1, True))
          b = Boolean(arith_d.constant(i1, True))

          r_and = and_(a, b)   # Python `and` path → __dsl_and__
          r_amp = a & b        # bitwise `&`  path → __and__

      print(m)
  ```
  Actual output (before fix) — and_() emits 6 extra ops:
  ```
  %true = arith.constant true
  %true_0 = arith.constant true
  %0 = arith.extui %true : i1 to i32        ← unnecessary widening
  %1 = arith.cmpi ne, %0, %c0_i32 : i32    ← (a != 0)
  %2 = arith.extui %true : i1 to i32        ← unnecessary widening (again)
  %3 = arith.extui %true_0 : i1 to i32      ← unnecessary widening
  %4 = arith.select %1, %3, %2 : i32        ← select on i32
  %5 = arith.cmpi ne, %4, %c0_i32 : i32    ← back to i1
  %6 = arith.andi %true, %true_0 : i1       ← & path: single clean op
  ```
  Real-world occurrence:
[...]

PR #3087, opened 2026-03-01
CuTeDSL] Fix Boolean.__dsl_and__ to emit arith.andi directly for i1 operands

Johnsonms:
## Summary
https://github.com/NVIDIA/cutlass/issues/3086
In the CuTe DSL, Python’s and operator between two Boolean (i1) values (for example, in a while loop condition) is rewritten by the AST preprocessor as __dsl_and__. Prior to this fix, __dsl_and__ unconditionally promoted both operands from i1 to i32 using arith.extui, performed an arith.select, and then converted the result back to i1 via arith.cmpi ne. This resulted in six unnecessary MLIR operations for what should have been a single arith.andi.

In contrast, the bitwise & operator (__and__) already emitted the correct single-operation IR. Since both operators are semantically equivalent for Boolean inputs, this fix updates __dsl_and__ to generate the same clean IR as &.
  Root Cause

  Boolean.__dsl_and__ calls _binary_op_type_promote with promote_bool=True, which unconditionally
   widens Boolean → Int32 before performing the operation. For two Boolean (i1) inputs, the i32
  round-trip is semantically identical to bitwise AND but generates unnecessary IR.

## Fix
  Add a fast path at the top of __dsl_and__ that delegates to __and__ when both operands are
  Boolean: python/CuTeDSL/cutlass/base_dsl/typing.py
```
  def __dsl_and__(self, other, *, loc=None, ip=None):
      # Fast path: Boolean & Boolean → single arith.andi i1, no i32 round-trip
      if isinstance(self, Boolean) and isinstance(other, Boolean):
          return self.__and__(other, loc=loc, ip=ip)
      ...  # existing code unchanged
```

  Before (6 ops):
  %0 = arith.extui  %a : i1 to i32
  %1 = arith.cmpi ne, %0, %c0_i32 : i32
  %2 = arith.extui  %a : i1 to i32
  %3 = arith.extui  %b : i1 to i32
  %4 = arith.select %1, %3, %2 : i32
  %5 = arith.cmpi ne, %4, %c0_i32 : i32

  After (1 op):
  %0 = arith.andi %a, %b : i1

### Files Changed
image

  
## Testing
`  pytest examples/python/CuTeDSL/hopper/test_dsl_and_fix.py -v`
  # 3 passed

  Tests cover:
  1. and_() and & emit identical IR — single arith.andi, zero extui/select
  2. Both return Boolean with MLIR type i1 (valid for scf.ConditionOp)
  3. Full scf.while loop condition generates identical IR for both operators

### DIFF
--- a/python/CuTeDSL/cutlass/base_dsl/typing.py
+++ b/python/CuTeDSL/cutlass/base_dsl/typing.py
@@ -1021,6 +1021,14 @@ class Numeric(metaclass=NumericMeta, is_abstract=True):
             0 and 3 -> 0
             3 and 0 and ... -> 0
         """
+        # Fast path: Boolean & Boolean → single arith.andi i1 instruction.
+        # The general path promotes i1 operands to i32 via arith.extui, performs
+        # arith.select, then converts back to i1 via arith.cmpi ne — generating
+        # 6 unnecessary MLIR operations.  For Boolean inputs the semantics of
+        # `and` are identical to bitwise AND, so delegate directly to __and__.
+        if isinstance(self, Boolean) and isinstance(other, Boolean):
+            return self.__and__(other, loc=loc, ip=ip)
+
         is_true = self.__dsl_bool__(loc=loc, ip=ip)
 
         def and_op(lhs, rhs):


python/CuTeDSL/cutlass/base_dsl/typing.py
```diff
@@ -1021,6 +1021,14 @@ def __dsl_and__(self, other, *, loc=None, ip=None):
             0 and 3 -> 0
             3 and 0 and ... -> 0
         """
+        # Fast path: Boolean & Boolean → single arith.andi i1 instruction.
+        # The general path promotes i1 operands to i32 via arith.extui, performs
+        # arith.select, then converts back to i1 via arith.cmpi ne — generating
+        # 6 unnecessary MLIR operations.  For Boolean inputs the semantics of
+        # `and` are identical to bitwise AND, so delegate directly to __and__.
+        if isinstance(self, Boolean) and isinstance(other, Boolean):
```
anakinxc, line 1029: LGTM. @fengxie to double check

anakinxc:
Hi @Johnsonms 

To help merge this MR, can you revert the two files under example folder?

Thanks

Johnsonms:
> Hi @Johnsonms
> 
> To help merge this MR, can you revert the two files under example folder?
> 
> Thanks

Revert them, as there is no proper place for testing_typing.py. 

Thanks @anakinxc @fengxie

anakinxc:
@Junkai-Wu Can you merge this one?

PR #3097, opened 2026-03-09
[CuTe] adding DSL INT8 MMA support on SM80+, with CuTe C++ example

ayghri:
This PR adds  INT8 warp-level MMA with an Ampere batched GEMM example.


ayghri:
@brandon-yujie-sun any feedback on this?

PR #3116, opened 2026-03-20
Fix typo in elementwise_add.py

HydraQYH:
This looks like a typo.

### DIFF
--- a/examples/python/CuTeDSL/cute/ampere/kernel/elementwise/elementwise_add.py
+++ b/examples/python/CuTeDSL/cute/ampere/kernel/elementwise/elementwise_add.py
@@ -348,17 +348,17 @@ def run_elementwise_add(
 
         c = torch.zeros_like(a)
 
-        if is_a_dynamic_layout:
+        if not is_a_dynamic_layout:
             a_tensor = from_dlpack(a).mark_layout_dynamic()
         else:
             a_tensor = a
 
-        if is_b_dynamic_layout:
+        if not is_b_dynamic_layout:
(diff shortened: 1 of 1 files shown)


fengxie (approved):

fengxie (approved):

fengxie:
@brandon-yujie-sun I think we can merge?

examples/python/CuTeDSL/ampere/elementwise_add.py
```diff
@@ -348,17 +348,17 @@ def generate_tensors():
 
         c = torch.zeros_like(a)
 
-        if not is_a_dynamic_layout:
+        if is_a_dynamic_layout:
```
brandon-yujie-sun: There's an issue in the original code for the static layout but the proposed fix does not address it either unfortunately.

I think it should be:
```
        if not is_a_dynamic_layout:
            a_tensor = from_dlpack(a)
        else:
            a_tensor = a
```

brandon-yujie-sun (commented):
Thanks for reporting the issue and proposing the fix!

examples/python/CuTeDSL/ampere/elementwise_add.py
HydraQYH: @brandon-yujie-sun Thank you very much, I have fixed it.

brandon-yujie-sun: LGTM, @Junkai-Wu would you please help merge?

Issue #3117, opened 2026-03-20
[BUG] incorrect spin-loop synchronization


gonzalobg:
### Which component has the problem?

CuTe DSL

### Bug Report

**Describe the bug**
The following sample is using a spin-loop to acquire system-scope writes, but is using relaxed instead of acquire: https://github.com/NVIDIA/cutlass/blob/982748aa7356fa838c2ea4994ddcb0b2a4b4cefa/examples/python/CuTeDSL/distributed/distributed_gemm_reduce_scatter_blackwell.py#L1386

It then uses a warp barrier, but that warp barrier only orders memory operations issued by the same warp, not other warps or other thread blocks, like the example is relying on.

**Expected behavior**
I expected the spin-loop to use acquire memory ordering.

**Additional context**
This is a customer request (ping me internally for more details if needed).

brandon-yujie-sun:
@shubaoyu2

shubaoyu2:
thanks for debugging this! here is the fix MR https://github.com/NVIDIA/cutlass/pull/3118

XiaoSongXS:
please ping Bangyu / Xiao on cutedsl distributed example. there is a reason we use relaxed here

XiaoSongXS:
both acquire or relaxed work on this case as subsequent read is multimem.ld_reduce, which bypass l1. what acquire does is to CCTL.IVALL invalidate l1. l1 does not play in this story.

for good perf in this case, yes use relax is enough 

for general purpose programing, use acquire is a good idea

will have some compiler optimization in future that allow use to use acquire and generate relaxed code based on use case

gau-nernst:
Hi @XiaoSongXS, I'm also studying this GEMM-RS kernel. There are 2 additional issues I want to get clarifications from you:
1. This is using `gpu` scope for both flag increment and flag polling. Why is `sys` scope not used here?
2. There is no fence between `relaxed` flag polling and LDMC. Shouldn't `fence.proxy.alias` be necessary? Is it the same reason as "CCTL.IVALL invalidate l1. l1 does not play in this story."? I noticed `fence.proxy.alias` compiles to `MEMBAR.SC.SYS + CCTL.IVALL`

XiaoSongXS:
> This is using gpu scope for both flag increment and flag polling. Why is sys scope not used here?

[...]

PR #3118, opened 2026-03-20
[bugfix] use acquire to prevent reordering.

shubaoyu2:
we need acquire semantics to prevent reordering as https://github.com/NVIDIA/cutlass/issues/3117 said.


brandon-yujie-sun:
hi @Junkai-Wu could you help merge this?

PR #3121, opened 2026-03-20
fix: Handle zero-stride TMA basis in fill_tma_gmem_shape_stride

RobTand:
## Summary

When a block-scaled GEMM layout folds scale factors over a broadcast dimension (stride == 0), `fill_tma_gmem_shape_stride` passes the zero-stride basis to `basis_get()`, which produces undefined `gmem_prob_shape` and `gmem_prob_stride` values in the TMA descriptor. This results in invalid TMA descriptors that cause silent data corruption or runtime errors.

**Fix**: detect zero-stride basis via `is_constant<0>` and emit the correct degenerate descriptor (`shape=1, stride=0`) for broadcast dimensions. This is the standard TMA encoding for a dimension that contributes no unique data.

The bug is latent in all block-scaled GEMM configurations but only manifests when `K/SFVectorSize` creates a broadcast fold — e.g. when the scale factor vector size equals or exceeds the tile K extent.

## Changes

1 file, +8/-2 lines in `include/cute/atom/copy_traits_sm90_tma.hpp`

## Test plan

- [ ] Existing block-scaled GEMM unit tests pass (K≥128 tiles unaffected — codepath is only reached with zero-stride basis)
- [ ] `sm120_bs_gemm_nvf4_nvf4_f32_bf16.cu` with `TileShape = Shape<_128,_128,_64>` no longer produces invalid TMA descriptors

Signed-off-by: Rob Tand <robert.tand@icloud.com>

### DIFF
--- a/include/cute/atom/copy_traits_sm90_tma.hpp
+++ b/include/cute/atom/copy_traits_sm90_tma.hpp
@@ -874,8 +874,14 @@ fill_tma_gmem_shape_stride(Tensor<GEngine,GLayout>   const& gtensor,           /
     if constexpr (tma_i_rank == 1) {
       // Trivial contribution of this gmem mode to this tma mode
       auto ej = unwrap(get<i>(tma_gbasis_stride));
-      gmem_prob_shape[i]  = basis_get(ej, gmem_shape);
-      gmem_prob_stride[i] = basis_get(ej, gmem_stride);
+      if constexpr (cute::is_constant<0, decltype(ej)>::value) {
+        // Zero-stride basis: broadcast dimension — no unique contribution to this TMA mode
+        gmem_prob_shape[i]  = 1;
+        gmem_prob_stride[i] = 0;
+      } else {
+        gmem_prob_shape[i]  = basis_get(ej, gmem_shape);
+        gmem_prob_stride[i] = basis_get(ej, gmem_stride);
+      }
     } else {
       // Apply a recurrence to each gmem mode that contributes to this tma mode
       for_each(get<i>(tma_gbasis_stride), [&](auto ej) {


brandonmmusic-max:
Thank you for the mentioned, I am not working on it as this time as I’ve
been busy trying to work on an sm120 optimized attention kernel, which is
proving to be harder than I thought.

I will be happy to help coordinate with you and anything you need.

voipmonitor:
## Testing on RTX PRO 6000 (SM120, TP=4)

We applied both this PR and FlashInfer [#2786](https://github.com/flashinfer-ai/flashinfer/pull/2786) together on 4x RTX PRO 6000 Blackwell Server Edition (SM120, PCIe, 96GB each) running SGLang with `nvidia/Qwen3.5-397B-A17B-NVFP4`.

**Setup:**
- Docker image: `voipmonitor/sglang:test-cu132` (CUDA 13.2, PyTorch nightly cu132)
- CUTLASS 4.4.2 + this PR patch applied to headers
- FlashInfer 0.6.6 reinstalled from PR #2786 branch
- Backend: `--fp4-gemm-backend flashinfer_cutlass --moe-runner-backend flashinfer_cutlass`
- No speculative decoding, no torch.compile
- Single-request decode benchmark: 512 tokens, temperature=0

**Results (single-user decode throughput):**

| Config | Run 1 | Run 2 | Run 3 |
|---|---|---|---|
| Baseline (K=128 only) | 68.5 tok/s | 70.2 tok/s | 70.3 tok/s |
| With K=64 patches (both PRs) | 69.9 tok/s | 70.2 tok/s | 70.2 tok/s |

No measurable difference on this configuration. The patches applied cleanly and FlashInfer JIT compiled the K=64 kernels without errors, but the dispatch doesn't seem to select K=64 tiles for this model/shape combination, or the benefit is offset by TP=4 communication overhead.

**Questions:**
1. What model and configuration did you benchmark for the ~2x speedup? (The PR mentions Nemotron-3 on DGX Spark SM121 — was that single-GPU unified memory?)
2. Does K=64 benefit primarily smaller batch sizes or specific expert dimensions?
3. Is there a way to force K=64 tile selection to verify it's being dispatched?

Both patches are functionally correct on SM120 — no crashes, no compilation errors. The CUTLASS TMA zero-stride fix and EffBlk_SF clamping work as described.

RobTand:
Thanks for taking the time to review this — really appreciate it.

To address your question about forcing K=64 tiles and demonstrating necessity:

**SMEM constraint on SM120/121**

SM120/SM121 devices (RTX 5090, DGX Spark GB10) have 99 KB of opt-in shared memory per block, compared to 227 KB on SM100. This changes which tile shapes are viable for block-scaled GEMM.

**Pipeline stage impact**

The builder computes pipeline stages as `available_smem / bytes_per_stage`. For NVF4 with SFVectorSize=32:

| Tile Shape | Bytes/Stage | SM100 Stages | SM120/121 Stages |
|---|---|---|---|
| 128×128×K=128 | ~17 KB | 12 | 5 |
| 128×256×K=128 | ~26 KB | 8 | 3 |
| 128×128×K=64 | ~9 KB | 25 | 10 |
| 128×256×K=64 | ~13 KB | 17 | 7 |

K=128 tiles do work on SM120/121 (3–5 stages), so this isn't strictly about making things functional at K=128. The issue is that K=64 tiles — the natural choice for the constrained SMEM budget — cannot be instantiated at all.

**Why K=64 fails without this fix**

Two bugs surface only when `K/SFVectorSize < Blk_SF` (i.e., K=64 with SFVectorSize=32 gives NumSFAlongK=2, but Blk_SF=4):

- **TMA descriptor corruption** (`copy_traits_sm90_tma.hpp`): Scale factor folding creates a broadcast dimension with zero stride. `basis_get()` on a zero-stride basis produces undefined `gmem_prob_shape`/`gmem_prob_stride` values, resulting in an invalid TMA descriptor.

- **Scale factor layout overflow** (`sm120_blockscaled_mma_builder.inl`): The division `Blk_SF/MMA_NSF` assumes `Blk_SF ≤ NumSFAlongK`. When that doesn't hold, the resulting `Blk_Elems` layout overflows, producing an incorrect TMA tensor map for scale factors.

**Why these were never caught on SM100**

With 227 KB of SMEM, K=128 tiles get 8–12 pipeline stages — more than sufficient. K=64 tiles are never needed, so the code path was never exercised. Both bugs are latent but only manifest with small K values.

**Performance expectations**

The performance improvement from deeper pipelining may be modest on SM120/121, since these devices tend to be memory-bandwidth-bound. The primary motivation is correctness: K=64 tile shapes should be valid instantiations of the block-scaled GEMM builder, and currently they are not.

**How to test**

The simplest way to verify is a one-line change to an existing unit test. In `test/unit/gemm/device/sm120_blockscaled_tensorop_gemm/sm120_bs_gemm_nvf4_nvf4_f32_bf16.cu`, change:

```cpp
using TileShape = Shape<_128,_128,_256>;
```
to:
```cpp
using TileShape = Shape<_128,_128,_64>;
```

Without this PR, this will either fail to compile or produce corrupt results. With the fix, it should compile and pass `TestSmall` correctness checks.

Both FlashInfer (v0.6.6) and vLLM (v0.18.0) already define K=64 tile configs in their SM120 enums (`CtaShape128x128x64B`, etc.) but deliberately exclude them from dispatch tables because they can't be instantiated against upstream CUTLASS today.

brandonmmusic-max:
I got no reason to 

> ## Testing on RTX PRO 6000 (SM120, TP=4)
> We applied both this PR and FlashInfer [#2786](https://github.com/flashinfer-ai/flashinfer/pull/2786) together on 4x RTX PRO 6000 Blackwell Server Edition (SM120, PCIe, 96GB each) running SGLang with `nvidia/Qwen3.5-397B-A17B-NVFP4`.
> 
> **Setup:**
> 
> * Docker image: `voipmonitor/sglang:test-cu132` (CUDA 13.2, PyTorch nightly cu132)
> * CUTLASS 4.4.2 + this PR patch applied to headers
> * FlashInfer 0.6.6 reinstalled from PR [Fixed compilation error when using StreamK scheduler + PDL. (#2686) #2786](https://github.com/NVIDIA/cutlass/pull/2786) branch
> * Backend: `--fp4-gemm-backend flashinfer_cutlass --moe-runner-backend flashinfer_cutlass`
> * No speculative decoding, no torch.compile
> * Single-request decode benchmark: 512 tokens, temperature=0
> 
> **Results (single-user decode throughput):**
> 
> Config	Run 1	Run 2	Run 3
> Baseline (K=128 only)	68.5 tok/s	70.2 tok/s	70.3 tok/s
> With K=64 patches (both PRs)	69.9 tok/s	70.2 tok/s	70.2 tok/s
> No measurable difference on this configuration. The patches applied cleanly and FlashInfer JIT compiled the K=64 kernels without errors, but the dispatch doesn't seem to select K=64 tiles for this model/shape combination, or the benefit is offset by TP=4 communication overhead.
> 
> **Questions:**
> 
> 1. What model and configuration did you benchmark for the ~2x speedup? (The PR mentions Nemotron-3 on DGX Spark SM121 — was that single-GPU unified memory?)
> 2. Does K=64 benefit primarily smaller batch sizes or specific expert dimensions?
> 3. Is there a way to force K=64 tile selection to verify it's being dispatched?
> 
> Both patches are functionally correct on SM120 — no crashes, no compilation errors. The CUTLASS TMA zero-stride fix and EffBlk_SF clamping work as described.

I think his benchmarks are fair.  I did my own benchmarking her e https://github.com/brandonmmusic-max/sm120-moe-bench, but I was testing for effects on different things.  The fix may help more on the prefill side (i was getting 17k or so)  The prefill numbers are pretty good.  But I haven’t ran benchmark is such a manner, that I can’t exclude anny increase being related to p2p or mtp; i was optimizing for my particular neuro-symbolic pipeline for my local workflow..  I’m glad to hear you were getting 2x speed up!  Would love to hear more about your setup!

voipmonitor:
## Update: Prefill benchmarks + methodology fix

My earlier decode results were correct (no difference), but the initial prefill numbers were wrong — the "+19%" was a cold-start artifact (first request without warmup hitting radix cache miss + JIT warmup). Re-ran with proper methodology: 3 warmup + 5 measured runs per prompt size.

**Hardware:** 4x RTX PRO 6000 Blackwell Server Edition (SM120, 96GB **GDDR7** each, PCIe Gen5)
**Model:** nvidia/Qwen3.5-397B-A17B-NVFP4, TP=4, flashinfer_cutlass backend
**Patches:** CUTLASS #3121 + FlashInfer #2786 both applied, JIT cache cleared, FlashInfer reinstalled from PR branch

### Decode throughput (single user, 512 output tokens)

| Config | Run 1 | Run 2 | Run 3 |
|---|---|---|---|
| Baseline (K=128 only) | 68.5 tok/s | 70.2 tok/s | 70.3 tok/s |
| With K=64 patches | 69.9 tok/s | 70.2 tok/s | 70.2 tok/s |

### Prefill throughput (3 warmup + 5 measured, max_tokens=1)

| Prompt tokens | Baseline (K=128) | K=64 patch | Diff |
|---|---|---|---|
| 885 | 8,761 tok/s | 8,767 tok/s | +0.1% |
| 1,769 | 10,507 tok/s | 10,563 tok/s | +0.5% |
| 3,550 | 11,855 tok/s | 11,967 tok/s | +0.9% |
| 7,099 | 12,445 tok/s | 12,549 tok/s | +0.8% |

All differences are within noise (<1%). Standard deviation across runs was 0.1–3.1ms.

### Analysis

The MoE GEMM dimensions for Qwen3.5-397B with TP=4 are:
- **GEMM1** (gate+up): M=tokens, N=512, K=4096
- **GEMM2** (down): M=tokens, N=4096, K=256

For **decode** (M=1): the operation is a GEMV, entirely memory-bandwidth bound. Pipeline stages (K=64's advantage) don't help because compute is not the bottleneck — reading weights from GDDR7 dominates.

For **prefill** (M=885–7099): with TP=4, per-GPU M is ~220–1775. Even at these sizes, GDDR7 bandwidth (1.79 TB/s per GPU) appears sufficient that the additional pipeline depth from K=64 tiles doesn't provide measurable benefit over K=128.

**Note on hardware difference:** Your PR reports 2x speedup on DGX Spark (SM121, unified LPDDR5X). LPDDR5X has significantly lower bandwidth (~273 GB/s) than GDDR7 (1.79 TB/s per GPU × 4 = 7.16 TB/s aggregate). This likely explains why K=64's compute efficiency gains are visible on DGX Spark (compute-bound) but not on discrete RTX PRO 6000 GPUs (memory-bandwidth bound even for prefill).

The patches are functionally correct — JIT compilation succeeds, kernels load, no errors. They just don't provide a throughput benefit on this specific hardware configuration.

johnnynunez:
@depaulmillz

depaulmillz:
When instantiating the 128x128x64 NVFP4 tile I am seeing refcheck failures with this MR.

For the 128x128x64 with MXFP4, the reason you are hitting issues is the layout for SMEM that you have computed is `(((32, 4),1),((32,2),1,1),10) : (((16, 4),256),((0,1),2,256),256)`. 

image

Along the contiguous dimension you are copying 2 elements (shaded in green) then skipping over 2 elements (not shaded) and repeating this pattern. This is not possible to copy since we require 16B of contiguous elements.

To use this tiling pattern, it requires switching to a universal copy atom instead (copies are 2B contiguous) or loading 1 SFA/SFB tile per every 2 MMA tiles at least.

RobTand:
@johnnynunez i updated this and tried to address @depaulmillz 's issues with tiling

RobTand:
dramatically reduced the scope here since k=64 doesn't seem to improve the performance and significantly increases the complexity. Just fixes the root bug that was causing things to barf

ccecka:
I believe this should result in a TMA admissibility failure -- the gmem layout and the smem layouts are incompatible for TMA to execute the copy.

This should be a static assertion instead, correct?

RobTand:
@ccecka Thanks for taking a look — appreciate the review.

Good question — I initially thought the same, but I think this is actually a valid case rather than an inadmissible one.

The zero-stride basis here comes from a broadcast dimension in block-scaled GEMM scale factor layouts. Concrete example:

With NVF4 (SFVectorSize=32) and a K=64 tile, the scale factor TMA basis decomposition produces a mode where 32 elements share the same scale factor — CuTe encodes this as stride=0. The basis for that mode is `_0`, meaning "this dimension doesn't index into unique gmem data."

Currently `basis_get(_0, gmem_shape)` gets called on that zero basis and returns garbage, which corrupts the TMA descriptor. But the *intent* is sound: the broadcast dimension genuinely contributes no unique data, so the correct TMA encoding is `shape=1, stride=0` — "this mode is degenerate, skip it."

The downstream admissibility problem @depaulmillz found (SMEM contiguity < 16B with the 128x128x64 tile) is real but it's in the builder's layout math, not here. The TMA descriptor itself is well-formed with the broadcast handled correctly.

A `static_assert` would reject any layout that produces a zero-stride basis during TMA setup, which would block legitimate broadcast folds. I think the `is_constant<0>` check is the right fix at this layer — it makes `fill_tma_gmem_shape_stride` handle broadcast dimensions the way CuTe already represents them elsewhere.

Happy to be wrong though — if there's a reason broadcast modes should never reach TMA descriptor construction, a static assert makes more sense and the fix should move to the caller.

RobTand:
@ccecka That's a really good point and I want to make sure I get this right.

My reasoning was that the zero-stride basis here comes from SFVectorSize broadcast folding — 32 SMEM positions aliased to the same physical address via stride=0, so I thought TMA could serve it by writing once to the aliased address with a degenerate `shape=1, stride=0` GMEM dimension. But I may be wrong about how TMA admissibility validates the box shape against the SMEM tile — if it enforces strict shape matching regardless of aliasing, then this would be invalid.

Could you help me understand: does a zero-stride basis in `tma_gbasis_stride` always indicate an inadmissible layout, or can it arise legitimately for broadcast dimensions? If it's always inadmissible, I'll change this to a `static_assert` so the incompatibility is caught at compile time rather than papered over. The builder would then need to handle the K < SFVectorSize × Blk_SF case before it reaches TMA descriptor construction.

Really appreciate you taking the time to look at this — your insight on the TMA internals is exactly what I need here.

ccecka:
To my knowledge, it should always be invalid -- TMA is a glorified memcpy into smem so the contiguous elements that it sees in SMEM must correspond to real elements in gmem.

If you print the smem layout and the gmem layout I suspect you will find the "broadcast" modes of each to be misaligned (either different sizes or corresponding to different coordinates) and, therefore, the source and target are TMA-incompatible.

RobTand:
@ccecka appreciate the feedback, will take a look!

RobTand:
Closing this out. The underlying zero-stride TMA basis bug is real but only manifests with K=64 tiles (where K/SFVectorSize creates a broadcast fold), and testing showed K=64 doesn't improve performance — we're already at 99% theoretical memory bandwidth with K=128.

ccecka's feedback that this should be a static_assert rather than a graceful workaround is fair, but a compile-time guard for a tile shape nobody is using doesn't add much practical value. If K=64 support is revisited in the future, the builder-level layout fix (avoiding the zero-stride basis entirely) is the right approach, and depaulmillz's SMEM contiguity finding would need to be addressed as well.

PR #3135, opened 2026-03-30
[CuTeDSL] Add a render function hook to allow render layout natively

kainzhong:
This PR adds an additional parameter to `print_latex` and `print_latex_tv` to allow users to pass a render function that accepts the tikz latex code ({tikzpicture} section only) and renders the image. It simplifies the visualization process since now if they have a latex environment they can write a simple function and render the image without having to copy and paste hundreds of lines of code and compile the latex manually. It works even better if they are using jupyter notebook.

For example, I use module `jupyter_tikz` and this is how I render the image directly in my notebook:

```python
from IPython.display import display
from jupyter_tikz import TexFragment

os.environ['JUPYTER_TIKZ_PDFTOCAIROPATH'] = '/home/linuxbrew/.linuxbrew/bin/pdftocairo'

def render_tikz(tikz_code):
    tikz = TexFragment(tikz_code)  # Create the tex template object
    display(tikz.run_latex(tex_program='/home/linuxbrew/.linuxbrew/bin/pdflatex', rasterize=True))
```

and passing `render_tikz` as the hook:

image

Other options would be writing it to a local file, or posting to a web API. This would allow more flexibility to visualize the layouts especially if they are complicated.

### DIFF
--- a/python/CuTeDSL/cutlass/utils/print_latex.py
+++ b/python/CuTeDSL/cutlass/utils/print_latex.py
@@ -59,7 +59,10 @@ def tikz_color_tv(tid: int, vid: int) -> str:
 
 
 def print_latex(
-    x: Union[Layout, ComposedLayout], *, color: Callable = tikz_color_bwx8, render_func: Optional[Callable] = None
+    x: Union[Layout, ComposedLayout],
+    *,
+    color: Callable = tikz_color_bwx8,
+    render_func: Optional[Callable[[str], None]] = None,
 ) -> None:
     """
     Prints a layout.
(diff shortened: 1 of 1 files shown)


depaulmillz:
Thanks for the contribution. This looks great.

depaulmillz:
Hi @kainzhong would you be able to rebase, we would like to merge this.

kainzhong:
@depaulmillz I just rebased. 
FYI: I also vibe coded a visualization tool which I find is more convenient than print_latex (no latex installation needed, more features than layout / TV layout). It's basically a node.js port of common CuTe operations with SVG visualization code. It's currently hosted as a static website on https://kainzhong.github.io/CuTe-Layout-Visualizer/

(It's vibe coded so I'm not sure if the code quality is mergeable but I think if cutlass has something similar it would be super useful)

depaulmillz:
@brandon-yujie-sun @hwu36 can we merge this?

hwu36 (approved):

PR #3136, opened 2026-03-30
feat(CuTeDSL): print benchmark time from Blackwell dense_gemm CLI

aidando73:
## Summary

Just prints the timing


aidando73:
cc @Junkai-Wu @jinzh-nvidia

PR #3152, opened 2026-04-07
Add support for empty dataclass arguments

nandor:
A dataclass with no fields exposed a bug in `extract_dataclass_members`:

```
@dataclass
class Dummy:
  pass
```

The type/return path was inconsistent. This PR fixes the function to support empty dataclasses, which are useful in unions.


brandon-yujie-sun:
Thanks for proposing the change. Would it be possible to add a corresponding test as well?

Junkai-Wu (approved):

PR #3166, opened 2026-04-13
[CMake] Fix SM70/72 support range upper bound for CUDA 13.0

### LINKED ISSUE #3164: [BUG] CUTLASS_NVCC_ARCHS_SUPPORTED has wrong values leading to compilation failure

CMake files

### Bug Report

The variable CUTLASS_NVCC_ARCHS_SUPPORTED stores the list of architectures that is supported according to the cuda version. 
Code reference : https://github.com/NVIDIA/cutlass/blob/main/CMakeLists.txt#L175

# This prevents CUTLASS from building its own extensive examples/tests
set(CUTLASS_ENABLE_EXAMPLES OFF CACHE BOOL "" FORCE)
set(CUTLASS_ENABLE_TESTS OFF CACHE BOOL "" FORCE)
# set(CUTLASS_ENABLE_LIBRARY OFF CACHE BOOL "" FORCE)

FetchContent_MakeAvailable(cutlass)
```

> nvcc fatal   : Unsupported gpu architecture 'compute_70'


I would like to open a pull request to fix this issue.

Flink-ddd:
## Purpose
Fixes #3164.

The CUDA 13.0 compiler no longer accepts SM70/72 as target architectures, causing a fatal error when CUTLASS_NVCC_ARCHS_SUPPORTED includes them. The existing condition only checked VERSION_GREATER_EQUAL 11.4 with no upper bound.

Fix: add VERSION_LESS 13.0 upper bound so SM70/72 are only included when the compiler supports them. Volta remains fully supported on CUDA < 13.0.

## Test Plan
Verified CUTLASS_NVCC_ARCHS_SUPPORTED output with CUDA_VERSION=13.0 before and after the fix.

## Test Result
**Before fix:**

-- CUTLASS_NVCC_ARCHS_SUPPORTED = 70;72;75;80;86;87;89;90;90a;100;...

**After fix:**

-- CUTLASS_NVCC_ARCHS_SUPPORTED = 75;80;86;87;89;90;90a;100;...


Flink-ddd:
Hi @hwu36  , Could you please take a look when you have time? Thanks!

hwu36:
We haven't decided to drop volta.  Since some are still use it.  Could we only drop it with 13.x compiler?

Flink-ddd:
Sorry for the confusion in the description, SM70/72 are only excluded when CUDA_VERSION >= 13.0, they remain supported for all earlier versions.

irshadcc:
Thanks for the fix 🙏 @Flink-ddd

irshadcc:
@hwu36 Can this PR be merged ?

hwu36 (approved):

PR #3192, opened 2026-04-29
Add Hopper FP8 blockwise GEMM CuTeDSL example

Johnsonms:
## Summary

Adds a CuTeDSL port of CUTLASS Example 67's blockwise variant (`67_hopper_fp8_warp_specialized_gemm_with_blockwise_scaling.cu`) at `examples/python/CuTeDSL/hopper/dense_gemm_fp8_blockwise.py`.

The kernel computes

```
D = scale_d * relu(alpha * scale_a * scale_b * accum + beta * scale_c * C)
accum[m, n] = sum_k SFA[m // 128, k // 128] * SFB[n // 128, k // 128]
                    * (A[m, k*128:(k+1)*128] @ B[k*128:(k+1)*128, n])
```

with one Float32 scale factor per 128-row block of A (SFA) and per 128-column block of B (SFB), indexed by K-tile. Pipeline: a DMA warpgroup feeds the MMA warpgroup over a multi-stage SMEM ring; the MMA warpgroup runs WGMMA per K-tile, multiplies the partial accumulator by SFA*SFB before adding it to the running accumulator, applies the alpha/beta/scale_c/scale_d + ReLU epilogue, then R2S + TMA store of D.

## Test plan

- [x] Bit-exact (or within tolerance) match against the PyTorch blockwise reference for every supported tile / cluster combination
- [x] `compute-sanitizer --tool memcheck` clean
- [x] `--iterations N --use_cold_l2` benchmark works
- [x] Performance comparison against the C++ Example 67 binary on H100

## Correctness

All 12 combinations of `--tile_shape_mn` × `--cluster_shape_mn` pass the PyTorch blockwise reference at `--mnkl 256,256,256,1`:

| tile_shape_mn | cluster=(1,1) | (2,1) | (1,2) | (2,2) |
|---|---|---|---|---|
| (128, 128) | PASS | PASS | PASS | PASS |
| (128, 64)  | PASS | PASS | PASS | PASS |
| (64, 64)   | PASS | PASS | PASS | PASS |

Tile (128, 256) is intentionally not exposed — with the cooperative atom layout (2, 1, 1) it would span two SFB entries per tile, which the blockwise mainloop doesn't handle. This is documented in the parser and the kernel class.

`compute-sanitizer --tool memcheck` on the smallest config: **0 errors**.

Numbers from the initial port (commit `ede8d187`, no perf optimizations), `--mnkl M,N,K,1 --tile_shape_mn 128,128 --cluster_shape_mn 1,2`, 200 iter + 10 warmup:

| Problem (M, N, K) | C++ runtime | C++ TFLOPS | CuTeDSL runtime | CuTeDSL TFLOPS | CuTeDSL / C++ |
|---|---|---|---|---|---|
| 1024 × 1024 × 1024 | 0.010 ms | 207 | 0.019 ms | 112 | 54% |
| 2816 × 3072 × 16384 | 0.257 ms | 1103 | 0.684 ms | 414 | 38% |
| 4096 × 4096 × 4096 | 0.161 ms | 853 | 0.413 ms | 333 | 39% |

Five small commits stacked on the initial port, each measured independently against the prior commit on H100 at `cluster=(1,2)`, 200 iter + 10 warmup. Cumulative delta is +15 to +29% depending on shape.

| Commit | Optimization | Δ on 4096³ | Notes |
|---|---|---|---|
| `a1dad1d3` | BF16 D output | 0% | Matches the FP8-GEMM family convention; halves SMEM-D and unblocks `ab_stage` 5 → 6, though pipeline depth was not the binding constraint here. |
| `472f3f76` | Programmatic Dependent Launch | 0% large / +7% (1024³) | `griddepcontrol.wait` at entry + `use_pdl=True` on launch. Lets the runtime overlap the kernel's prologue with the prior kernel's tail on the same stream. Visible only on launch-bound shapes. |
| `74f5221a` | Cluster-aware rasterization swizzle | +1% | `--swizzle_size` default switches to a cluster-aware optimum (1 for any multicast cluster, 2 for `(1,1)`); user can still override. Sweep showed `sz=1` consistently beats `sz=2` on the three multicast configs. |
| `ef58af70` | Lazy `tma_store` producer_acquire | +1% | Acquire moved to immediately *before* the next TMA store instead of trailing the commit, matching DeepGEMM's `tma_store_wait<0>` pattern at the start of the next tile's epilogue. |
| `34a82c45` | Register split 24/240 (was 40/232) | **+26%** | DMA warpgroup only holds TMA descriptors + barrier addrs — 24 regs is enough. Shifting 8 regs to the math warpgroups eliminates the spill on the FP32 accumulator + scale registers in the inner loop. Dominant single trick. |

## Final performance

| Problem (M, N, K) | C++ runtime | C++ TFLOPS | CuTeDSL runtime | CuTeDSL TFLOPS | CuTeDSL / C++ | Δ vs baseline |
|---|---|---|---|---|---|---|
| 1024 × 1024 × 1024 | 0.010 ms | 207 | 0.015 ms | 144 | 70% | +29% |
| 2816 × 3072 × 16384 | 0.257 ms | 1103 | 0.596 ms | 476 | 43% | +15% |
| 4096 × 4096 × 4096 | 0.161 ms | 853 | 0.325 ms | 423 | 50% | +27% |

C++: `examples/67_..._gemm_with_blockwise_scaling --m=M --n=N --k=K --iterations=200 --verify=false`. CuTeDSL: `--iterations 200 --warmup_iterations 10`. The implementation is still meant as a CuTeDSL learning example and not a drop-in replacement for the C++ collective builders.

## Notes

- C must be passed with `is_dynamic_layout=False` so that `thr_mma.partition_C` sees static strides; with dynamic strides the per-thread offsets collapse to zero and `beta != 0` produces wrong results.
- Scale-factor indexing uses `(tile_coord * tile_size) // 128` so the small-tile (64×64, 128×64) configurations index SFA/SFB correctly.
- D output is BF16 (the FP8-GEMM family convention, matching the groupwise and grouped-blockwise ports); C input remains Float32 to keep the per-element `beta * scale_c * C` path in full precision.
- `--swizzle_size` defaults to a cluster-aware optimum (1 for any multicast cluster, 2 for the single-CTA case); pass an explicit value to override.
- Programmatic Dependent Launch (`griddepcontrol.wait` + `use_pdl=True`) is enabled at kernel entry; lets the runtime overlap this kernel's prologue with the prior kernel's tail on the same stream.

### DIFF
--- a/examples/python/CuTeDSL/hopper/dense_gemm_fp8_blockwise.py
+++ b/examples/python/CuTeDSL/hopper/dense_gemm_fp8_blockwise.py
@@ -203,8 +203,13 @@ class HopperFP8BlockwiseGemmKernel:
         self.epi_store_warp_id = (
             self.num_dma_warp_groups * self.num_warps_per_warp_group
         )
-        self.load_register_requirement = 40
-        self.mma_register_requirement = 232
+        # Register split: math warpgroups want as many regs as possible to
+        # avoid spills in the WGMMA inner loop + scale-and-accumulate path.
+        # Sweep on H100 (4096^3): 24/240 wins by ~25% over the standard 40/232,
+        # 32/240 is within noise of 24/240, 48/224 regresses. DMA warps only
+        # hold TMA descriptors + barrier addrs, so 24 regs is enough.
+        self.load_register_requirement = 24
+        self.mma_register_requirement = 240
         self.smem_capacity = utils.get_smem_capacity_in_bytes("sm_90")
 
         self.ab_stage = None
--- a/examples/python/CuTeDSL/hopper/dense_gemm_fp8_blockwise.py
+++ b/examples/python/CuTeDSL/hopper/dense_gemm_fp8_blockwise.py
@@ -114,7 +114,11 @@ def parse_arguments() -> argparse.Namespace:
         choices=[(1, 1), (2, 1), (1, 2), (2, 2)],
         default=(1, 2),
     )
-    parser.add_argument("--swizzle_size", type=int, default=2)
+    parser.add_argument(
+        "--swizzle_size", type=int, default=None,
+        help="Rasterisation swizzle. Defaults to cluster-aware optimum "
+             "(1 for multicast clusters, 2 for the single-CTA case).",
+    )
     parser.add_argument(
         "--raster_order", choices=["along_m", "along_n"], default="along_m"
     )
@@ -938,6 +942,12 @@ class HopperFP8BlockwiseGemmKernel:
         gd = cute.zipped_divide(d, tiler=c_shape)
         num_ctas_mnl = gd[(0, (None, None, None))].shape
         cluster_shape_mnl = (*cluster_shape_mn, 1)
+        # Cluster-aware swizzle: with TMA multicast on either axis, swizzle=1
+        # (tighter rasterisation) keeps the multicast peer's tile resident in
+        # L2 across consecutive CTAs and consistently beats sz=2 by 1-2% on
+        # H100. Without multicast, swizzle=2 is best.
+        if swizzle_size is None:
+            swizzle_size = 2 if cluster_shape_mn == (1, 1) else 1
         tile_sched_params = utils.PersistentTileSchedulerParams(
             num_ctas_mnl, cluster_shape_mnl, swizzle_size, raster_along_m
         )
@@ -1033,7 +1043,7 @@ def run(
     mnkl: Tuple[int, int, int, int],
     tile_shape_mn: Tuple[int, int] = (128, 128),
     cluster_shape_mn: Tuple[int, int] = (1, 2),
-    swizzle_size: int = 2,
+    swizzle_size: int | None = None,
     raster_along_m: bool = True,
     alpha_val: float = 1.0,
     beta_val: float = 0.0,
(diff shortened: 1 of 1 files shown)


hwu36:

Johnsonms:
Rebased onto current `main` (`147295a3`, CUTLASS 4.8) and re-validated. Ready for review.

Per the #3202 examples-tree reorganization, the file now sits at `examples/python/CuTeDSL/cute/hopper/kernel/dense_gemm/dense_gemm_fp8_blockwise.py`.

**Contents** — 5 commits: baseline port, then 4 independently reviewable optimizations (BF16 D output, PDL, cluster-aware rasterization swizzle, register reconfiguration 40/232 -> 24/240).

- Correctness vs `torch.einsum` reference: 10/10 runs at 2816x3072x16384, plus passes across cluster (1,1 / 1,2 / 2,1) and raster (along_m / along_n) configurations
- Performance, baseline commit -> tip (GFLOPS, 10 warmup + 200 iterations):

  | shape | baseline | tip | delta |
  |---|---|---|---|
  | 2816x3072x16384 | 392795 | 484241 | +23.3% |
  | 4096^3 | 307291 | 389313 | +26.7% |
  | 1024^3 | 100992 | 141929 | +40.5% |

Register reconfiguration is the dominant contributor; it eliminates spills in the math warpgroup.

**One follow-up, happy to fold in if preferred:** the example uses `utils.StaticPersistentTileScheduler` from the wheel, which now warns it has migrated to `examples/python/CuTeDSL/helpers/static_persistent_tile_scheduler.py`. I can switch to the helpers copy in this PR or a separate one.

PR #3193, opened 2026-04-29
Add Hopper FP8 groupwise GEMM CuTeDSL example

Johnsonms:
## Summary

Adds a CuTeDSL port of CUTLASS Example 67's groupwise variant (`67_hopper_fp8_warp_specialized_gemm_with_groupwise_scaling.cu`) at `examples/python/CuTeDSL/hopper/dense_gemm_fp8_groupwise.py`.

The kernel computes

```
D = scale_d * relu(alpha * scale_a * scale_b * accum)
accum[m, n] = sum_k SFA[m, k] * SFB[n // 128, k]
                    * (A[m, k*128:(k+1)*128] @ B[k*128:(k+1)*128, n])
```

`ScaleGranularityM = 1`: one Float32 scale per M row per K-tile for A (SFA) and one per 128-column block of B per K-tile (SFB). Per-row SFA is applied per accumulator element via a coordinate tensor that maps every WGMMA register fragment to its absolute M row, so the scale lookup does not depend on the WGMMA register layout.

Cooperative pipeline: one DMA warpgroup feeds two MMA warpgroups via a multi-stage SMEM ring. Tile shape is fixed at (256, 128, 128); output D is BFloat16 N-major.

## Test plan

- [x] Bit-exact (or within tolerance) match against the PyTorch groupwise reference across all four cluster shapes
- [x] `compute-sanitizer --tool memcheck` clean
- [x] `--iterations N --use_cold_l2` benchmark works
- [x] Performance comparison against the C++ Example 67 binary on H100

## Correctness

`--mnkl 256,256,256,1` × all four cluster shapes:

| cluster=(1,1) | (2,1) | (1,2) | (2,2) |
|---|---|---|---|
| PASS | PASS | PASS | PASS |

`compute-sanitizer --tool memcheck`: **0 errors**.

Numbers from the initial port (commit `b5e6f3e2`, no perf optimizations), `--cluster_shape_mn 1,2`, 200 iter + 10 warmup:

| Problem (M, N, K) | C++ runtime | C++ TFLOPS | CuTeDSL runtime | CuTeDSL TFLOPS | CuTeDSL / C++ |
|---|---|---|---|---|---|
| 1024 × 1024 × 1024 | 0.011 ms | 204 | 0.026 ms | 82 | 40% |
| 2816 × 3072 × 16384 | 0.216 ms | 1315 | 0.664 ms | 427 | 32% |
| 4096 × 4096 × 4096 | 0.111 ms | 1242 | 0.350 ms | 393 | 32% |

Four small commits stacked on the initial port. Each delta is measured against the prior commit at H100, `cluster=(1,2)`, 200 iter + 10 warmup. Cumulative delta is +10 to +15% depending on shape.

| Commit | Optimization | Δ on 4096³ | Notes |
|---|---|---|---|
| `f1387583` | Register split 24/240 (was 40/232) | **+9%** | DMA warpgroup only holds TMA descriptors + barrier addrs — 24 regs is plenty. Shifting 8 regs to the math warpgroups eliminates the spill on FP32 accumulator + per-row SFA registers. Same change as PR1 (where it gave +26%); smaller win here because PR2's gap mix is more memory-bound. |
| `125c322c` | Programmatic Dependent Launch | 0% large / +6% (1024³) | `griddepcontrol.wait` at entry + `use_pdl=True` on launch. Lets the runtime overlap this kernel's prologue with the prior kernel's tail on the same stream. Visible only on launch-bound shapes. |
| `9edc9a9e` | Cluster-aware rasterization swizzle | +0.5% | `--swizzle_size` default switches to a cluster-aware optimum (1 for any multicast cluster, 2 for `(1,1)`); user can still override. |
| `4117b0c6` | Lazy `tma_store` producer_acquire | ~0% | Acquire moved to immediately *before* the next TMA store instead of trailing the commit. Structurally aligned with DeepGEMM; near-zero perf delta because the existing `epi_stage=4` already provided enough slack. |

## Final performance

| Problem (M, N, K) | C++ runtime | C++ TFLOPS | CuTeDSL runtime | CuTeDSL TFLOPS | CuTeDSL / C++ | Δ vs baseline |
|---|---|---|---|---|---|---|
| 1024 × 1024 × 1024 | 0.011 ms | 204 | 0.023 ms | 94 | 46% | +15% |
| 2816 × 3072 × 16384 | 0.216 ms | 1315 | 0.604 ms | 469 | 36% | +10% |
| 4096 × 4096 × 4096 | 0.111 ms | 1242 | 0.322 ms | 432 | 35% | +10% |

C++: `examples/67_..._gemm_with_groupwise_scaling --m=M --n=N --k=K --iterations=200 --verify=false`. CuTeDSL: `--iterations 200 --warmup_iterations 10`. A further perf opportunity left as a follow-up: staging the per-row SFA tile through SMEM (TMA-loaded alongside A/B) rather than reading from GMEM per accumulator element — see DeepGEMM `sm90_fp8_gemm_1d2d.cuh:192-200`. The implementation is still meant as a CuTeDSL learning example and not a drop-in replacement for the C++ collective builders.

### DIFF
--- a/examples/python/CuTeDSL/hopper/dense_gemm_fp8_groupwise.py
+++ b/examples/python/CuTeDSL/hopper/dense_gemm_fp8_groupwise.py
@@ -361,6 +361,7 @@ class HopperFP8GroupwiseGemmKernel:
             cluster=(*self.cluster_shape_mn, 1),
             min_blocks_per_mp=1,
             stream=stream,
+            use_pdl=True,
         )
 
     # ------------------------------------------------------------------
@@ -411,6 +412,10 @@ class HopperFP8GroupwiseGemmKernel:
             cute.nvgpu.cpasync.prefetch_descriptor(tma_atom_a)
             cute.nvgpu.cpasync.prefetch_descriptor(tma_atom_b)
             cute.nvgpu.cpasync.prefetch_descriptor(tma_atom_d)
+
+        # PDL: wait for prior kernel on this stream to drain before any TMA.
+        # Paired with use_pdl=True on .launch(); no-op without it.
+        cute.arch.griddepcontrol_wait()
 
         cta_rank_in_cluster = cute.arch.make_warp_uniform(
             cute.arch.block_idx_in_cluster()
--- a/examples/python/CuTeDSL/hopper/dense_gemm_fp8_groupwise.py
+++ b/examples/python/CuTeDSL/hopper/dense_gemm_fp8_groupwise.py
@@ -182,8 +182,12 @@ class HopperFP8GroupwiseGemmKernel:
         self.epi_store_warp_id = (
             self.num_dma_warp_groups * self.num_warps_per_warp_group
         )
-        self.load_register_requirement = 40
-        self.mma_register_requirement = 232
+        # Register split: math warpgroups want as many regs as possible to
+        # avoid spills in the WGMMA inner loop + scale-and-accumulate path.
+        # Sweep (cf. PR1) showed 24/240 dominates 40/232. DMA warps only hold
+        # TMA descriptors + barrier addrs, so 24 regs is sufficient.
+        self.load_register_requirement = 24
+        self.mma_register_requirement = 240
         self.smem_capacity = utils.get_smem_capacity_in_bytes("sm_90")
 
         self.ab_stage = None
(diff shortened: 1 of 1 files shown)


hwu36:

Johnsonms:
Rebased onto current `main` (`147295a3`, CUTLASS 4.8) and re-validated. Ready for review.

Per the #3202 examples-tree reorganization, the file now sits at `examples/python/CuTeDSL/cute/hopper/kernel/dense_gemm/dense_gemm_fp8_groupwise.py`.

**Contents** — 4 commits: baseline port, then 3 independently reviewable optimizations (register reconfiguration 40/232 -> 24/240, PDL, cluster-aware rasterization swizzle).

- Correctness vs reference: 5/5 runs at 2816x3072x16384
- Performance, baseline commit -> tip (GFLOPS, 10 warmup + 200 iterations): 2816x3072x16384 393636 -> 458600 (**+16.5%**), 4096^3 371513 -> 433969 (**+16.8%**)

PR #3195, opened 2026-04-29
Add Hopper FP8 grouped blockwise GEMM (sparse-groups) CuTeDSL example

Johnsonms:
## Summary

Adds a CuTeDSL port of CUTLASS Example 68's sparse-groups variant (`68_..._grouped_gemm_with_blockwise_scaling_with_sparse_groups.cu`) at `examples/python/CuTeDSL/hopper/dense_gemm_fp8_grouped_blockwise_sparse_groups.py`.

Same per-row SFA (`ScaleGranularityM = 1`) + blockwise SFB (`ScaleGranularityN = 128`) FP8 grouped GEMM as the dense Example 68 port, plus host-side support for problem distributions where many groups have zero problem sizes.

Sparse-groups behaviour:

- `--problem_sizes` accepts groups with any zero dim. Validation only requires multiples of 128 for non-empty groups.
- Empty groups share a single stub GMEM allocation, so the metadata pointer table is always valid (the kernel never reads from the stubs).
- The host zeros all dims of every empty group before handing the problem-size table to the kernel. The persistent group tile scheduler computes a group's tile count from M*N alone, so a group with M, N > 0 but K = 0 would otherwise consume M*N linear tile slots and offset every later group's tiles. Forcing all dims to zero makes the scheduler reserve zero linear tiles for empty groups uniformly. The original sizes are kept for the reference and bandwidth paths.
- Reported GBPS uses the original (un-padded) sizes; empty groups contribute nothing to throughput.
- `--sparse_fraction` (with `--seed`) randomly empties a fraction of groups; the all-empty case is short-circuited before kernel launch.

Schedule note: the C++ source uses `KernelPtrArrayTmaWarpSpecializedPingpongFP8Blockwise`. This Python port keeps the cooperative schedule from the dense Example 68 variant (`atom_layout_mnk = (2, 1, 1)`). The sparse-groups host-side semantics are independent of pingpong vs cooperative scheduling, and the per-WG tensormap workspace required for true pingpong with grouped GEMM is left as a follow-up.

## Test plan

- [x] Bit-exact match for dense (no empty groups) across all four cluster shapes
- [x] Bit-exact match with explicit empty groups (M=0, N=0, K=0 in different positions, mixed)
- [x] Bit-exact match with `--sparse_fraction 0.5 / 0.9 / 1.0`
- [x] `compute-sanitizer --tool memcheck` clean (including with `--use_cold_l2`)
- [x] All-empty case short-circuits before kernel launch
- [x] Per-iteration `JitArguments._keepalive` ensures `--use_cold_l2` pre-generated workspaces survive the full benchmark
- [x] Performance comparison against the C++ Example 68 sparse-groups binary on H100

## Correctness

| Config | Result |
|---|---|
| Default 4 groups, dense, all four cluster shapes | PASS × 4 |
| `--num_groups 5 --problem_sizes "(256,256,256),(0,256,256),(128,256,256),(256,0,256),(256,256,0)"` | PASS (2 active groups) |
| `--num_groups 8 --sparse_fraction 0.5 --seed 11` | PASS (4 active groups) |
| `--num_groups 8 --sparse_fraction 0.9 --seed 3` | PASS (1 active group) |
| `--num_groups 4 --sparse_fraction 1.0` | PASS (no work, short-circuited) |

`compute-sanitizer --tool memcheck`, sparse + `--use_cold_l2`: **0 errors**.

Numbers from the initial port (commit `e1bcd94a`, no perf optimizations). All non-empty groups M=N=K=2048, `--sparse_fraction 0.5 --seed 11 --cluster_shape_mn 1,2`, 200 iter + 10 warmup. The C++ binary's `--m=N --groups=K` mode randomizes per-group sizes, so its GFLOPS reflect a different work distribution than CuTeDSL's "exactly half-empty" — they are not directly comparable but are included as a sanity baseline.

| Groups (half empty) | C++ runtime (random sizes) | C++ TFLOPS | CuTeDSL runtime | CuTeDSL TFLOPS |
|---|---|---|---|---|
| 16 (8 active 2048³) | 0.829 ms | 332 | 0.159 ms | 863 |
| 64 (32 active 2048³) | 3.275 ms | 336 | 0.592 ms | 929 |

CuTeDSL numbers reflect actual non-empty work only and are the appropriate signal for "throughput on the active groups".

Three small commits stacked on the initial port. PR4 shares PR3's mainloop and was already near its perf ceiling at baseline, so the cumulative delta is modest — each commit is kept primarily for structural alignment with the rest of the FP8 family.

| Commit | Optimization | Δ (largest case) | Notes |
|---|---|---|---|
| `520db59c` | Register split 24/240 (was 40/232) | ~0% | DMA warpgroup only holds TMA descriptors + barrier addrs. Effect is small here because PR4's mainloop matches PR3's and PR3 was already at 86-88% of the C++ ratio at baseline; the spill this trick eliminates is not the binding constraint. |
| `35420264` | Programmatic Dependent Launch | +0.8% (16 groups) | `griddepcontrol.wait` + `use_pdl=True`. Visible on the smallest-runtime configuration. |
| `e663f652` | Lazy `tma_store` producer_acquire | +0.5% | Acquire moved to immediately *before* the next TMA store instead of trailing the commit. Structural alignment with DeepGEMM. |

Cluster-aware swizzle skipped on PR4 — the grouped tile scheduler does not take a swizzle parameter. Two further follow-up opportunities are left out of scope: TMA-loaded per-row SFA staged through SMEM (see PR2 / DeepGEMM `sm90_fp8_gemm_1d2d.cuh:192-200`), and the in-kernel "compute-invalid but pipeline-valid" tile path that DeepGEMM uses to skip work for unfilled rows without stalling the barrier (`sm90_fp8_gemm_1d2d.cuh:255-352`) — this last one is particularly relevant for very sparse PR4 workloads.

## Final performance

| Groups (half empty) | C++ runtime | C++ TFLOPS | CuTeDSL runtime | CuTeDSL TFLOPS | Δ vs baseline |
|---|---|---|---|---|---|
| 16 (8 active 2048³) | 0.829 ms | 332 | 0.157 ms | 873 | +1% |
| 64 (32 active 2048³) | 3.275 ms | 336 | 0.590 ms | 931 | ~0% |

C++: `examples/68_..._with_blockwise_scaling_with_sparse_groups --m=2048 --n=2048 --k=2048 --groups=N --iterations=200`. CuTeDSL: `--iterations 200 --warmup_iterations 10 --sparse_fraction 0.5 --seed 11 --skip_ref_check`. For an apples-to-apples C++ comparison the maintainers should generate a benchmark file (`--benchmark=path.txt`) listing exactly the same half-empty problem distribution.

### DIFF
--- a/examples/python/CuTeDSL/hopper/dense_gemm_fp8_grouped_blockwise_sparse_groups.py
+++ b/examples/python/CuTeDSL/hopper/dense_gemm_fp8_grouped_blockwise_sparse_groups.py
@@ -306,8 +306,12 @@ class HopperFP8GroupedBlockwiseSparseGroupsGemmKernel:
         self.num_mma_threads = self.num_mma_warp_groups * self.num_threads_per_warp_group
         self.load_warp_id = 0
         self.epi_store_warp_id = self.num_dma_warp_groups * self.num_warps_per_warp_group
-        self.load_register_requirement = 40
-        self.mma_register_requirement = 232
+        # Register split: math warpgroups want as many regs as possible to
+        # avoid spills in the WGMMA inner loop + scale-and-accumulate path.
+        # Sweep (cf. PR1) showed 24/240 dominates 40/232. DMA warps only hold
+        # TMA descriptors + barrier addrs, so 24 regs is sufficient.
+        self.load_register_requirement = 24
+        self.mma_register_requirement = 240
         self.smem_capacity = utils.get_smem_capacity_in_bytes("sm_90")
         self.buffer_align_bytes = 1024
 
--- a/examples/python/CuTeDSL/hopper/dense_gemm_fp8_grouped_blockwise_sparse_groups.py
+++ b/examples/python/CuTeDSL/hopper/dense_gemm_fp8_grouped_blockwise_sparse_groups.py
@@ -480,6 +480,7 @@ class HopperFP8GroupedBlockwiseSparseGroupsGemmKernel:
             cluster=(*self.cluster_shape_mn, 1),
             min_blocks_per_mp=1,
             stream=stream,
+            use_pdl=True,
         )
 
     # ------------------------------------------------------------------
@@ -518,6 +519,10 @@ class HopperFP8GroupedBlockwiseSparseGroupsGemmKernel:
             cute.nvgpu.cpasync.prefetch_descriptor(tma_atom_a)
             cute.nvgpu.cpasync.prefetch_descriptor(tma_atom_b)
             cute.nvgpu.cpasync.prefetch_descriptor(tma_atom_d)
+
+        # PDL: wait for prior kernel on this stream before TMA. Paired with
+        # use_pdl=True on .launch(); no-op without it.
+        cute.arch.griddepcontrol_wait()
 
         cta_rank_in_cluster = cute.arch.make_warp_uniform(cute.arch.block_idx_in_cluster())
         cluster_coord_mnk = cta_layout_mnk.get_flat_coord(cta_rank_in_cluster)
--- a/examples/python/CuTeDSL/hopper/dense_gemm_fp8_grouped_blockwise_sparse_groups.py
+++ b/examples/python/CuTeDSL/hopper/dense_gemm_fp8_grouped_blockwise_sparse_groups.py
@@ -1,0 +1,1703 @@
+# Copyright (c) 2025 - 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: BSD-3-Clause
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+# 1. Redistributions of source code must retain the above copyright notice,
+#    this list of conditions and the following disclaimer.
+# 2. Redistributions in binary form must reproduce the above copyright notice,
+#    this list of conditions and the following disclaimer in the documentation
+#    and/or other materials provided with the distribution.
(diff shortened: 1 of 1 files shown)


hwu36:

Johnsonms:
Rebased onto current `main` (`147295a3`, CUTLASS 4.8), ported to the current NVVM bindings, and re-validated. Ready for review.

Per the #3202 examples-tree reorganization, the file now sits at `examples/python/CuTeDSL/cute/hopper/kernel/grouped_gemm/grouped_gemm_fp8_blockwise_sparse_groups.py`.

**NVVM TMA binding port**, identical to #3194: the previous revision no longer imported on DSL >= 4.7 (`CpAsyncBulkTensorLoadMode` removed, `elect_sync` now keyword-only). Ported with a capability-detecting wrapper following #3369, folded into the baseline port commit so it still builds and validates standalone.

**Contents** — 3 commits: baseline port, then 2 independently reviewable optimizations (register reconfiguration 40/232 -> 24/240, PDL).

- Correctness vs reference: 3/3 seeds at 64 groups x (2048, 2048, 4096) with `--sparse_fraction 0.5` (32 active / 32 empty), and at `--sparse_fraction 0.9` (3 active / 29 empty)
- Performance, 64 groups at 50% sparse: 1063062 GFLOPS, within noise of the baseline port — same limited-headroom situation as #3194

Deferred follow-ups from the description (sparse invalid-tile fast path, ping-pong buffering with per-warpgroup tensormap workspace) remain out of scope for this PR.

Issue #3065, opened 2026-02-24
[BUG] Compile error from including `"cute/tensor.hpp"` when building with MSVC with `/permissive-` option


edgchen1:
### Which component has the problem?

CUTLASS C++

### Bug Report

**Describe the bug**
When compiling for C++20 with MSVC and including `"cute/tensor.hpp"`, I observed the following compile error:

```
D:\source\cutlass\include\cute\stride.hpp(299,46): error C3545: 'Ints': parameter pack expects a non-type template argument [D:\test\cute_cpp_test\build\cute_cpp_test_program.vcxproj]
  (compiling source file '../main.cc')
      D:\source\cutlass\include\cute\stride.hpp(299,46):
      the template instantiation context (the oldest one first) is
          D:\source\cutlass\include\cute\atom\mma_traits.hpp(70,8):
          while compiling class template partial specialization 'cute::MMA_Traits<cute::UniversalFMA<D,A,B,C>,>'
          D:\source\cutlass\include\cute\atom\mma_traits.hpp(86,35):
          while processing the default template argument of 'cute::Layout '
          D:\source\cutlass\include\cute\layout.hpp(98,51):
          see reference to alias template instantiation 'cute::LayoutLeft::Apply<Shape>' being compiled
          with
          [
              Shape=cute::tuple<cute::_1,cute::_1>
          ]
          D:\source\cutlass\include\cute\stride.hpp(369,26):
          see reference to function template instantiation 'auto cute::compact_major<cute::LayoutLeft,Shape,cute::C<1>,0x0>(const Shape &,const Current &)' being compiled
          with
          [
              Shape=cute::tuple<cute::_1,cute::_1>,
              Current=cute::C<1>
          ]
          D:\source\cutlass\include\cute\stride.hpp(357,27):
          see reference to function template instantiation 'auto cute::detail::compact<cute::LayoutLeft,Shape,Current>(const Shape &,const Current &)' being compiled
          with
          [
              Shape=cute::tuple<cute::_1,cute::_1>,
              Current=cute::C<1>
          ]

D:\source\cutlass\include\cute\atom\mma_traits.hpp(86,19): error C2976: 'cute::Layout': too few template arguments [D:\test\cute_cpp_test\build\cute_cpp_test_program.vcxproj]
  (compiling source file '../main.cc')
      D:\source\cutlass\include\cute\layout.hpp(99,8):
      see declaration of 'cute::Layout'

[...]

PR #3198, opened 2026-04-30
[CuTe] [Fix] MSVC's inability to deduce a non-type parameter pack from a dependent template alias

SystemPanic:
When compiling FlashInfer on Windows, usage of CuTe stride.hpp causes error C3545 at `using Seq = typename Lambda::template seq<Shape>;`

MSVC fails at this because it can't deduce the non-type parameter pack Ints when it's accessed through a dependent template alias (a member of a dependent type).

The fix bypasses the dependent member alias entirely by using a standalone struct with explicit partial specializations.

This solves https://github.com/NVIDIA/cutlass/issues/3065

### DIFF
--- a/include/cute/stride.hpp
+++ b/include/cute/stride.hpp
@@ -286,6 +286,22 @@ namespace detail {
 template <class Major>
 struct CompactLambda;
 
+// Workaround for MSVC's inability to deduce a non-type parameter pack from a dependent template alias, causing error C3545.
+#ifdef _MSC_VER
+template <class Major, class Shape>
+struct CompactSeq;
+
+template <class Shape>
+struct CompactSeq<LayoutLeft, Shape> {
+    using type = tuple_seq<Shape>;
+};
+
+template <class Shape>
+struct CompactSeq<LayoutRight, Shape> {
+    using type = tuple_rseq<Shape>;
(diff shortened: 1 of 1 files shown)


include/cute/stride.hpp
```diff
@@ -286,6 +286,22 @@ namespace detail {
...
+template <class Shape>
+struct CompactSeq<LayoutLeft, Shape> {
+    using type = tuple_seq<Shape>;
+};
+
+template <class Shape>
+struct CompactSeq<LayoutRight, Shape> {
+    using type = tuple_rseq<Shape>;
+};
+#endif
```
ccecka: Is there any reason to guard this with MSC_VER? This implementation should work equally well for all compilers, correct?

d-k-b (approved):

include/cute/stride.hpp
SystemPanic: Yes, it works for any compiler. Removed in the last commit.

zcbenz:
@ccecka I verify that this PR fixes the compilation error with MSVC. Very weirdly that we didn't run into the error in MLX until CUDA 13.3.

ccecka (approved):

PR #3200, opened 2026-05-02
[CuTeDSL] Fix loop carried target scope

TungtungQia:
### Issue with scf.for IR Construction

Issue Description: 
https://github.com/TungtungQia/cutlass_scf_for/blob/f74fea9ce35868d3ae9f8d1dce1969d7250d3f90/python/CuTeDSL/cutlass/base_dsl/ast_preprocessor.py#L1506
`active_symbols.remove(active_symbol)`

This link deletes variables at the same scope level as the induction variable, causing scf.for's iter_args to miss capturing some 'active_symbol' variables.

In the case([test/examples/CuTeDSL/test_for_control_flow.py](https://github.com/TungtungQia/cutlass_scf_for/blob/3f646b74839933d6e837b96d7de9cbd4963b5eef/test/examples/CuTeDSL/test_for_control_flow.py#L1)), the for loop cannot modify the value of the variable 'acc'.

`for loop carried reused target actual=[0, 4], expected=[1, 4]
    assert actual == expected
AssertionError`

Fix: Only delete the induction variable, and retain other variables at the same scope level as the induction variable in the active_symbols.


anakinxc:
LGTM. @grypp to take a look as well

Junkai-Wu (approved):

PR #3204, opened 2026-05-05
[CuTeDSL] Make editable installs use exact runtime companion wheels

alecco:
This PR tightens CuTe DSL editable-install hygiene and removes a few import hazards that can make local development and test collection depend on ambient Python environment state. The main change teaches `prep_editable_install.py` to resolve the generated runtime payload from the downloaded `nvidia-cutlass-dsl` wheel’s metadata instead of copying files from whatever compatible-looking package happens to be installed in `site-packages`. 

### What changed

* Read the downloaded `nvidia-cutlass-dsl` wheel `METADATA` and extract exact-pinned `nvidia-cutlass-dsl-libs-*` companion requirements.
* Download and extract the selected runtime companion wheel into the same temporary install workspace.
* Support runtime provider selection through:

  * default provider: `base`
  * `CUTLASS_DSL_RUNTIME_PROVIDER=cu13`
  * matching package extras when present
* Clean stale generated editable-install payload before copying new files:

  * `cutlass/_mlir`
  * `lib/`
  * copied `py.typed` markers
* Validate that the editable runtime payload contains `cutlass/_mlir`, `lib/`, and at least one shared library before writing `VERSION.EDITABLE`.
* Add `.gitignore` entries for generated CuTe DSL editable-install artifacts.
* Replace the test-sharding fallback `device_info` import with CuTe DSL’s own CUDA runtime capability helper.
* Remove import-time `sys.path` mutation from the base DSL compiler module.
* Clean up ruff-visible unused imports / exception bindings.

### Why

Recent `nvidia-cutlass-dsl` packaging can split the metadata wheel from the generated runtime payload in companion `nvidia-cutlass-dsl-libs-*` wheels. Copying generated Python and shared libraries from ambient `site-packages` is fragile because different runtime providers can install overlapping payload paths. This PR makes editable installs reproducible by sourcing the payload from the exact companion wheel declared by the downloaded DSL wheel.

It also prevents stale generated runtime files from surviving provider or version changes, and avoids test/import behavior that depends on unrelated packages installed in the developer environment.

### Validation

* `python -m py_compile python/CuTeDSL/prep_editable_install.py`
* `ruff check python/CuTeDSL/prep_editable_install.py`
* `git diff --check`
* temp-only probe resolved default `libs-base` and explicit `cu13` companion wheels
* temp-only stale payload probe verified cleanup, fresh copy, and runtime validation
* `python -m py_compile test/utils/test_sharding.py`
* `ruff check --select F401 test/utils/test_sharding.py`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,11 @@
 __pycache__/
 cutlass_library.egg-info/
 /build*
+
+# CuTe DSL editable-install payload copied from nvidia-cutlass-dsl-libs-*.
+python/CuTeDSL/VERSION.EDITABLE
+python/CuTeDSL/cutlass/_mlir/
+python/CuTeDSL/cutlass/base_dsl/py.typed
+python/CuTeDSL/cutlass/cute/py.typed
+python/CuTeDSL/lib/
+python/CuTeDSL/nvidia_cutlass_dsl.egg-info/
--- a/python/CuTeDSL/cutlass/base_dsl/compiler.py
+++ b/python/CuTeDSL/cutlass/base_dsl/compiler.py
@@ -18,15 +18,12 @@ and executes it using MLIR's ExecutionEngine.
 from typing import Any
 import collections.abc
 import os
-import sys
 import inspect
+import importlib
 import types
 from .common import DSLRuntimeError
 from .utils.logger import log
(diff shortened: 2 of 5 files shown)


alecco:
Rebased on latest `main` branch.

PR #3206, opened 2026-05-07
[CuTeDSL] Update atomic_max_float32 to atomic_fmax in blockscaled GEMM

questa-wang:
The internal DSL package refactored atomic_max_float32 to atomic_fmax, which properly handles negative floats via sign-bit-aware integer atomics. Update the example to use the new API so it works with current DSL wheels.


brandon-yujie-sun:
Thanks, LGTM

Junkai-Wu (approved):

PR #3246, opened 2026-05-19
Fix FastDivmod divisor SSA transport for kernel regions (#3243)

### LINKED ISSUE #3243: [BUG] cutlass 4.5.0/4.5.1`cute.FastDivmodDivisor.divisor` can reference SSA value from outside isolated region

CuTe DSL

### Bug Report

   In CUTLASS / CuTe DSL 4.5.1, `cute.FastDivmodDivisor` has a public `.divisor` property, but accessing it after the FDD is stored in a JIT-created params object and passed to a kernel fails MLIR verification with a region-isolation error.

   The fast-divmod object itself works for `divmod(x, fdd)`, but `fdd.divisor` appears to preserve the original construction-side value instead of reconstructing/lowering a region-local divisor value.

   ### Repro

   Save as `repro_fdd_divisor.py` and run `python repro_fdd_divisor.py`.

   ```python
   from dataclasses import dataclass

   import cutlass
   import cutlass.cute as cute
   from cutlass import Int32


   @dataclass
   class Params:
       fdd: cute.FastDivmodDivisor


   @cute.jit
   def make_params(divisor: Int32) -> Params:
       return Params(cute.FastDivmodDivisor(divisor))


   @cute.kernel
   def write_divisor(out: cute.Tensor, params: Params):
       if tidx == 0:
           # This access fails MLIR verification.
           out[0] = params.fdd.divisor


   @cute.jit
   def entry(out: cute.Tensor, divisor: Int32):
       params = make_params(divisor)
       write_divisor(out, params).launch(grid=(1, 1, 1), block=(32, 1, 1))


   if __name__ == "__main__":
       print("cutlass", cutlass.__version__, cutlass.__file__)
       out_fake = cute.runtime.make_fake_tensor(
           cutlass.Int32, (1,), stride=(1,), assumed_align=4
       )
       cute.compile(entry, out_fake, Int32(0))
 ```

 ### Observed

 With cutlass==4.5.1:

 ```text
   cutlass 4.5.1 .../site-packages/nvidia_cutlass_dsl/python_packages/cutlass/__init__.py

   Verification failed:
   error: "out[0] = params.fdd.divisor"(...): 'cute.memref.store' op using value defined outside the region
    note: see current operation: "cute.memref.store"(%arg0, %15, <<UNKNOWN SSA VALUE>>) ...
    note: required by region isolation constraints
 ```

 ### Expected

 Accessing params.fdd.divisor inside the kernel should produce a region-local value and compile successfully, since .divisor is documented/public on FastDivmodDivisor.

 ### Additional note

Likely cause: FastDivmodDivisor.__extract_mlir_values__() only extracts _divisor_mlir, while __new_from_mlir_values__() preserves _original_divisor from the original object. Later .divisor returns that preserved original value, which may be defined
 outside the isolated kernel/JIT region.

zhils:
Fixes #3243

Expose encoded divisor plus scalar divisor IR through `FastDivmodDivisor::__extract_mlir_values__` and rebuild `.divisor` from region-local SSA in `__new_from_mlir_values__`. Update persistent tile schedulers to slice the FastDivmod tail using `extract_mlir_values` length. Add a CuTeDSL compile regression aligned with the issue repro.

## Test plan
- `pytest test/examples/CuTeDSL/test_fast_divmod_divisor_param_transport.py`

### DIFF
--- a/python/CuTeDSL/cutlass/cute/core.py
+++ b/python/CuTeDSL/cutlass/cute/core.py
@@ -6342,31 +6342,36 @@ class FastDivmodDivisor:
 
     def __extract_mlir_values__(self) -> List[ir.Value]:
         """Extract MLIR values for Host->Device transfer."""
-        # CRITICAL: Extract the FastDivmodDivisor MLIR value directly.
+        # Extract the encoded FastDivmod SSA ("_divisor_mlir") plus the scalar divisor SSA
+        # used when building it.
         #
         # This allows GridInvariantCodeMotionPass to:
         # 1. Recognize FastDivmodCreateDivisorOp in the IR
         # 2. Hoist it to the host side before kernel launch
         # 3. Pass the pre-computed divisor as a kernel argument
         #
-        # We only extract the _divisor_mlir to maintain compatibility with
-        # other code that assumes each FastDivmodDivisor has exactly 1 MLIR value.
-        # The _original_divisor is preserved in the object structure.
-        return [self._divisor_mlir]
+        # The scalar divisor must travel as its own SSA value so that after
+        # ``__new_from_mlir_values__`` inside an isolated kernel region,
+        # ``.divisor`` refers to SSA defined in that region (see GitHub issue #3243).
+        divisor_for_pack = self._original_divisor
+        if isinstance(divisor_for_pack, ir.Value):
+            divisor_ir = divisor_for_pack
+        else:
+            divisor_ir = Int32(divisor_for_pack).ir_value()
+        return [self._divisor_mlir, divisor_ir]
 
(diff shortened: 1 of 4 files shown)


questa-wang (approved):
LGTM. Thanks for the fix.

questa-wang:
Hi @zhils, thanks a lot for putting up this fix! We've been tracking the same root cause internally and have landed a broader patch that touches additional consumers beyond what's visible in the public tree, plus a small fix avoid re-emitting IR on the device side during reconstruction.
That fix is on track for the next public release (4.6). We'll keep this PR open as a reference until then, and close it out once 4.6 ships. Thanks again for the contribution!

PR #3270, opened 2026-05-25
[CuTeDSL] Make @cute.struct instances flattenable across scf.if / scf.while

### LINKED ISSUE #3268: [BUG] [CuTeDSL] `storage.<field>.get_tensor(...)` inside a dynamic `if` block fails with "encountered a user-defined Python object"

CuTe DSL

### Bug Report

**Describe the bug**

Calling `storage.<field>.get_tensor(...)` from inside a dynamic `if`
block fails at DSL trace time with:

```
DSLRuntimeError: The 'if' statement encountered a user-defined Python
object, which cannot be automatically converted into an dynamic
expression.
```

This blocks the natural warp-specialization pattern, where each
`if warp_idx == <role>:` branch extracts the SMEM tile it uses.

```python
import cutlass
import cutlass.cute as cute
import cutlass.utils as utils
from cutlass import Int32, Float32
from cutlass.cute.runtime import make_fake_tensor


@cute.struct
class Smem:
    sA: cute.struct.MemRange[Float32, 128]


@cute.kernel
def k(gA: cute.Tensor):
    smem_alloc = utils.SmemAllocator()
    storage = smem_alloc.allocate(Smem)
    sA_layout = cute.make_layout((128,), stride=(1,))

    if tidx == Int32(0):
        sA = storage.sA.get_tensor(sA_layout)   # fails here
        sA[0] = gA[0]


@cute.jit
def entry(gA: cute.Tensor):
    k(gA).launch(grid=(1, 1, 1), block=(32, 1, 1), smem=512)


gA = make_fake_tensor(Float32, (128,), stride=(1,), assumed_align=4)
cute.compile(entry, gA)
```

**Expected behavior**

`get_tensor` should be callable from inside a dynamic `if`. The SMEM
itself was already allocated by `smem_alloc.allocate(Smem)`; the
`get_tensor` call only constructs a typed view over it.

 - Environment location: Docker (`nvidia/cuda:12.8.1` base image; no
   GPU required for the trace-time failure)
 - `nvidia-cutlass-dsl` versions tested: **4.4.2** and **4.5.1** (both
   fail on 4.5.1 with the Python-level message above; 4.4.2 produced a
   more cryptic MLIR-level message)

**Additional context**

Workaround: hoist every `storage.<field>.get_tensor(...)` call to
outer kernel scope, before the warp-dispatch chain. Each role then
references the resulting view variable from inside its branch.

If the constraint stays, the diagnostic could name the fix
("extract storage views at outer kernel scope before the dynamic
branch") instead of just saying "user-defined Python object."

Related: #3243: `cute.FastDivmodDivisor.divisor` can reference an
SSA value from outside isolated region.

cheshire:
Fixes https://github.com/NVIDIA/cutlass/issues/3268

A `@cute.struct` instance captured into an `scf.if` branch or `scf.while` body fails the DSL trace with:

```
    DSLRuntimeError: The 'if' statement encountered a user-defined Python
    object, which cannot be automatically converted into an dynamic
    expression.
```

This blocks the natural warp-specialization pattern, where each `if warp_idx == <role>:` branch reads its tile from a shared storage struct.

A struct instance is fully described by its `base` pointer (already DynamicExpression-aware via `_Pointer`); every field instance is re-derived from `base + static offsets` on construction. Implement the DynamicExpression protocol on each decorated class by forwarding `__get_mlir_types__` / `__extract_mlir_values__` to `base`, and `__new_from_mlir_values__` to a fresh decorator invocation that re-derives the fields from a rebuilt base pointer.

### DIFF
--- a/test/python/CuTeDSL/test_struct_in_if.py
+++ b/test/python/CuTeDSL/test_struct_in_if.py
@@ -1,0 +1,192 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 - 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: LicenseRef-NvidiaProprietary
+#
+# Use of this software is governed by the terms and conditions of the
+# NVIDIA End User License Agreement (EULA), available at:
+# https://docs.nvidia.com/cutlass/latest/media/docs/pythonDSL/license.html
+#
+# Any use, reproduction, disclosure, or distribution of this software
+# and related documentation outside the scope permitted by the EULA
+# is strictly prohibited.
+
+"""
+Unit tests for using @cute.struct instances across DSL control flow.
+
+A struct instance is implicitly captured into the value set of an
+`scf.if` whenever its fields are accessed inside the branch (the natural
+warp-specialization pattern). For this to work, the struct must
+implement the DynamicExpression protocol so the DSL can flatten/unflatten
+it across the branch boundary.
+"""
+
+import unittest
+
+import cutlass
+import cutlass.cute as cute
+import cutlass.utils as utils
+from cutlass import Float32, Int32
+from cutlass.cute.runtime import make_fake_tensor
+
+
+@cute.struct
+class _OneTile:
+    sA: cute.struct.MemRange[Float32, 128]
+
+
+@cute.struct
+class _Inner:
+    sX: cute.struct.MemRange[Float32, 64]
+
+
+@cute.struct
+class _Outer:
+    inner: _Inner
+    sY: cute.struct.MemRange[Float32, 64]
+
+
+class TestStructInIf(unittest.TestCase):
+    def test_get_tensor_inside_dynamic_if(self):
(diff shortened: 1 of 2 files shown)


anakinxc:
LGTM @grypp to double check

test/python/CuTeDSL/test_struct_in_if.py
```diff
@@ -0,0 +1,192 @@
...
+    sA: cute.struct.MemRange[Float32, 128]
+
+
+@cute.struct
+class _Inner:
+    sX: cute.struct.MemRange[Float32, 64]
+
+
+@cute.struct
+class _Outer:
```
grypp, line 42: we could add a test with union as well.

grypp (approved):
Looks good to me. Thanks for contributing

test/python/CuTeDSL/test_struct_in_if.py
anakinxc: I'll add one when landing this internally

Junkai-Wu (approved):

PR #3276, opened 2026-05-26
[CuTeDSL] Fix the incorrect import of from_dlpack from docs

kainzhong:
The correct import path should be `from cutlass.cute.runtime import from_dlpack`. If you follow the faulty doc and use `from cutlass.cute import from_dlpack` it will throw an import error.


anakinxc:
LGTM

PR #3283, opened 2026-05-28
Enable tcgen05 blockscaled ops on Thor SM110

xiangg-nv:
Edge-LLM NvFP4 MoE CuTeDSL kernels on Thor use tcgen05 blockscaled MMA and SMEM-to-TMEM scale-factor copies. (Currently use the [patch](https://github.com/NVIDIA/TensorRT-Edge-LLM/blob/main/kernelSrcs/nvfp4_moe_cutedsl/README.md#thor-sm110-required-cutedsl-patches) to WR)

The existing checks only admitted the SM100/SM103 paths, so source-built CuTeDSL rejected SM110.

Admit Thor's blockscaled MMA arch aliases sm_101a and sm_110a, and allow the SM110f family for S2T tcgen05 copy ops.

Validation:

- git diff --check

- python3 -m py_compile python/CuTeDSL/cutlass/cute/nvgpu/tcgen05/mma.py python/CuTeDSL/cutlass/cute/nvgpu/tcgen05/copy.py

- grouped_blockscaled_gemm.py documented 4-group example on Thor SM110: PASS

- Edge-LLM nvfp4_moe AOT for sm_110/aarch64: 12/12 variants PASS

### DIFF
--- a/python/CuTeDSL/cutlass/cute/nvgpu/tcgen05/mma.py
+++ b/python/CuTeDSL/cutlass/cute/nvgpu/tcgen05/mma.py
@@ -387,6 +387,7 @@ class BlockScaledMmaOp(Tcgen05MmaOp):
     admissible_archs = [
         Arch.sm_100a,
         Arch.sm_103a,
+        Arch.sm_110a,
     ]
 
     def __post_init__(self) -> None:
--- a/python/CuTeDSL/cutlass/cute/nvgpu/tcgen05/copy.py
+++ b/python/CuTeDSL/cutlass/cute/nvgpu/tcgen05/copy.py
@@ -786,7 +786,8 @@ class _S2TCopyBase(CopyOp):
 
     :param cta_group: Cooperative Thread Array (CTA) group configuration
     :type cta_group: CtaGroup
-    :raises OpError: If the current architecture is not SM100f family or if invalid parameters are provided
+    :raises OpError: If the current architecture is not SM100f or SM110f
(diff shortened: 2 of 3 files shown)


python/CuTeDSL/cutlass/cute/nvgpu/tcgen05/mma.py
```diff
@@ -386,7 +386,9 @@ class BlockScaledMmaOp(Tcgen05MmaOp):
 
     admissible_archs = [
         Arch.sm_100a,
+        Arch.sm_101a,
```
brandon-yujie-sun: This is the same as sm_110a so it's a duplicate. I'd recommend we keep the sm_110a below and remove this since sm_110a is the new canonical name with cuda toolkit 13+

brandon-yujie-sun (commented):
Thanks for contributing the codes! Would it be possible to also add some test cases?

python/CuTeDSL/cutlass/cute/nvgpu/tcgen05/mma.py
xiangg-nv: Updated, thanks!

xiangg-nv:
> Thanks for contributing the codes! Would it be possible to also add some test cases?

Not sure which kind of test case should I add, it seems that we don't have any SM110 test cases, can you help to resolve this?

fengxie (approved):

PR #3289, opened 2026-06-01
[CuTeDSL] Add docs of how to call cute.jit functions in C++ via TVM-FFI

kainzhong:
Add some documentation to explain how to call cute.jit function in C++.

In our use case, we would like to avoid python overhead and stay in C++ environment, but also want to use CuTeDSL to implement kernels. Therefore we plan to use this approach so we can completely bypass python. I feel like some brief description here will make it easier for people to just get it working since people who work with kernels might not be familiar with TVM-FFI. This would act as a quick walkthrough so people can quickly get CuTeDSL working in C++.

Also, some explanation here also applies to using TVM-FFI exported function in C++ if you choose to export using an object file, so I figure it would be helpful./

### DIFF
--- a/media/docs/pythonDSL/guides/tvm_ffi_compilation.rst
+++ b/media/docs/pythonDSL/guides/tvm_ffi_compilation.rst
@@ -599,11 +599,12 @@ Then we need to compile and load this extension into PyTorch:
 
    def build_extension():
       include_dir = _tvm_ffi_config("--includedir")
+      dlpack_include_dir = _tvm_ffi_config("--dlpack-includedir")
       lib_dir = _tvm_ffi_config("--libdir")
       return load(
          name="tvm_ffi_demo_ext",
          sources=["extension.cpp"],
-         extra_include_paths=[include_dir],
+         extra_include_paths=[include_dir, dlpack_include_dir],
          extra_cflags=["-std=c++17"],
          # -ltvm_ffi to link, and -rpath so the .so is found at runtime. This is
          # the same libtvm_ffi.so that `import tvm_ffi` loads -> shared registry.
@@ -660,6 +661,9 @@ and then register it in the TVM FFI global registry with a string name. This req
 Next, you need to register this function object as a `Global Function <https://tvm.apache.org/ffi/guides/export_func_cls.html#global-functions>`_ with this API: ``tvm_ffi.register_global_func(func_name, f=None, override=False)``,
 where ``func_name`` is the string name to identify the function in the global registry, ``f`` is the TVM FFI function object returned by ``cute.compile``, and ``override=True`` allows overwriting an existing function with the same name in the registry.
 In this way, you can make this compiled ``@cute.jit`` TVM-FFI function accessible from other languages, including C++.
+
+Note this is not the only option to obtain a TVM-FFI function in C++. You can also export the compiled module to an object file and load it in C++ with TVM-FFI APIs.
+See `Exporting Compiled Module <https://docs.nvidia.com/cutlass/latest/media/docs/pythonDSL/cute_dsl_general/compile_with_tvm_ffi.html#exporting-compiled-module>`_ for more details.
 
 In C++, you can load the compiled function from the global registry with ``tvm::ffi::Function::GetGlobal`` (or ``tvm::ffi::Function::GetGlobalRequired``, which will throw if the function is not found).
 The returned object will be of type ``tvm::ffi::Function``.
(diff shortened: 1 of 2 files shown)


fengxie:
Thanks a lot! @tqchen

media/docs/pythonDSL/guides/tvm_ffi_compilation.rst
```diff
@@ -539,6 +539,157 @@ before launching the kernel based on that tensor's device index.
...
+
+   if __name__ == "__main__":
+      main()
+
+Calling convention of TVM-FFI in C++
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+When you compile a ``@cute.jit`` function with the ``--enable-tvm-ffi`` option, ``cute.compile`` will return a TVM FFI function object.
+Then you can register it as a `Global Function <https://tvm.apache.org/ffi/guides/export_func_cls.html#global-functions>`_ to make it accessible from other languages.
+
+In C++, you can load the compiled function from the global registry with ``tvm::ffi::Function::GetGlobal`` (or ``tvm::ffi::Function::GetGlobalRequired``, which will throw if the function is not found).
```
Observer007, line 668: How can we dump the tvm ffi function from cutedsl, can we leverage AOT? Can you explain this part more clearly. Other parts look good to me, thanks

fengxie (approved):

media/docs/pythonDSL/guides/tvm_ffi_compilation.rst
```diff
@@ -539,6 +539,157 @@ before launching the kernel based on that tensor's device index.
...
+   #include <torch/extension.h>      // pybind11 + at::Tensor
+   #include <tvm/ffi/container/tensor.h>  // tvm::ffi::TensorView
+   #include <tvm/ffi/function.h>     // tvm::ffi::Function
+
+   #include <string>
+
+   void apply_tvm_function(const std::string& name, at::Tensor &x, at::Tensor &y, at::Tensor &z) {
+      tvm::ffi::Function fn = tvm::ffi::Function::GetGlobalRequired(name);
+      DLTensor dl_x = {};
+      DLTensor dl_y = {};
```
tqchen, line 571: one possible option is to actually make use of tvm ffi's export mechanism just like flashinfer 

https://github.com/flashinfer-ai/flashinfer/blob/c3c40a7b90b792fc59f90f8f55c9e2de9c1b6833/csrc/flashinfer_gemm_binding.cu#L35

This way exported function also can be used in various tvm-ffi compatible scenarios and callable from pytorch


In this case, you can also do

```c++

using ffi = tvm::ffi;

void ApplyTVMFFIFunc(ffi::Function f, ffi::TensorView x, ffi::TensorView y, ffi::TensorView z) {
    f(f, x, y, z);
}

TVM_FFI_DLL_EXPORT_TYPED_FUNC(ApplyTVMFFIFunc);
```

tqchen (changes requested):

media/docs/pythonDSL/guides/tvm_ffi_compilation.rst
tqchen: related https://tvm.apache.org/ffi/packaging/python_packaging.html

kainzhong: Hi @tqchen

I think the use case here is to create and export a TVM-FFI function in python, then call it in C++, which would be helpful if you are working with a C++ heavy framework and you just want to leverage CuTeDSL's flexibility and run a kernel without going to python region.

I feel like `TVM_FFI_DLL_EXPORT_TYPED_FUNC` is more like exporting a C++ function to python, which seems to be the case in flashinfer and it's rather the opposite direction (we want to export in python and call it in C++). 
I include the pybind code in my example only because I need to run the C++ function to demonstrate the code is working (as I mentioned in the doc, in practice you probably want to call the C++ function in C++).

media/docs/pythonDSL/guides/tvm_ffi_compilation.rst
kainzhong: I add some description on this but really it's just doing AOT compilation and then calling this TVM-FFI API as in my example.

tqchen: One thing that may be relevant surfacing is https://github.com/NVIDIA/cutlass/blob/main/examples/python/CuTeDSL/dsl_tutorials/tvm_ffi/aot_use_in_cpp_bundle.cpp 

see how it is not using the global function registry and relies on the symbol being available

kainzhong: Ah yes. I also added some explanation of how to load symbols from shared libraries. Although I think using the global registry is more straightforward if I'm working with frameworks running in both C++ and python since in this way I don't need to handle the shared library manually.
Also the script path in `aot_use_in_cpp_bundle.cpp` is a bit outdated. I fixed that a bit as well

tqchen (approved):

fengxie (approved):

kainzhong:
Resolved conflicts

kainzhong:
@fengxie Hi I just rebased and resolved conflicts. Could you approve again and let's see if we can merge this? Thanks!

fengxie:
Thanks a lot! Merging.

PR #3292, opened 2026-06-02
add tileN = 8,16 for SM120 blockscale GEMM.

b8zhong:
image

It will be for use with SwapAB.

### DIFF
--- a/include/cutlass/epilogue/collective/builders/sm120_builder.inl
+++ b/include/cutlass/epilogue/collective/builders/sm120_builder.inl
@@ -191,7 +191,9 @@ sm120_compute_tile_shape_or_override() {
         }
       }
       else {
-        return Shape<_64, _32>{};
+        constexpr int EpiN = CTA_N < 32 ? CTA_N : 32;
+        constexpr int EpiM = CTA_N < 16 ? 128 : 64;
+        return Shape<Int<EpiM>, Int<EpiN>>{};
       }
     }
   } // EpilogueTileAuto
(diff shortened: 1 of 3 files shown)


b8zhong:
Hi @depaulmillz , I was wondering if you could take a look at this PR? Since I noticed you were the last one to add TileN = 32. Thanks!

depaulmillz:
Awesome. Have you been able to try with group GEMM as well?

b8zhong:
@depaulmillz Yes. Technically, it works (this PR is also compatible with group GEMM changes as well). But for example when testing on two common cases, DSR1 TP = 8 and Qwen-3 MoE TP = 1, the speedup can only be 3-5% for BS = 1. So it's faster (as expected), but not by much.

depaulmillz:
It looks like you will need to add an assertion to prevent compiling ping-pong with MMA_N=8 which will expect a (2,2,1) layout shape for the MMA. I saw some ref check errors when testing the MR on pingpong MMA_N=8 kernels due to this.

b8zhong:
@depaulmillz Just added a guard for it. Could you take a try now?

b8zhong:
Hi @depaulmillz , just bumping this PR for visibility. Thanks a lot!

depaulmillz:
@hwu36 can we merge this?

hwu36 (approved):

b8zhong:
Thanks @depaulmillz @hwu36 !

PR #3300, opened 2026-06-05
Fix print_tensor self-contained include

### LINKED ISSUE #3205: [BUG] [CuTe] print_tensor.hpp missing include for cute/pointer_flagged.hpp

CuTe DSL

### Bug Report

Description

`cute/util/print_tensor.hpp` appears to use symbols related to flagged pointers, such as `smem_ptr_flag_bits`, but it does not directly include the header that defines them.
This can cause clangd/static analysis diagnostics such as:

    Use of undeclared identifier 'smem_ptr_flag_bits'

Proposed fix：

add

     #include <cute/pointer_flagged.hpp>
 to include/cute/util/print_tensor.hpp

Expected behavior

  print_tensor.hpp should be self-contained with respect to the symbols it uses.

Actual behavior

  clangd reports smem_ptr_flag_bits as undeclared unless another header happens to include cute/pointer_flagged.hpp first.

Rohithmatham12:
## Summary
- Include `cute/pointer_flagged.hpp` directly from `cute/util/print_tensor.hpp` for `smem_ptr_flag_bits`.
- Add `cute/util/print_tensor.hpp` to the self-contained include checks.

Fixes #3205.

## Testing
- `git diff --check`
- Not run: self-contained include target, because this local environment does not have nvcc/CUDA headers installed.

### DIFF
--- a/include/cute/util/print_tensor.hpp
+++ b/include/cute/util/print_tensor.hpp
@@ -33,6 +33,7 @@
 #include <cute/config.hpp>           // CUTE_HOST_DEVICE
 
 #include <cute/layout.hpp>
+#include <cute/pointer_flagged.hpp>
 #include <cute/tensor_impl.hpp>
 
 namespace cute
--- a/test/self_contained_includes/CMakeLists.txt
+++ b/test/self_contained_includes/CMakeLists.txt
@@ -53,6 +53,7 @@ set(header_files_to_check
     cute/tensor.hpp
     cute/tensor_impl.hpp
     cute/underscore.hpp
+    cute/util/print_tensor.hpp
     # cute/algorithm
     cute/algorithm/axpby.hpp
     cute/algorithm/clear.hpp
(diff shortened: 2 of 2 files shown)


Rohithmatham12:
Still relevant and still intended for review.

This remains a small self-contained-include fix for `cute/util/print_tensor.hpp`: it directly includes the header that provides `smem_ptr_flag_bits` and adds `print_tensor.hpp` to the self-contained include check list so the missing dependency is covered going forward.

The branch is still mergeable from the current PR state. I have no new local CUDA validation to add beyond the original `git diff --check`, but I can refresh or adjust the patch if maintainers prefer a different include-check placement.

Rohithmatham12:
Still relevant and still intended for review.

This PR remains a focused self-contained include fix for `cute/util/print_tensor.hpp`: it directly includes the header that defines `smem_ptr_flag_bits` and adds `print_tensor.hpp` to the self-contained include check list so the dependency is covered going forward.

No scope changes from the original patch. I can refresh or adjust the include-check placement if maintainers prefer a different location.

PR #3317, opened 2026-06-14
[CuteDSL] Fix source location info for cute.arch.elect_one

pchen7e2:
The `loc` info was missed when creating `nvvm.elect_sync`. This means the later pipelines would not be able to tell that NVVM Op was from "this line of kernel source". The fix is to attach it.

We can check the effect by looking at source mapping in ncu:

Before: (elect.sync was not mapped correctly. In fact it's mapped to a line `@cute.jit`)

Screenshot 2026-06-14 at 15 33 51

After: (elect.sync was correctly mapped)

Screenshot 2026-06-14 at 15 35 36


pchen7e2:
Fixed merge conflict

PR #3332, opened 2026-06-19
[CuTeDSL] Fix unsigned integer constant lowering above INT64_MAX (#3312)

### LINKED ISSUE #3312: [BUG] UInt64 constants are compared to i64::max instead of u64::max in consteval

CuTe DSL

### Bug Report

**Describe the bug**
When compiling with Uint64 constants, if any of these constants are larger than the largest possible Int64 value, the code crashes. I encountered this bug while trying to implement a PRNG, many of which make use of full-width u64 constants. This goes for any intermediate consteval u64 value as well, so I can't construct a sufficiently large value from smaller ones.

This bug does not occur with runtime u64 values, i.e. I've resorted to passing all large constants as extra arguments to the kernel, but this is suboptimal. 

```python
import torch
import numpy as np
import cutlass

# 0xFFFFFFFFFFFFFFFF == 18446744073709551615
u64_max = np.iinfo(np.uint64).max

@cutlass.cute.kernel
def _set_val(a):
    a[0] = u64_max

@cutlass.cute.jit
def set_val(a):
    _set_val(a).launch()

array = torch.zeros(2, dtype=torch.uint64, device="cuda")
set_val(cutlass.cute.runtime.from_dlpack(array))
```
```
TypeError: get(): incompatible function arguments. The following argument types are supported:
    1. get(type: cutlass._mlir._mlir_libs._cutlass_ir._mlir.ir.Type, value: int) -> cutlass._mlir._mlir_libs._cutlass_ir._mlir.ir.IntegerAttr

Invoked with types: cutlass._mlir._mlir_libs._cutlass_ir._mlir.ir.IntegerType, int
```
(The error is rather cryptic, but it disappears when you use a number <= i64::max, so I'm fairly certain it's about a hidden overflow check.)

**Expected behavior**
All u64 constants up to and including u64::max should be valid.

Local, sm120

**Additional context**
I've tested it with all recent versions, up to nvidia-cutlass-dsl-libs-cu13==4.5.2

waynehacking8:
## Problem

Lowering an unsigned-integer DSL constant whose magnitude exceeds `INT64_MAX` raises a `TypeError` during IR emission. `IntegerAttr.get()` takes a signed C `int64_t`, so any value `> INT64_MAX` (i.e. `>= 2**63`) is rejected. `Uint32(2**32-1)` happened to work only because `4294967295 < INT64_MAX`, and signed types are always in range.

Fixes #3312.

This affects **two distinct lowering paths**:

```python
import cutlass.cute as cute
from cutlass import Uint64

@cute.jit
def f(a: Uint64):
    Uint64(2**64 - 1).ir_value()   # path 1: standalone constant
    (a | (2**64 - 1)).ir_value()   # path 2: all-ones mask via an operator

cute.compile(f, Uint64(1))
# TypeError: get(): incompatible function arguments.
```

1. **`Numeric.to(ir.Value)` / `.ir_value()`** — a standalone constant (the case in the issue).
2. **`_binary_op`** — a compile-time unsigned operand combined with a dynamic value, e.g. `a | 0xFFFFFFFFFFFFFFFF` (all-ones mask), `a + (2**63)`, or a comparison `a < (2**63 + 7)`. The operand's raw `.value` is handed to the operator, which builds the constant — so a fix in `Numeric.to` alone does **not** cover it.

## Fix

Add a shared `_lowered_const_value` helper that returns such constants as their **two's-complement signed value** (identical bit pattern), and call it from both paths. Scoped narrowly to unsigned integer types of width ≤ 64 whose value `>= 2**63`, so signed integers, floats, `bool`, and narrower / in-range unsigned values are byte-for-byte unchanged.

```python
def _lowered_const_value(operand):
    value = operand.value
    if (getattr(type(operand), "signed", None) is False
            and isinstance(value, int) and not isinstance(value, bool)
            and operand.dtype.width <= 64 and value >= (1 << 63)):
        value -= 1 << operand.dtype.width
    return value
```

(`Uint128`/`Int128` magnitudes above `INT64_MAX` are a separate concern — they exceed the `int64_t` parameter even after wrapping and would need binding-level support — so they are intentionally out of scope here.)

## Test

Adds `test/python/CuTeDSL/test_unsigned_constant.py` covering Uint64 above `INT64_MAX` via both the standalone-constant path and the binary-operator path (mask / offset / comparison), plus Uint64-in-range and Uint32/Int64 regressions.

## Validation

Validated against the released CuTe DSL `4.5.2` wheel, whose `Numeric.to` and `_binary_op` paths are identical to this file at the fix sites:
- On the unmodified wheel, both the standalone and binary-operator test cases fail with exactly the reported `TypeError`.
- Applying this patch to the wheel makes all cases compile cleanly; Uint32/Int64/in-range cases are unaffected.

The reproduction and fix are host-side IR construction (GPU-free); the original report is on sm_120 (RTX PRO 6000).

 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/python/CuTeDSL/cutlass/base_dsl/typing.py
+++ b/python/CuTeDSL/cutlass/base_dsl/typing.py
@@ -849,6 +849,27 @@ def _binary_op_type_promote(
         return a.to(b.dtype), b, b.dtype
 
 
+def _lowered_const_value(operand: "Numeric") -> Any:
+    """Return a compile-time numeric's value, adjusted for MLIR lowering.
+
+    ``IntegerAttr.get()`` takes a signed C ``int64_t``, so an unsigned integer
+    constant whose magnitude exceeds ``INT64_MAX`` must be lowered as its
+    two's-complement signed value (same bit pattern); otherwise the binding
+    raises ``TypeError: get(): incompatible function arguments``. Non-integer
+    and in-range values are returned unchanged. (#3312)
+    """
+    value = operand.value
+    if (
+        getattr(type(operand), "signed", None) is False
+        and isinstance(value, int)
+        and not isinstance(value, bool)
+        and operand.dtype.width <= 64
+        and value >= (1 << 63)
+    ):
+        value -= 1 << operand.dtype.width
+    return value
+
+
 def _binary_op(
     op: Callable[..., Any],
     promote_operand: bool = True,
@@ -931,13 +952,13 @@ def _binary_op(
         if isinstance(lhs.value, ArithValue) and isinstance(lhs, Integer):
             lhs_val = lhs.value.with_signedness(lhs.signed)
         else:
-            lhs_val = lhs.value
+            lhs_val = _lowered_const_value(lhs)
 
         rhs_val: Union[bool, int, float, ir.Value, ArithValue]
         if isinstance(rhs.value, ArithValue) and isinstance(rhs, Integer):
             rhs_val = rhs.value.with_signedness(rhs.signed)
         else:
-            rhs_val = rhs.value
+            rhs_val = _lowered_const_value(rhs)
 
         if flip:
             lhs_val, rhs_val = rhs_val, lhs_val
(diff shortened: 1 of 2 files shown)


waynehacking8:
Closing: the v4.6 release sync (#3362) fixed this independently — `_python_int_for_integer_attr` in `python/CuTeDSL/_mlir_helpers/arith.py` now coerces out-of-range unsigned literals on the `const()` NumericMeta path. Verified the original repro end-to-end: `Uint64(2**64 - 1).ir_value()` raises the reported TypeError on DSL 4.5.2 and compiles fine on 4.6.0, so #3312 can be closed too.

PR #3344, opened 2026-06-23
[CLI] Recover ssd and blockwise group gemm perf

myu-guo:
- reconfig to reduce register spilling issue in SSD kernel
- make uniform register to reduce spilling issue in Blockwise GGEMM.
- force to use fma to reduce register presure.


anakinxc:
LGTM

brandon-yujie-sun:
LGTM thanks!

Junkai-Wu (approved):

PR #3369, opened 2026-07-07
Fix grouped GEMM issue

JiaxuanBai:
The Hopper grouped GEMM wrapper selected the current NVVM TMA bindings only when the CUTLASS version prefix was exactly 4.6. Detect the TMALoadMode capability instead, preserving the legacy fallback for older bindings.


anakinxc:
LGTM

brandon-yujie-sun:
LGTM

Junkai-Wu (approved):

PR #3384, opened 2026-07-13
[CuTeDSL] Allow `reassoc` fastmath flag on vector reduction ops

antoniofrighetto:
Let Vector.reduce() accept optional fastmath flags (e.g., `reassoc`), meant to be leveraged by callers during MLIR/LLVM lowering.

### DIFF
--- a/python/CuTeDSL/_mlir_helpers/vector.py
+++ b/python/CuTeDSL/_mlir_helpers/vector.py
@@ -829,7 +829,9 @@ class Vector(ArithValue):
         :param fastmath: Optional fast-math flags for the reduction. Pass
             ``arith.FastMathFlags.reassoc`` to allow the compiler to lower
             the reduction as a tree instead of a strict left-to-right chain.
-            Defaults to ``None``.
+            Defaults to ``None``. Only supported for a full reduction of a
+            1-D vector to a scalar (``dim is None``), raises ValueError
+            for multi-dimensional reductions.
         :param acc: Optional accumulator.  For scalar reduction a scalar value;
             for multi-dim reduction a vector matching the result shape.
         :return: Scalar (when ``dim is None``) or :class:`Vector` (when
@@ -887,6 +889,9 @@ class Vector(ArithValue):
             return self._dtype(raw)
 
         # Multi-dimension reduction
+        if fastmath is not None:
+            raise ValueError("Fastmath flags are only supported for 1-D full reductions.")
+
         if dim is None:
             # Reduce all dims for N-D vector
             reduction_dims = list(range(ndim))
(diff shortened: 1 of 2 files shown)


python/CuTeDSL/_mlir_helpers/vector.py
```diff
@@ -927,7 +933,7 @@ def reduce(
             vec_1d = vector.multi_reduction(
                 kind, self, acc=p_acc, reduction_dims=partial_dims, loc=loc, ip=ip
             )
-            raw = vector.reduction(elem_ty, kind, vec_1d, acc=acc, loc=loc, ip=ip)
+            raw = vector.reduction(elem_ty, kind, vec_1d, acc=acc, **fmf_kwargs, loc=loc, ip=ip)
```
brandon-yujie-sun: This looks good, but for the multi_reduction case, shall we raise error to mention it's not supported?

test/python/CuTeDSL/test_vector_reduction_fastmath.py
```diff
@@ -0,0 +1,42 @@
...
+from cutlass._mlir_helpers.vector import Vector
+
+
+def _reduce_ir(fastmath=None):
+    """Return the IR for a ``Vector.reduce("add")`` over ``vector<16xf32>``."""
+    with ir.Context(), ir.Location.unknown():
+        module = ir.Module.create()
+        vec_ty = ir.VectorType.get([16], ir.F32Type.get())
+        with ir.InsertionPoint(module.body):
+            fn = func.FuncOp("test", ir.FunctionType.get([vec_ty], []))
```
brandon-yujie-sun, line 29: I think we'd also need a func.ReturnOp([]) to make it a real valid func op?

test/python/CuTeDSL/test_vector_reduction_fastmath.py
```diff
@@ -0,0 +1,42 @@
...
+            with ir.InsertionPoint(fn.add_entry_block()):
+                Vector(fn.arguments[0]).reduce("add", fastmath=fastmath)
+        return str(module)
```
brandon-yujie-sun, line 34: Maybe do a module.operation.verify()  before returning to make sure the constructed op is valid so that the validation on the serialized string would be meaningful?

brandon-yujie-sun (approved):
Thanks for contributing the codes! LGTM with some minor questions :-)

python/CuTeDSL/_mlir_helpers/vector.py
antoniofrighetto: Added, thanks. Raising ValueError when handling multi-dimensional reductions earlier (so dropped passing the flag here too; not sure if we somehow meant to only reject partial reductions, and allow passing the flag here).

test/python/CuTeDSL/test_vector_reduction_fastmath.py
antoniofrighetto: Sure thing, thanks!

python/CuTeDSL/_mlir_helpers/vector.py
brandon-yujie-sun: Thanks for the update. LGTM

PR #3386, opened 2026-07-14
Fix CuTe tuple protocol compatibility across CCCL versions

xiufanl:
## Summary

Recent CCCL versions no longer provide the private `cuda/std/__tuple_dir/structured_bindings.h` aggregation header, while older CCCL layouts do not always expose the tuple protocol primary declarations through the newer split headers. CuTe's NVRTC path needs to support both layouts.

Newer CCCL versions also provide `cuda::std::get` for `integer_sequence`. When CuTe aliases that type directly, its associated namespace brings the newer overload into unqualified lookup and conflicts with CuTe's tuple protocol.

## Changes

- Use the public `cuda/std/tuple` entry point for NVRTC compilation.
- Detect the available CCCL header layout and retain fallback tuple primary declarations only when neither the legacy nor split headers are available.
- Remove direct dependencies on the private structured-bindings header from CuTe tuple-like containers.
- Define a CuTe-owned `integer_sequence` and adapt the configured STL `make_integer_sequence` result into it.
- Keep tuple algorithms on their original unqualified `get` path; ordinary tuple-like ADL remains unchanged, while CuTe integer sequences no longer associate with the CCCL namespace.
- Add focused regression coverage for two- and three-input transforms and fold operations on integer sequences, including a type-ownership assertion.

## Compatibility

- Preserves older CUDA/CCCL support by continuing to use the configured STL namespace to generate integer packs.
- Uses CCCL-provided tuple declarations when supported and retains the existing fallback only for older layouts.
- Avoids requiring newer library APIs or compiler-specific forwarding helpers.

## Testing

- Added focused CuTe tuple regression cases in `test/unit/cute/core/tuple.cpp`.
- `git diff --check` passes; full CUDA validation is pending on the updated commit.

### DIFF
--- a/include/cute/config.hpp
+++ b/include/cute/config.hpp
@@ -111,22 +111,26 @@
 #endif
 
 #if defined(__CUDACC_RTC__)
-// Use CCCL's public tuple entry point so std tuple protocol declarations
-// follow the layout provided by the active CCCL version.
-#  include <cuda/std/tuple>
-
-// Older libcu++ releases do not provide the host std tuple primary templates
-// through a top-level tuple protocol header. Keep the legacy forward
-// declarations only for those layouts. Newer CCCL releases provide them via
-// either structured_bindings.h or the split tuple_size/tuple_element headers.
+// Load the tuple protocol from the layout provided by the active CCCL version.
+// Older CUDA toolkits may not provide the public <cuda/std/tuple> entry point,
+// while newer releases may remove the legacy detail/libcxx include path.
 #  if defined(__has_include)
-#    if !__has_include(<cuda/std/__tuple_dir/structured_bindings.h>) && \
-        !__has_include(<cuda/std/detail/libcxx/include/__tuple_dir/structured_bindings.h>) && \
-        (!__has_include(<cuda/std/__tuple_dir/tuple_size.h>) || \
-         !__has_include(<cuda/std/__tuple_dir/tuple_element.h>))
+#    if __has_include(<cuda/std/tuple>)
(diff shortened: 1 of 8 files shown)


include/cute/algorithm/tuple_algorithms.hpp
```diff
@@ -67,12 +67,25 @@ namespace cute
...
 {
-  return f(get<I>(static_cast<T&&>(t))...);
+  // Newer CCCL releases provide cuda::std::get for integer_sequence. Select
+  // CuTe's longstanding overload for that type, and preserve tuple-like ADL
+  // for every other type without adding a forwarding layer on older GCC.
+  if constexpr (is_integer_sequence<remove_cvref_t<T>>::value) {
+    return f(cute::get<I>(static_cast<T&&>(t))...);
+  } else {
+    return f(get<I>(static_cast<T&&>(t))...);
+  }
```
ccecka: Would much rather implement CuTe's own `integer_sequence` and/or fix this overload properly than WAR it and require this repetition everywhere.

xiufanl: Thanks - agreed. I replaced the per-call branching with a CuTe-owned `integer_sequence` and adapted the configured STL `make_integer_sequence` result into it. Since the associated namespace is now `cute`, the original unqualified `get` calls work without colliding with newer `cuda::std::get`, so `tuple_algorithms.hpp` is back to its direct ADL path. I also kept the focused transform/fold coverage and added a type-ownership assertion.

PR #3408, opened 2026-07-24
[CuTeDSL] Make autotuning more user friendly & new docs

kainzhong:
This PR adds a lot of things to the overall CuTeDSL autotuning process to make it easier to use and more convenient:

- Add `derived_params` to autotuning API so it computes derived arguments that must be decided during compile time but rely on autotuned parameters (whose values are different per configuration tuned)
- Make `testing.autotune_jit` able to decorate a class. This is for people who prefer to implement kernels as classes that fix the compile time params in `__init__` method and pass runtime params to `__call__`, so now they can do this
```
@testing.autotune_jit(
    params_dict={"COPY_BITS": [32, 64, 128], "THREADS_M": [4, 8]},
    update_on_change=["M", "N"],
    warmup_iterations=10,
    iterations=50,
)
class ElementwiseAddKernel:
    THREADS_M = 4

    def __init__(self, COPY_BITS=128):
        self.COPY_BITS = COPY_BITS

    @cute.jit
    def __call__(self, mA, mB, mC, M, N):
        elementwise_add(
            mA, mB, mC, M, N,
            copy_bits=self.COPY_BITS,
            threads_m=self.THREADS_M,
        )
```
- Support pruning undesired configurations (either invalid or known to be suboptimal) -- this aligns with triton's `prune_configs_by` in their autotuning API
- Add `autotune_suite` API so now user can provide a list of representative shapes and find the best overall kernel candidates with the ranking rule: sort according to how many times a configuration is accepted across all inputs, where "accepted" mean for a input case the configuration is in the top `accept_percentile` fastest ones
- Better autotuning docs. Previously the doc only explains how to autotune manually which is quite painful at least for me

### DIFF
--- a/media/docs/pythonDSL/guides/autotuning_gemm.rst
+++ b/media/docs/pythonDSL/guides/autotuning_gemm.rst
@@ -393,10 +393,20 @@ Note: autotune_suite expects a non-autotuned CuTeDSL function or class that is n
     compiled_gemm(*args)
 
 ``reports`` holds one timing table per case (each configuration's measured time, fastest first) so user can use this to decide what's their favorite configuration, 
-and ``recommended`` ranks configurations by how many cases accepted them within each input case's top ``accept_percentile`` configurations. 
+and ``recommended`` ranks configurations by how many cases accepted them within each input case's top ``accept_percentile`` configurations
 
 The heuristic behind this is simple. You don't want a configuration that is the fastest for one case but slow for others, and a configuration that is consistently good
-among all these representative cases is more likely to be the preferred choice in general (even if it's never the best for any of those cases). 
+among all these representative cases is more likely to be the preferred choice in general (even if it's never the best for any of those cases). It will return the top 5
+most widely accepted configurations like this:
+
+.. code-block:: text
+
+    recommended configs:
+        {'cluster_shape_mn': (2, 1), 'max_active_clusters': 76, 'mma_tiler_mn': (128, 128), 'use_2cta_instrs': True, 'use_tma_store': True}: accepted in 2/2 cases
+        {'cluster_shape_mn': (2, 1), 'max_active_clusters': 76, 'mma_tiler_mn': (256, 256), 'use_2cta_instrs': True, 'use_tma_store': True}: accepted in 2/2 cases
+        {'cluster_shape_mn': (2, 2), 'max_active_clusters': 36, 'mma_tiler_mn': (256, 128), 'use_2cta_instrs': True, 'use_tma_store': True}: accepted in 1/2 cases
+        {'cluster_shape_mn': (2, 1), 'max_active_clusters': 76, 'mma_tiler_mn': (256, 128), 'use_2cta_instrs': True, 'use_tma_store': True}: accepted in 1/2 cases
+        {'cluster_shape_mn': (2, 2), 'max_active_clusters': 36, 'mma_tiler_mn': (256, 128), 'use_2cta_instrs': True, 'use_tma_store': False}: accepted in 1/2 cases
 
 Note: about ``workspace_count``, this is the parameter that is crucial to ensure the benchmarking is accurate in terms of memory bandwidth.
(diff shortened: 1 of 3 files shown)


kainzhong:
@Junkai-Wu Hi, could you show some pointers for possible reviewers for this PR? This file doesn't show history other than version updates so I don't know who maintains this file. Thanks!

kainzhong:
@brandon-yujie-sun can I have a review? It's a lot of code but the logic should be fairly straightforward.

media/docs/pythonDSL/guides/autotuning_gemm.rst
```diff
@@ -148,3 +148,293 @@ approximation doesn't negatively impact performance for your specific use case.
...
+            cutlass.Float32,  # acc_dtype
+            use_2cta_instrs,
+            mma_tiler_mn,
+            cluster_shape_mn,
+            use_tma_store,
+        )
+        kernel(a, b, c, max_active_clusters, stream)
+
+Since ``cluster_shape_mn`` is a tunable parameter and it's unknown when we do the grid sweep and search the optimal configuration, 
+``autotune_jit`` needs to know how to compute ``max_active_clusters`` on the host side based on the current  ``cluster_shape_mn`` value when tuning.
```
brandon-yujie-sun: `on the host side` - I think you meant runtime?

python/CuTeDSL/cutlass/cute/testing.py
```diff
@@ -754,15 +823,188 @@ def tuning_wrapper(*args: Any, **kwargs: Any) -> Any:
...
+                )
+                outer.log().info(f"Total tuning time: {time() - start} s")
+                self._best_kernel[tuning_key] = best_kernel
+                self._best_config[tuning_key] = best_config
+                self._timings[tuning_key] = config_timings
+                return best_kernel(*args, **kwargs)
+
+        AutoTunedKernel.__name__ = kernel_cls.__name__
+        AutoTunedKernel.__qualname__ = kernel_cls.__qualname__
+        return AutoTunedKernel
```
brandon-yujie-sun: Seems we just return a totally different class from the original class being decorated and tuned. Does that mean we also lose all the original class's static methods and class attributes?

media/docs/pythonDSL/guides/autotuning_gemm.rst
```diff
@@ -148,3 +148,293 @@ approximation doesn't negatively impact performance for your specific use case.
 
 By following the methods above, you can customize your own auto-tuner to find the optimal GEMM kernel configuration 
 for specific matrix dimensions and data types, significantly improving computational performance for models.
+
+Autotuning JIT kernels with CuTeDSL functions
```
brandon-yujie-sun: Thank you for adding the detailed walkthrough on autotune_jit, very nice!

python/CuTeDSL/cutlass/cute/testing.py
```diff
@@ -786,15 +1036,26 @@ def __init__(
...
         """
-        # Create wrapper function
-        decorated_func = self._create_tuning_wrapper(
-            func, self.warmup_iterations, self.iterations, self.update_on_change
-        )
+        # Create wrapper function; an already-wrapped func (carrying
+        # _autotune_params) is returned unchanged by either path so stacked
+        # decorators merge their params instead of nesting tuners.
+        if inspect.isclass(func):
+            decorated_func = self._create_class_tuning_wrapper(
```
brandon-yujie-sun: What about adding test cases for the class path?

python/CuTeDSL/cutlass/cute/testing.py
```diff
@@ -605,18 +625,43 @@ def log(cls) -> logging.Logger:
...
+        return derived
+
     @classmethod
     def _create_tuning_wrapper(
         cls,
         func: Callable[..., Any],
         warmup_iterations: int,
         iterations: int,
         autotune_update_params: list[str],
+        prune_configs: Optional[Callable[..., bool]] = None,
```
brandon-yujie-sun: Thanks, would it make sense to add test case to cover the new `prune_configs`?

brandon-yujie-sun:
cc @fengxie

PR #3410, opened 2026-07-26
Require CUDA 12.9 for SM121 default architectures

davidkny22:
### Summary

CUTLASS currently includes `121` and `121a` in its default architecture list for CUDA 12.8,
but NVCC 12.8.61 advertises neither `compute_121` nor `sm_121`. NVCC 12.9.41 advertises both.
This empirically locates the SM121 compile-support boundary at CUDA 12.9.

This PR moves only those two entries to the existing CUDA 12.9 block beside `121f`.

### Mechanism

`CUTLASS_NVCC_ARCHS` defaults to `CUTLASS_NVCC_ARCHS_SUPPORTED`, and the enabled entries are
translated to CMake `CUDA_ARCHITECTURES`. The current 12.8 block therefore makes a fresh default
build request two targets that its active compiler does not advertise. CUTLASS's architecture
header already requires CUDA 12.9 for `CUTLASS_ARCH_MMA_SM121_SUPPORTED` and
`CUTLASS_ARCH_MMA_SM121A_ENABLED`; this change aligns the default CMake policy with that gate.

### Reproducer

For each installed toolkit, I ran a fresh native CMake configure and compared its default
`CUTLASS_NVCC_ARCHS` value with that compiler's own `nvcc --list-gpu-arch` output:

| Toolkit | Before | After | SM121-family defaults |
|---|---:|---:|---:|
| CUDA 12.8.61 | 2 unsupported of 17 | 0 unsupported of 15 | 2 → 0 |
| CUDA 12.9.41 | 0 unsupported of 23 | 0 unsupported of 23 | 3 → 3 |
| CUDA 13.0.48 | 0 unsupported of 23 | 0 unsupported of 23 | 3 → 3 |

Every non-moved default remains in the same order and maps to a target advertised by the
matching compiler. A CUDA 12.8 `CUTLASS_NVCC_ARCHS=121` request still configures, remains cached
as `121`, and is warned as unsupported.

### Testing done

Fresh configure-only checks completed on Windows build 26200 with CMake 4.3.4, Ninja 1.13.0,
MSVC 19.44, and CUDA 12.8.61/12.9.41/13.0.48. The configures used
`--allow-unsupported-compiler`. The host GPU was an RTX 4070 Laptop GPU (SM89), driver 595.79;
no GPU execution was involved. Reconfiguring the pre-change CUDA 12.8 cache retained its old
list and warned about `121;121a`.

CUTLASS has no in-tree test harness for the root version-to-default-architecture policy. I did
not add a parser or new test framework for this two-token policy change.

### Limitations

- No SM121 runtime behavior was tested; this is a compiler-target policy change.
- The change affects fresh default lists. Existing configured build directories retain their
  cached `CUTLASS_NVCC_ARCHS` value until that cache is cleared or updated.
- The explicit CUDA 12.8 `121` override was checked, but explicit overrides were not tested
  exhaustively.
- Validation covers NVCC 12.8.61, 12.9.41, and 13.0.48; it does not cover Clang-CUDA or other
  toolkit builds.

cc @Junkai-Wu


Junkai-Wu (approved):

Junkai-Wu:
Thanks for the fix.

PR #3417, opened 2026-07-29
[CuTeDSL] Fix prep_editable_install.py for split nvidia-cutlass-dsl wheels

### LINKED ISSUE #3416: [BUG] CuTeDSL editable install broken: prep_editable_install.py extracts nothing from the split nvidia-cutlass-dsl wheels (>= 4.4.0)
`python/CuTeDSL/prep_editable_install.py` downloads `nvidia-cutlass-dsl` with `pip download --no-deps`. Since 4.4.0 that wheel is a metadata-only shim (`nvidia_cutlass_dsl-4.6.1-py3-none-any.whl` is ~10 KB: one `.pth` + dist-info); the actual libraries live in `nvidia-cutlass-dsl-libs-{base,core,cu12,cu13}` wheels, under `dsl_packages/` rather than `python_packages/`. The script therefore copies nothing — `Summary: 0 lib files, 0 Python files copied` — but still logs "Setup completed successfully!" and exits 0. The subsequent `pip install -e .` / `setup.sh --editable` also succeeds, and the breakage only surfaces at first use: `import cutlass` → `ModuleNotFoundError: No module named 'cutlass._mlir'`.

Repro: fresh venv on x86_64/aarch64 Linux, `python prep_editable_install.py` at any commit since the script landed.

Fix proposal in the linked PR: detect the shim, follow its `nvidia-cutlass-dsl-libs-*` `Requires-Dist` for one CUDA variant, merge all payload dirs, error loudly on empty extraction. PR #3204 previously attempted this but is stale and predates the 4.6.x layout (its default `base` provider no longer contains `lib/*.so`).

zkyue:
Fixes #3416.

### Symptom
The CuTeDSL editable dev flow (`python/CuTeDSL/prep_editable_install.py`, then `setup.sh --editable`) produces an install that fails at `import cutlass` with `ModuleNotFoundError: No module named 'cutlass._mlir'`. The prep script prints "Setup completed successfully!" while its summary shows `0 lib files, 0 Python files copied`.

### Root cause
Since 4.4.0 the `nvidia-cutlass-dsl` wheel on PyPI is a ~10 KB metadata-only shim; the runtime libraries and python payload moved to `nvidia-cutlass-dsl-libs-{base,core,cu12,cu13}` wheels (declared via `Requires-Dist`, cu13 behind the `cu13` extra), and the in-wheel payload directory was renamed `python_packages/` → `dsl_packages/`. The script still runs `pip download --no-deps nvidia-cutlass-dsl`, so it gets only the shim, extracts nothing, and exits 0 anyway.

### Fix
- Detect a metadata-only top-level wheel by content; wheels that carry their own payload (<= 4.3.x) are handled exactly as before.
- For split wheels, follow the `Requires-Dist` entries within the `nvidia-cutlass-dsl-libs-*` family transitively, restricted to one CUDA variant (`--cu12` default / `--cu13`, matching `setup.sh`), and extract all of them.
- Merge every bundled `python_packages/cutlass` / `dsl_packages/cutlass` tree (the payload now spans several wheels) instead of only the first match.
- Raise a clear error (instead of reporting success) when no matching platform artifact exists or no runtime library / cutlass python payload was extracted.

### Repro / verification
```
python -m venv v && . v/bin/activate && pip install -U pip
cd python/CuTeDSL
python prep_editable_install.py          # before: 0 files copied, exit 0; after: libcute_dsl_runtime.so + cutlass/_mlir vendored
CUTLASS_IR_BUILD_DIR=$PWD ./setup.sh --editable
python -c "import cutlass, cutlass.cute as cute; print(cutlass.__file__)"
```
Verified: 4.6.1 cu12 end-to-end (import + `cute.compile` of a trivial kernel, no GPU required), cu13 dependency resolution, unchanged behavior on the pre-split 4.3.5 wheel, and the new failure messages.

Related: #3204 attempted this earlier (now stale); this PR is confined to `prep_editable_install.py`, follows the shim's own `Requires-Dist` pins (so it keeps working when the companion set changes, e.g. the cu12/core additions in 4.6.x), and keeps the CUDA-variant choice consistent with `setup.sh --cu12/--cu13`.

### DIFF
--- a/python/CuTeDSL/prep_editable_install.py
+++ b/python/CuTeDSL/prep_editable_install.py
@@ -17,14 +17,16 @@ extracts required libraries and Python packages, and sets up the development
 environment for CuTeDSL.
 """
 
+import argparse
 import subprocess
 import sys
 import shutil
 import tempfile
 import zipfile
 import re
+from email.parser import Parser
 from pathlib import Path
-from typing import Tuple
+from typing import List, Optional, Tuple
 import logging
 
 # Configure logging
@@ -33,6 +35,14 @@ logger = logging.getLogger(__name__)
 
 # Constants
 PACKAGE_NAME = "nvidia-cutlass-dsl"
+# Starting with 4.4.0, the nvidia-cutlass-dsl wheel on PyPI is a metadata-only
+# shim: the runtime libraries and python packages moved to per-platform wheels
+# (nvidia-cutlass-dsl-libs-base/-core/-cu12/-cu13) referenced via Requires-Dist.
+SPLIT_PACKAGE_PREFIX = f"{PACKAGE_NAME}-libs-"
+DEFAULT_CUDA_VARIANT = "cu12"
+# Directory names that carry the bundled cutlass python package inside the
+# wheel(s): pre-split wheels use python_packages/, split wheels dsl_packages/.
+PYTHON_PACKAGE_DIR_NAMES = ("python_packages", "dsl_packages")
 
 
 class CutlassDSLSetupError(Exception):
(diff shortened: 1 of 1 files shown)


brandon-yujie-sun:
@zekunf-nv

Issue #2636, opened 2025-09-14
[BUG] Cutlass Python DSL: could not get source code


LRlr239:
### Which component has the problem?

CuTe DSL

### Bug Report

**Describe the bug**
```python
import cutlass
import cutlass.cute as cute


@cute.jit
def add_demo_02():
   a = cutlass.Int32(1)
   b = cutlass.Int32(2)
   cute.printf("a + b = {}", a + b)

add_demo_02()
```

got following Error:
```text
>>> import cutlass.cute as cute
>>> 
>>> 
>>> @cute.jit
... def add_demo_02():
...    a = cutlass.Int32(1)
...    b = cutlass.Int32(2)
...    cute.printf("a + b = {}", a + b)
... 
>>> add_demo_02()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.12/inspect.py", line 1260, in getsourcelines
    lines, lnum = findsource(object)
                  ^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/inspect.py", line 1089, in findsource
    raise OSError('could not get source code')
OSError: could not get source code

```

**Steps/Code to reproduce bug**
1. open a python shell / jupyter
2. paste the code above

**Expected behavior**

**Environment details (please complete the following information):**
- python 3.12.3 main
- cuda V12.9.86 
- nvidia-cutlass-dsl 4.1.0

**Additional context**

LRlr239:
```text
root@ubuntu:~/workspace/cute_learn# vim 1.py
root@ubuntu:~/workspace/cute_learn# python 1.py 
a + b = 3
root@ubuntu:~/workspace/cute_learn# 
```

Using a .py script yields the usual behavior.

fengxie:
Thanks for reporting. I can reproduce with interactive mode. I think that's because we assume there is source file when attempt to get source file locations @anakinxc .

anakinxc:
Yes, we should fix that

LRlr239:
Occasionally the same issue can occur in Jupyter Notebook, but I have no way to reproduce it. Only the Python interactive shell can reproduce it consistently.

brandon-yujie-sun:
Getting source code in REPL mode is not supported. A clear error message has been added in 4.3 dev for this. @LRlr239 please let us know if you see any other issues.

LRlr239:
[...]

PR #3424, opened 2026-07-30
[CuTe DSL] Fix REPL crash: clear diagnostic and keep the interpreter alive (issue #3413)

### LINKED ISSUE #3413: [BUG] Using CuTe DSL in REPL crashes Python

CuTe DSL

### Bug Report

**Describe the bug**
Trying to use CuTe DSL in the Python REPL causes an `[Internal Error]` and crashes the interpreter.

```
$ CUTE_DSL_SHOW_STACKTRACE=1 python3
Python 3.12.3 (main, Jun 19 2026, 12:46:00) [GCC 13.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from cutlass import cute
>>> @cute.jit
... def run(f):
...     f()
...
>>> run(lambda: 0); print("ran")
  File "/usr/local/lib/python3.12/dist-packages/nvidia_cutlass_dsl/dsl_packages/cutlass/base_dsl/ast_preprocessor.py", line 764, in transform_function
    lines, start_line = inspect.getsourcelines(function_pointer)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/inspect.py", line 1260, in getsourcelines
    lines, lnum = findsource(object)
                  ^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/inspect.py", line 1089, in findsource
    raise OSError('could not get source code')
OSError: could not get source code

During handling of the above exception, another exception occurred:

  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.12/dist-packages/nvidia_cutlass_dsl/dsl_packages/cutlass/base_dsl/dsl.py", line 866, in jit_wrapper
    BaseDSL._preprocess_and_replace_code(func)
  File "/usr/local/lib/python3.12/dist-packages/nvidia_cutlass_dsl/dsl_packages/cutlass/base_dsl/dsl.py", line 833, in _preprocess_and_replace_code
    fcn_ptr = func._dsl_object.run_preprocessor(
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/nvidia_cutlass_dsl/dsl_packages/cutlass/base_dsl/dsl.py", line 2372, in run_preprocessor
    return self._run_preprocessor_impl(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/nvidia_cutlass_dsl/dsl_packages/cutlass/base_dsl/dsl.py", line 2388, in _run_preprocessor_impl
    transformed_ast = preprocessor_session.transform(
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/nvidia_cutlass_dsl/dsl_packages/cutlass/base_dsl/ast_preprocessor.py", line 1016, in transform
    transformed_tree = self.transform_function(
                       ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/nvidia_cutlass_dsl/dsl_packages/cutlass/base_dsl/ast_preprocessor.py", line 790, in transform_function
    raise DSLRuntimeError(
cutlass.base_dsl.common.DSLRuntimeError:
[Internal Error] The compiler hit a problem it could not trace back to your code.
This is a bug in the DSL, not a mistake in your kernel.

Detail: Failed to parse function run

What to do:
  Please report this with the snippet above and your kernel.
  Re-run with CUTE_DSL_SHOW_STACKTRACE=1 to include the full technical detail.

====================================================================================================
$ 
```

**Expected behavior**
I would expect either a successful run or a clear error message, whichever is the intended behavior in this case. In either case, Python shouldn't crash.

 - Environment location: [Docker] 

**Additional context**
No additional context.

layahaasini:
Fixes #3413

Calling a `@cute.jit` function defined in the REPL (or via `exec()`) causes 2 problems:

1. The "no source available" failure gets raised as `DSLRuntimeError`, rendering as "[Internal Error] This is a bug in the DSL ... please report this". It isn't a DSL bug, REPL is a known unsupported case (#2636 says a clear message was added for it in 4.3, but it doesn't reach the user because of the error class).
2. The DSL excepthook unconditionally calls `sys.exit(1)`, so the uncaught error kills the whole interactive session instead of returning to the prompt. Same thing happens when debugging a normal file-based kernel with `python -i`.

Changes:

- `ast_preprocessor.py`: catch `OSError` from `inspect.getsourcelines` separately and raise `DSLUserCodeError` with a new catalog entry `UNSUP_NO_SOURCE` ("save the function to a .py file"). Other parse failures still raise `DSLRuntimeError`, now with the cause chained.
- `common.py`: skip the `sys.exit(1)` in `_dsl_excepthook` when the session is interactive. Scripts still exit 1 on uncaught DSL errors.

This doesn't add REPL support, it just makes the failure behave as intended.

With the patch, the repro from the issue now gives:

    >>> run(lambda: 0)
    error[UNSUP_NO_SOURCE]: The source of `run` is not available (e.g. defined in the REPL or via exec()), so it cannot be compiled.
      = note: Caused exception: could not get source code
      suggestion: Save the function to a .py file and import it from there.
    >>> print("still alive")
    still alive

Tested against the 4.6.1 wheel on a DGX Spark (GB10): the two new regression tests in `test/python/CuTeDSL/test_no_source_diagnostics.py` fail on the stock wheel and pass with the patch; a third checks script exit status is unchanged. Also verified a normal file-based kernel still compiles and launches.

Note: #3396 touches the same `except` block for a different bug (stale on-disk source), small rebase likely needed

### DIFF
--- a/python/CuTeDSL/cutlass/base_dsl/ast_preprocessor.py
+++ b/python/CuTeDSL/cutlass/base_dsl/ast_preprocessor.py
@@ -785,12 +785,11 @@ class DSLPreprocessor(ast.NodeTransformer):
                         _node.col_offset += _col_shift  # type: ignore[attr-defined]
                     if getattr(_node, "end_col_offset", None) is not None:
                         _node.end_col_offset += _col_shift  # type: ignore[attr-defined]
-        except Exception:
-            # Under REPL mode, there is no way to get source of a function object, error out
-            raise DSLRuntimeError(
-                f"Failed to parse function {func_name}",
-                suggestion="DSL does not support REPL mode, save the function to a file instead.",
-            )
+        except OSError as e:
+            # No retrievable source (REPL / exec())
+            raise DSLUserCodeError(DiagId.UNSUP_NO_SOURCE, func=func_name, cause=e)
+        except Exception as e:
+            raise DSLRuntimeError(f"Failed to parse function {func_name}", cause=e)
 
         # Step 1.2 Check the decorator
         if not self.check_decorator(tree.body[0]):
--- a/python/CuTeDSL/cutlass/base_dsl/ast_preprocessor.py
+++ b/python/CuTeDSL/cutlass/base_dsl/ast_preprocessor.py
@@ -785,7 +785,7 @@ class DSLPreprocessor(ast.NodeTransformer):
                         _node.col_offset += _col_shift  # type: ignore[attr-defined]
                     if getattr(_node, "end_col_offset", None) is not None:
                         _node.end_col_offset += _col_shift  # type: ignore[attr-defined]
-        except OSError as e:
+        except (OSError, TypeError) as e:
             # No retrievable source (REPL / exec())
             raise DSLUserCodeError(DiagId.UNSUP_NO_SOURCE, func=func_name, cause=e)
         except Exception as e:
--- a/python/CuTeDSL/cutlass/base_dsl/common.py
+++ b/python/CuTeDSL/cutlass/base_dsl/common.py
@@ -99,8 +99,8 @@ def _dsl_excepthook(
         else:
             # Just print the formatted message (which is in __str__)
             print(str(exc_value), file=sys.stderr)
-        # Don't kill an interactive session (REPL, `python -i`)
-        if not (hasattr(sys, "ps1") or sys.flags.interactive):
+        # Don't kill an interactive session (REPL, `python -i`, PYTHONINSPECT)
+        if not (hasattr(sys, "ps1") or sys.flags.interactive or sys.flags.inspect):
             sys.exit(1)
     else:
         # Use the original exception hook for other exceptions
--- a/python/CuTeDSL/cutlass/base_dsl/common.py
+++ b/python/CuTeDSL/cutlass/base_dsl/common.py
@@ -99,7 +99,9 @@ def _dsl_excepthook(
         else:
             # Just print the formatted message (which is in __str__)
             print(str(exc_value), file=sys.stderr)
-        sys.exit(1)
+        # Don't kill an interactive session (REPL, `python -i`)
+        if not (hasattr(sys, "ps1") or sys.flags.interactive):
+            sys.exit(1)
     else:
         # Use the original exception hook for other exceptions
         _original_excepthook(exc_type, exc_value, exc_traceback)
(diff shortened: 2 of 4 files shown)


anakinxc:
LGTM, thanks for contributing

@brandon-yujie-sun

python/CuTeDSL/cutlass/base_dsl/common.py
```diff
@@ -99,7 +99,9 @@ def _dsl_excepthook(
         else:
             # Just print the formatted message (which is in __str__)
             print(str(exc_value), file=sys.stderr)
-        sys.exit(1)
+        # Don't kill an interactive session (REPL, `python -i`)
+        if not (hasattr(sys, "ps1") or sys.flags.interactive):
```
brandon-yujie-sun: Besides -i, can we also cover the PYTHONINSPECT=1 case which sets sys.flags.inspect?

python/CuTeDSL/cutlass/base_dsl/ast_preprocessor.py
```diff
@@ -785,12 +785,11 @@ def transform_function(
                         _node.col_offset += _col_shift  # type: ignore[attr-defined]
                     if getattr(_node, "end_col_offset", None) is not None:
                         _node.end_col_offset += _col_shift  # type: ignore[attr-defined]
-        except Exception:
-            # Under REPL mode, there is no way to get source of a function object, error out
-            raise DSLRuntimeError(
-                f"Failed to parse function {func_name}",
-                suggestion="DSL does not support REPL mode, save the function to a file instead.",
-            )
+        except OSError as e:
```
brandon-yujie-sun: Seems OSError might not be enough: https://docs.python.org/3/library/inspect.html#inspect.getsourcelines Can we also handle the TypeError?

brandon-yujie-sun (approved):
LGTM with a couple comments, and thanks for contributing the changes!

layahaasini:
Done, thanks for the feedback!

brandon-yujie-sun:
@Junkai-Wu for further proceeding.

PR #3440, opened 2026-08-08
Fix the atomicCAS retry condition in SubbyteReference::set()

### LINKED ISSUE #3428: [BUG] Broken atomicCAS RMW loop in SubbyteReference::set() for straddling elements

CUTLASS C++

### Describe the bug

The multi-storage-unit (straddling) path of `SubbyteReference::set()` in `include/cutlass/subbyte_reference.h` contains a broken atomic compare-and-swap read-modify-write loop.

Specifically, after a successful `atomicCAS` the retry condition compares the updated value against the value returned by `atomicCAS` instead of comparing the originally assumed value against the returned value. As a result, the loop may:

- Spin indefinitely after a successful CAS.
- Produce torn or incorrect writes under contention.

This affects concurrent updates to adjacent sub-byte elements whose bit ranges span two `StorageUnit`s (`high_storage_unit_idx_ != low_storage_unit_idx_`).

### Steps to reproduce

Any concurrent write that enters the multi-storage-unit branch of `SubbyteReference::set()`.

Minimal device-side sketch (packed `int4b_t` values stored in a `uint8_t` array):

```cpp
using Element = cutlass::int4b_t;
using StorageVec = cutlass::Array<uint8_t, N>;  // N sufficiently large

SubbyteReference<Element, StorageVec> ref0(ptr, offset);
SubbyteReference<Element, StorageVec> ref1(ptr, offset + 1);  // Adjacent element crossing a StorageUnit boundary

// Executed concurrently from different threads:
ref0.set(value0);
ref1.set(value1);
```

The affected code currently follows this pattern:

```cpp
do {
  original_low_bits = ((*ptr_)[low_storage_unit_idx_]);
  update_low_bits = (original_low_bits & kLowUpdateMask) | low_new_bits;
  original_low_bits = atomicCAS(
      &((*ptr_)[low_storage_unit_idx_]),
      original_low_bits,
      update_low_bits);
} while (update_low_bits != original_low_bits);
```

The identical issue exists in the corresponding high-half CAS loop.

### Expected behavior

The CAS loop should follow the standard read-modify-write retry pattern already used elsewhere in the same file (including the single-storage-unit path and the previously fixed `ConstSubbyteReference` implementation):

```cpp
Storage original = *ptr;
Storage assumed;

do {
  assumed = original;
  Storage updated = (assumed & mask) | new_bits;
  original = atomicCAS(ptr, assumed, updated);
} while (original != assumed);
```

A successful CAS should exit the loop immediately while a failed CAS caused by a concurrent modification should simply retry.

### Environment

- **Environment:** Bare metal
- **CUDA architectures:** Any architecture taking the `__CUDA_ARCH__` path
- **Observed on:** Recent `main` branch and 4.5.x–4.6.x releases

### Additional context

- The single-storage-unit implementation already uses the correct CAS retry pattern.
- The `ConstSubbyteReference` implementation was previously fixed to use the same correct algorithm (referenced in the 4.5.x changelog).

However, both CAS retry loops in this path still use the incorrect termination condition.

- Because the bug only manifests when multiple threads concurrently update packed sub-byte elements that span a `StorageUnit` boundary, it is difficult to reproduce with conventional unit tests and is more likely to surface in real packed-tensor workloads involving multiple threads or CTAs.

sunnypatell:
`SubbyteReference::set()`'s device path is a compare-and-swap loop over a storage unit shared with neighbouring elements. the multi-storage-unit specialization exits on `updated != cas_return`, which is not the read-modify-write retry contract: [`atomicCAS` returns the value that was already in memory](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#atomiccas), so a CAS that successfully changes the slot returns the *old* value and never the new one. the loop therefore does not stop when it succeeds. it goes around again, re-reads the slot, recomputes the same bits, and issues a second CAS purely to observe its own write. every store through this path costs at least two atomics, and under contention it keeps re-entering a race it has already won.

this is the same defect that was fixed for `ConstSubbyteReference` in 4.5 ("Fix atomicCAS read-modify-write loop in `ConstSubbyteReference`" in the [changelog](https://github.com/NVIDIA/cutlass/blob/main/CHANGELOG.md)), and the sibling specialization higher up in this same file already uses the correct `assumed` / `original` form. the three loops in the multi-storage-unit specialization were missed. closes #3428.

the fix is the standard idiom: seed `original` once, then reuse the CAS return as the next `assumed`.

### which types this actually touches

this specialization is selected when `sizeof_bits<StorageUnit> % sizeof_bits<Element> != 0`, which in practice means the 6-bit types, `float_e2m3_t` and `float_e3m2_t`. it is the scalar element store reached through `TensorRef::Reference`, not the tensor-core mainloop, so the blast radius is element-wise access on packed fp6 tensors rather than GEMM inner loops.

### numbers

measured through the real `cutlass::SubbyteReference<float_e2m3_t>::set()` path rather than a replica, by compiling one benchmark against the upstream and the patched include tree. rtx 5070 ti (sm_120), cuda 13.0.3, driver 610.74, medians of 9 runs:

| | upstream | patched |
| --- | --- | --- |
| contended stores | 105.0 ms `[104.2, 106.1]` | 44.6 ms `[44.3, 45.3]` |
| disjoint stores | 0.12 ms | 0.08 ms |

2.35x on the contended case, with non-overlapping ranges. clocks were not locked on this part, which is why medians and ranges are reported instead of a single number. a separate isolated reproduction of the two loop forms counts the CAS operations directly: 2 versus 1 for a single uncontended store, and a mean of 268 versus 1.98 with 4096 threads on one field.

### what the new tests prove, and what they do not

`test/unit/core/subbyte_reference.cu` is new; there was no coverage for this class at all. it walks every offset in a container so both the straddling and the single-unit branch are exercised deterministically rather than by scheduling luck, and it runs one thread per element so every store races its neighbours.

being straight about it: these tests pass on unpatched main as well. the old loop does converge, it just pays for it, so no correctness test can catch this. the tests are here to hold the store contract while the loop is rewritten, and the measurements above are what justify rewriting it.

### verified

- 505/505 `cutlass_test_unit_core` green on sm_120a, both before and after the header change
- the host path (`#else`) and the sibling specialization are untouched
- diff is functional only: no copyright header edits, no reformatting, no whitespace churn

@hwu36 flagging for routing since you've taken the recent core C++ correctness fixes (#3427, #3244, #3048). happy to add coverage elsewhere or re-measure with locked clocks if that would help.

### DIFF
--- a/include/cutlass/subbyte_reference.h
+++ b/include/cutlass/subbyte_reference.h
@@ -815,29 +815,30 @@ public:
     if(high_storage_unit_idx_ != low_storage_unit_idx_){
       /// Only need update 2 storage unit at once.
       /// consider misaligned address issue, we need to do atomicCAS twice 
-      StorageUnit original_low_bits, original_high_bits, update_low_bits, update_high_bits;
+      StorageUnit assumed;
+      StorageUnit original = ((*ptr_)[low_storage_unit_idx_]);
       do {
-        original_low_bits  = ((*ptr_)[low_storage_unit_idx_]);
-        update_low_bits  = (original_low_bits & kLowUpdateMask) | low_new_bits;
-        original_low_bits = atomicCAS(&((*ptr_)[low_storage_unit_idx_]), original_low_bits, update_low_bits);
-      } while (update_low_bits != original_low_bits);
+        assumed = original;
+        StorageUnit updated = StorageUnit((assumed & kLowUpdateMask) | low_new_bits);
+        original = atomicCAS(&((*ptr_)[low_storage_unit_idx_]), assumed, updated);
+      } while (original != assumed);
+
+      original = ((*ptr_)[high_storage_unit_idx_]);
       do {
-        original_high_bits = ((*ptr_)[high_storage_unit_idx_]);
-        update_high_bits  = (original_high_bits & kHighUpdateMask) | high_new_bits;
-        original_high_bits = atomicCAS(&((*ptr_)[high_storage_unit_idx_]), original_high_bits, update_high_bits);
-      } while (update_high_bits != original_high_bits);
+        assumed = original;
+        StorageUnit updated = StorageUnit((assumed & kHighUpdateMask) | high_new_bits);
+        original = atomicCAS(&((*ptr_)[high_storage_unit_idx_]), assumed, updated);
+      } while (original != assumed);
     }
     else {
       /// Only need update 1 storage unit.
-      StorageUnit original, updated;
+      StorageUnit assumed;
+      StorageUnit original = ((*ptr_)[low_storage_unit_idx_]);
       do {
-        original = ((*ptr_)[low_storage_unit_idx_]);
-
-        updated = (original & kLowUpdateMask) | low_new_bits;
-
-        original = atomicCAS(&((*ptr_)[low_storage_unit_idx_]), original, updated);
-
-      } while (updated != original);
+        assumed = original;
+        StorageUnit updated = StorageUnit((assumed & kLowUpdateMask) | low_new_bits);
+        original = atomicCAS(&((*ptr_)[low_storage_unit_idx_]), assumed, updated);
+      } while (original != assumed);
     }
 #else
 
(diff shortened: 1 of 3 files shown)


hwu36:
There is an issue about this which is fixed internally already. @jackkosaian

sunnypatell:
closing. i'll pick it up from the release.

PR #3446, opened 2026-08-11
[CuTeDSL] Skip the numpy round-trip for in-range integer scalars

### LINKED ISSUE #3443: [BUG][CuTe DSL] Constructing cutlass Numeric scalars from Python values costs ~0.3-0.8us each

CuTe DSL

### Bug Report

**Describe the bug**
Calling `cutlass.Int32(val)` costs ~750ns and `cutlass.Float32(val)` costs ~270ns. This is quite noticeable, on the order of kernel launch latency itself. `Integer.__init__` (cutlass/base_dsl/typing.py) converts a plain Python int through a numpy round-trip: `x_val = int(np.array(x).astype(np_dtype))` which alone costs ~400us.

```python
import timeit

import numpy as np

import cutlass

INT32_MIN, INT32_MAX = -(2**31), 2**31 - 1


def numpy_path(x: int) -> int:
    # The conversion Integer.__init__ performs today.
    return int(np.array(x).astype(np.int32))


def fast_path(x: int) -> int:
    # Range check first; fall back to numpy only for wrap-around.
    if INT32_MIN <= x <= INT32_MAX:
        return int(x)  # int() normalizes bool -> int, matching the numpy path
    return numpy_path(x)


def check_equivalence():
    edge_cases = [
        0, 1, -1, 5, True, False,
        INT32_MIN, INT32_MAX, INT32_MIN - 1, INT32_MAX + 1,
        2**31, 2**32, -(2**32), 2**63 - 1, 123456789,
    ]
    for x in edge_cases:
        assert fast_path(x) == numpy_path(x), x
    print(f"fast path matches numpy path on {len(edge_cases)} edge cases")


def bench(label, fn, n=100_000):
    t = min(timeit.repeat(fn, number=n, repeat=5)) / n
    print(f"  {label:40s} {t * 1e9:7.0f} ns")


def main():
    check_equivalence()
    print("per-construction cost:")
    bench("cutlass.Int32(5)", lambda: cutlass.Int32(5))
    bench("cutlass.Int64(5)", lambda: cutlass.Int64(5))
    bench("cutlass.Float32(0.5)", lambda: cutlass.Float32(0.5))
    bench("int(np.array(5).astype(np.int32))", lambda: numpy_path(5))
    bench("range-checked fast path", lambda: fast_path(5))

    # Context: a realistic launch prologue wrapping 8 scalar kernel arguments.
    scalars = [1, 8, 128, 576, 4, 32, 1024, 7]
    bench(
        "8x Int32 (typical launch prologue)",
        lambda: [cutlass.Int32(s) for s in scalars],
        n=20_000,
    )


if __name__ == "__main__":
    main()
```

**Expected behavior**
These casts shouldn't take this long. A workaround is to define, e.g.,
```python
@lru_cache
def _int32(val: int) -> cutlass.Int32:
	return cutlass.Int32(val)
```

 - `nvidia-cutlass-dsl 4.6.0.dev0` and `4.7.0`, no GPU needed.

wilyan09007:
Fixes #3443.

`Integer.__init__` sends every Python scalar through `int(np.array(x).astype(np_dtype))`, and the float path calls `np.isnan`/`np.isinf` on a Python float before that. The cast is there to reproduce C wrap-around, but it is an identity whenever the value already fits the target dtype, which is the usual case for kernel arguments. This range checks first and keeps the cast for the values that really do wrap.

`Float.__init__` already dropped its numpy conversion (the replaced code is still commented in place at the top of that method), so this is the same treatment applied to the integer side. There is nothing equivalent left to remove on the `Float32` path, and its cost is unchanged here.

### Numbers

CPython 3.12.3, numpy 2.5.2, `nvidia-cutlass-dsl` 4.7.0, no GPU. Minimum of 5 runs of 100k constructions each:

| construction | before | after | |
| --- | ---: | ---: | ---: |
| `Int32(5)` | 1165 ns | 607 ns | 1.9x |
| `Int64(5)` | 1056 ns | 651 ns | 1.6x |
| `Int32(3.7)` | 2450 ns | 693 ns | 3.5x |
| `Boolean(1)` | 1281 ns | 805 ns | 1.6x |
| 8x `Int32`, a launch prologue | 10128 ns | 4978 ns | 2.0x |
| `Int32(2**32)`, wraps | 1170 ns | 1268 ns | 0.9x |

The last row is the tradeoff: a value that wraps now pays one extra comparison before reaching the cast, roughly 100 ns against the 1.2 us the cast already costs.

### Where the bounds come from

`IntegerMeta` computes the range once per type at class creation, from `np.iinfo` on the dtype, and stores it as `_exact_range`. It deliberately follows the numpy dtype rather than `width`, because the cast is what it has to agree with and the two do not always match:

- `Boolean` is declared signed at width 1, which puts `min`/`max` at -1/0 while the dtype is `np.bool_`. It is special-cased to (0, 1), the only two values `astype(np.bool_)` leaves alone.
- `WgmmaSmemDesc` and `Tcgen05SmemDesc` assign `width` after the class is built, so their `min`/`max` describe 64 bits while they still cast through `int32`. Keying the check off `min`/`max` would make `WgmmaSmemDesc(2**40)` store 1099511627776 instead of the 0 the cast produces.

`min` and `max` themselves are untouched.

### Behavior

Unchanged, including the parts that are arguably surprising:

- Out-of-range values keep numpy's wrap-around (`Int8(128)` is -128, `Uint8(-1)` is 255), and the `OverflowError` numpy raises for integers too wide for any C type.
- `Int4`, `Int128` and `Uint128` have no numpy dtype and still trip the same assert. The shortcut sits behind that guard rather than in front of it, so they do not quietly gain support.
- NaN and infinity still raise `ValueError` and `OverflowError` with the same messages, except for `Boolean`, which has always folded them to 1 because it runs `bool()` before `Integer.__init__` sees the value.

Checked by snapshotting every integer and float type against 57 inputs (boundaries, wrap-around candidates, bools, floats, NaN and both infinities) plus construction from other DSL scalars, recording either the stored value or the exact exception, and diffing the run before the change against the run after: 1505 cases, no differences.

### Tests

`test/python/CuTeDSL/test_numeric_construction.py` is new, 14 tests, no GPU or compilation needed. The broadest one asserts the constructor against `int(np.array(x).astype(dtype))` directly, so the claim this PR rests on is the thing under test. The rest pin the wrap-around, the truncation direction, the NaN and infinity handling, the types with no numpy dtype, and the dtype-versus-width case above.

I also ran them against four deliberately broken variants to confirm they fail: bounds keyed off `min`/`max`, the range guard removed, the `Boolean` range widened, and the NaN check dropped. Each is caught.

`test/python/CuTeDSL/test_struct_in_if.py` passes before and after (6 tests, real `cute.compile`).

### DIFF
--- a/python/CuTeDSL/cutlass/base_dsl/typing.py
+++ b/python/CuTeDSL/cutlass/base_dsl/typing.py
@@ -10,6 +10,7 @@
 # is strictly prohibited.
 
 import ctypes
+import math
 from abc import abstractmethod
 from itertools import chain
 import numpy as np
@@ -551,6 +552,9 @@ class IntegerMeta(NumericMeta):
     """
 
     signed: bool
+    # Value range that ``_np_dtype`` stores exactly, or None when there is no
+    # numpy dtype to store into. See ``Integer.__init__``.
+    _exact_range: Optional[tuple[int, int]]
 
     def __new__(
         cls,
@@ -590,6 +594,19 @@ class IntegerMeta(NumericMeta):
             cls, name, bases, attrs | new_attrs, width, np_dtype, mlir_type, is_abstract
         )
         new_cls.signed = signed
+        # Precomputed once per type so ``Integer.__init__`` can range-check without
+        # rebuilding the bounds on every construction. These track ``np_dtype``
+        # rather than ``width``, because the numpy cast is what they must agree
+        # with, and a few subclasses (WgmmaSmemDesc) patch ``width`` afterwards.
+        if np_dtype is None:
+            new_cls._exact_range = None
+        elif width == 1:
+            # np.bool_ folds everything nonzero to True, so only 0 and 1 survive
+            # the cast unchanged.
+            new_cls._exact_range = (0, 1)
+        else:
+            info = np.iinfo(np_dtype)
+            new_cls._exact_range = (int(info.min), int(info.max))
         return new_cls
 
     def __str__(cls) -> str:
@@ -1718,6 +1735,7 @@ class Integer(Numeric, metaclass=IntegerMeta, mlir_type=T.i32, is_abstract=True)
 
     # Injected by IntegerMeta.__new__ on every concrete subclass.
     signed: ClassVar[bool]
+    _exact_range: ClassVar[Optional[tuple[int, int]]]
 
     def __init__(
         self,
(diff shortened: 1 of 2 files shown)


anakinxc:
LGTM

anakinxc:
@Junkai-Wu The fix has landed internally, can you help merge this one? Thanks

Junkai-Wu (approved):

PR #3448, opened 2026-08-12
[CuTeDSL] Fix export_to_c shape slot width for 64-bit dynamic dims

### LINKED ISSUE #3447: [BUG] cuteDSL export_to_c outputs i32 dynamic shapes in headers for i64 inputs

CuTe DSL

### Bug Report

**Describe the bug**

If one passes a 64b dynamic shape, then compile & export_to_c, the generated header produces `int32_t`  dynamic shapes instead of the expected `int64_t`

```
  """export_to_c emits int32_t shape slots even for cute.sym_int64() shapes.

  Run: CUTE_DSL_ARCH=sm_100a python repro_minimal.py
  """

  import os
  import re
  import tempfile

  import cutlass
  import cutlass.cute as cute
  from cuda.bindings import driver as cuda
  from cutlass import Float32


  @cute.kernel
  def k(mA: cute.Tensor):
      pass


  @cute.jit
  def f(mA: cute.Tensor, stream: cuda.CUstream):
      k(mA).launch(grid=[1, 1, 1], block=[1, 1, 1], stream=stream)


  print(f"cutlass-dsl {cutlass.__version__}")
  for name in ("sym_int32", "sym_int64"):
      sym = getattr(cute, name)
      # Both dims dynamic, so the header struct has two shape slots.
      a = cute.runtime.make_fake_tensor(
          Float32, (sym(), sym()), stride=(cute.sym_int64(), 1)
      )
      c = cute.compile(f, a, cute.runtime.make_fake_stream())

      out = tempfile.mkdtemp()
      c.export_to_c(file_path=out, file_name="p", function_prefix="p")
      header = open(os.path.join(out, "p.h")).read()

      ciface = next(
          l for l in str(c.ir_module).splitlines() if "llvm.func @_mlir_ciface" in l
      )
      print(f"\n{name}:")
      print("  ABI   :", ciface[ciface.index("%arg0:") : ciface.index("i64)>)>") + 8])
      print("  header:", re.search(r"int\d+_t dynamic_shapes\[\d+\]", header).group(0))
```

**Expected behavior**

```
  sym_int32:
    ABI   : %arg0: !llvm.struct<(ptr<1>, struct<(struct<(i32, i32)>, i64)>)>,
    header: int32_t dynamic_shapes[2]

  sym_int64:
    ABI   : %arg0: !llvm.struct<(ptr<1>, struct<(struct<(i64, i64)>, i64)>)>,
    header: int32_t dynamic_shapes[2]
```

Bare metal, nvidia-cutlass-dsl==4.5.2, 4.7.0

wilyan09007:
## Summary

`_generate_arguments` derives a tensor descriptor's stride type from the tensor but passes `shape_type="int32_t"` unconditionally, so `cute.sym_int64()` dimensions get `int32_t` shape slots while the compiled function reads them as `i64`. A C caller filling in that struct hands over a descriptor half the expected size.

The widths are already known one layer down: `_tensor_binding` keeps a `SymInt` dimension at its declared width and gives every other dynamic dimension 32 bits. This derives the shape type the same way at both call sites. One array cannot hold dimensions of two widths, so a tensor mixing them is refused, and the list/tuple descriptor check now compares width alongside the masks it already had.

## Verification

Added `test/python/CuTeDSL/test_export_c_header_shape_type.py`, 8 tests. Four fail on main: the `sym_int64` slot width, the mixed-width rejection, and both tensor-list cases. The `sym_int32`, static-shape, and DLPack-dynamic cases pass before and after.

Fixes #3447

### DIFF
--- a/python/CuTeDSL/cutlass/cute/export/c_header_generator.py
+++ b/python/CuTeDSL/cutlass/cute/export/c_header_generator.py
@@ -9,7 +9,7 @@
 # and related documentation outside the scope permitted by the EULA
 # is strictly prohibited.
 
-from cutlass.cute.typing import NumericMeta, Integer
+from cutlass.cute.typing import NumericMeta, Integer, SymInt
 from cutlass.base_dsl.export import CHeaderGenerator, CHeaderArguments
 from cutlass.base_dsl.common import DSLRuntimeError
 from cutlass.base_dsl.jit_executor import ExecutionArgs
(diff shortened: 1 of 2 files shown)


anakinxc:
LGTM

Observer007 (approved):
LGTM

Observer007:
Just curious, in what real cases do we need i64 shape representation?

wilyan09007:
A genuinely 64-bit extent shows up on a flat 1-D view over a whole buffer: `red.py` and `reduction/block_smem_reduce.py` both declare `(cute.sym_int64(),)`, and 2^31 elements is 2 GB of fp8. For GEMM M/N/K it is rare.

The corruption doesn't need a large value though. Shape and stride slots share one descriptor struct, so int32 shapes against an i64 ABI both narrow the shape read and shift the strides. M=128 breaks the same way a 3e9 dim would.

anakinxc:
@Junkai-Wu The fix has landed internally, can you help merge this one? Thanks

Junkai-Wu (approved):

Issue #3462, opened 2026-08-14
[BUG] CuTe DSL 4.7.0: import-time CUTE_DSL_LIBS env write breaks Slurm jobs on other machines


sshleifer:
**Component:** CuTe DSL
**Version:** nvidia-cutlass-dsl 4.7.0 (regression vs 4.5.3, which does not touch the environment)

## Bug

**This breaks Slurm-style clusters where the submitter's environment is forwarded to jobs on other machines** (Slurm `sbatch` defaults to `--export=ALL`). Two new behaviors in 4.7.0 combine:

**1. `import cutlass` mutates `os.environ`.** `_select_and_load_cutlass_ir_toolkit` (`cutlass/_mlir/_mlir_libs/__init__.py`) prepends the absolute path of the chosen `libcute_dsl_runtime.so` — inside the importing interpreter's site-packages — to `CUTE_DSL_LIBS`:

```console
$ python -c "import os, cutlass; print(os.environ.get('CUTE_DSL_LIBS'))"
/path/to/submitter/venv/site-packages/nvidia_cutlass_dsl/cu12/lib/libcute_dsl_runtime.so
```

A launcher that merely imports `cutlass` while preparing a job therefore ships a submitter-machine-local path to every worker.

**2. Workers hard-fail on the stale entry, despite having a valid runtime.** On the worker, `import cutlass` prepends its own (valid) path — but `get_shared_libs` (`cutlass/base_dsl/dsl.py`) raises on the first missing entry, so every `cute.compile` dies even though the first entry in the list is a good local runtime:

```console
$ CUTE_DSL_LIBS=/nonexistent/libcute_dsl_runtime.so python - <<'EOF'
import cutlass, cutlass.cute as cute
from cutlass import Int32

@cute.jit
def trivial(a: Int32):
    pass

cute.compile(trivial, Int32(0))
EOF
FileNotFoundError: [Errno 2] No such file or directory: '/nonexistent/libcute_dsl_runtime.so'
```

Reproduced on GB300 (aarch64, Python 3.13, split wheels `-libs-core`/`-libs-cu12` 4.7.0). Upgrading 4.5.3 → 4.7.0 took down two independent launch pipelines on our Slurm cluster in one night; the failing jobs never referenced `CUTE_DSL_LIBS` themselves.

## Expected behavior

- Importing a library shouldn't write machine-local absolute paths into the process environment — keep the chosen runtime path in module state instead.
[...]

Issue #3329, opened 2026-06-18
[BUG] Unhelpful error when cutedsl runtime shared libs aren't found


slayton58:
### Which component has the problem?

CuTe DSL

### Bug Report

**Describe the bug**

If cutedsl cannot auto-resolve the location of the cutedsl runtime `.so` (or potentially auto-resolves to a version that is not compatible with the python package), and `CUTE_DSL_LIBS` is not set (or set inappropriately), code fails -- this part is fine, it should fail, but the error message ends up as something like:

```
RuntimeError: Unknown function cutlass_add_one_FakeTensorFloat322561_FakeTensorFloat322561_FakeStream
```

Which doesn't indicate anything about the actual problem. Now, prior to this, you should see:

```
JIT session error: Symbols not found: [ cuda_dialect_unload_library_once, cuda_dialect_init_library_once, cuda_dialect_get_error_name, _cuKernelGetAttribute, _cudaDeviceGetAttribute, _cudaFuncSetAttribute, _cudaGetDevice, _cudaKernelSetAttributeForDevice, _cudaLaunchKernelEx, _cudaLibraryGetKernel, _cudaLibraryLoadData, _cudaSetDevice ]
```

But this doesn't seem to be a hard-error, instead it continues and fails elsewhere. This made a debug of the above `RuntimeError` in production workflows significantly harder by leading us in a very (very) wrong direction.

**Steps/Code to reproduce bug**
Run the following script with `CUDA_DSL_LIBS` set to something silly.
```
import traceback

import cutlass
import cutlass.cute as cute
from cutlass import Float32

print("cutlass version:", getattr(cutlass, "__version__", "?"))


@cute.kernel
def _add_one_kernel(src: cute.Tensor, dst: cute.Tensor):
    tidx, _, _ = cute.arch.thread_idx()
    if tidx < src.shape[0]:
        dst[tidx] = src[tidx] + Float32(1.0)


@cute.jit
def add_one(src: cute.Tensor, dst: cute.Tensor, stream):
    _add_one_kernel(src, dst).launch(grid=(1, 1, 1), block=(256, 1, 1), stream=stream)


def main():
    src = cute.runtime.make_fake_tensor(Float32, (256,), stride=(1,), assumed_align=4)
    dst = cute.runtime.make_fake_tensor(Float32, (256,), stride=(1,), assumed_align=4)
    stream = cute.runtime.make_fake_stream(use_tvm_ffi_env_stream=True)
    try:
[...]

PR #3466, opened 2026-08-16
[CuTeDSL] Skip unusable CUTE_DSL_LIBS entries instead of failing the compile

yentur:
`get_shared_libs` raises on the first missing entry of `CUTE_DSL_LIBS`, so one stale path fails the compile even when a valid runtime sits earlier in the same list. Reported by @sshleifer in #3462.

@anakinxc said on that issue that a fix is already in progress, so this may well duplicate work already done. It covers only the part that lives in this repository, and I am glad to close it if it overlaps with yours. The test may still be worth keeping either way.

Partial fix for #3462.

## Root cause

`get_prefix_dsl_libs` returns `{prefix}_LIBS` verbatim whenever it is set, which also means auto-discovery never runs. `get_shared_libs` then treats that list as all or nothing, so any single unusable entry is fatal.

## Change

Missing entries are skipped with a warning. If no entry is usable, auto-discovery is retried before giving up, and `FileNotFoundError` is raised only when that also finds nothing.

To make the retry possible the discovery block is split out of `get_prefix_dsl_libs` into `discover_dsl_libs`, which does not consult the environment. `get_prefix_dsl_libs` keeps its previous behaviour and signature.

The hard failure added for #3329 is preserved. When no runtime can be found, the same `FileNotFoundError` is raised from the same place with a concrete path, rather than deferring to a downstream symbol error. `test_raises_when_no_usable_runtime_is_found` pins that.

## Reproduction

CPU-only container, no GPU:

```
docker run -d --name cutlass-repro --platform linux/arm64 -v $PWD:/repo -w /work python:3.12-slim sleep infinity
docker exec cutlass-repro pip install nvidia-cutlass-dsl==4.7.0
```

`repro.py` imports cutlass with a stale value already in the environment, as a Slurm worker would inherit it, then compiles a trivial jit function. `import cutlass` prepends the worker's own valid runtime, so a good path is first in the list.

Before:

```
CUTE_DSL_LIBS seen by compile = /usr/local/.../cu12/lib/libcute_dsl_runtime.so:/nonexistent/submitter-venv/libcute_dsl_runtime.so
  exists=True  /usr/local/.../cu12/lib/libcute_dsl_runtime.so
  exists=False /nonexistent/submitter-venv/libcute_dsl_runtime.so
RESULT: FileNotFoundError: [Errno 2] No such file or directory: '/nonexistent/submitter-venv/libcute_dsl_runtime.so'
```

After:

```
UserWarning: Ignoring missing CUTE_DSL_LIBS entries: /nonexistent/submitter-venv/libcute_dsl_runtime.so. Using /usr/local/.../cu12/lib/libcute_dsl_runtime.so.
RESULT: compiled OK
```

An unset or fully valid `CUTE_DSL_LIBS` is unaffected.

## Tests

`test/python/CuTeDSL/test_shared_libs_env.py`, five cases, following `test_struct_in_if.py` in the same directory.

```
python -m unittest discover -s test/python/CuTeDSL -p 'test_*.py'
Ran 11 tests in 0.252s

OK
```

That count includes the existing `test_struct_in_if.py`. Reverting only the two source files and keeping the test leaves 4 of the 5 failing, including `test_stale_entry_is_skipped_when_a_valid_one_remains` with the production `FileNotFoundError`. `test_valid_entries_are_returned_unchanged` passes either way and guards against regressing the normal path.

## What this does not fix

The other half of #3462 is the import-time write in `_select_and_load_cutlass_ir_toolkit`, which prepends an absolute site-packages path into `os.environ["CUTE_DSL_LIBS"]`. That code is in `cutlass/_mlir/_mlir_libs/__init__.py`, which is not in this repository and never has been. `git log --all -- python/CuTeDSL/cutlass/_mlir` is empty, and `prep_editable_install.py` obtains the directory by downloading and unpacking the published wheel. I confirmed the write is present in the installed 4.7.0 wheel but had no way to change it here.

So a launcher that imports cutlass still exports a submitter-local path to its workers. This change stops that from being fatal on the receiving side, which is the part that takes jobs down.

## Notes

Tested in a CPU-only linux/arm64 container, so no GPU execution path was exercised. `cute.compile` does run without a device, so the failure and the fix are both reachable end to end through the documented reproduction rather than only at unit level. `get_shared_libs` itself is path handling and does not touch the driver.

The installed `base_dsl/dsl.py` and `base_dsl/env_manager.py` in the 4.7.0 wheel are byte identical to the ones in this tree, so the container exercises these sources.

`black` leaves the changed regions alone. It does want to reformat other parts of both files, but it wants the identical hunks at `HEAD`, so I left those alone.

### DIFF
--- a/python/CuTeDSL/cutlass/base_dsl/dsl.py
+++ b/python/CuTeDSL/cutlass/base_dsl/dsl.py
@@ -54,7 +54,12 @@ import warnings
 import threading
 
 from . import typing as t
-from .env_manager import EnvironmentVarManager, dump_sass, is_cutlass_family_dsl_prefix
+from .env_manager import (
+    EnvironmentVarManager,
+    discover_dsl_libs,
+    dump_sass,
+    is_cutlass_family_dsl_prefix,
+)
 from .compiler import CompileOptions, CompilerDiagnosticError, LinkLibraries
 from .ast_helpers import DSLOptimizationWarning
 from .common import DSLRuntimeError, active_env_manager
@@ -1823,13 +1828,35 @@ class BaseDSL(metaclass=DSLSingletonMeta):
         shared_libs = []
         support_libs = self.envar.shared_libs
         if support_libs is not None:
-            _libs = support_libs.split(":")
-            for lib in _libs:
-                if not os.path.exists(lib):
+            missing = []
+            for lib in support_libs.split(":"):
+                if not lib:
+                    continue
+                if os.path.exists(lib):
+                    shared_libs.append(lib)
+                else:
+                    missing.append(lib)
+
+            if missing and not shared_libs:
+                # Every entry is stale. This happens when {prefix}_LIBS is
+                # inherited from a machine that does not share this
+                # filesystem, e.g. Slurm's default --export=ALL. Retry
+                # auto-discovery so a valid local runtime is still found.
+                discovered = discover_dsl_libs(self.name)
+                if discovered:
+                    shared_libs = [
+                        lib for lib in discovered.split(":") if os.path.exists(lib)
+                    ]
+
+            if missing:
+                if not shared_libs:
                     raise FileNotFoundError(
-                        errno.ENOENT, os.strerror(errno.ENOENT), lib
+                        errno.ENOENT, os.strerror(errno.ENOENT), missing[0]
                     )
-                shared_libs.append(lib)
+                self.print_warning(
+                    f"Ignoring missing {self.name}_LIBS entries: "
+                    f"{', '.join(missing)}. Using {':'.join(shared_libs)}."
+                )
         else:
             if is_cutlass_family_dsl_prefix(self.name):
                 self.print_warning(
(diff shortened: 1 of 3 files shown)


yentur:
Closing this: @anakinxc said they were already working on it and I missed that before opening. Sorry for the noise.

Two things I established while reproducing, in case either saves time.

The env-write half is not fixable from this repo. `python/CuTeDSL/cutlass/_mlir/` has never been tracked (`git log --all` on that path is empty) and `prep_editable_install.py` obtains it by downloading and unpacking the published wheel. The prepend lives in the wheel at `cutlass/_mlir/_mlir_libs/__init__.py:179-183`.

`get_shared_libs` is not the only blocker for the fallback the reporter asked for. `get_prefix_dsl_libs` returns `{prefix}_LIBS` verbatim whenever it is set, which short-circuits auto-discovery before `get_shared_libs` ever runs, so skipping bad entries there is not enough on its own.

Reproduced end to end in a CPU-only linux/arm64 container on nvidia-cutlass-dsl 4.7.0; `cute.compile` does run without a GPU, and the wheel's `base_dsl/dsl.py` is byte-identical to the checkout.

anakinxc:
Hi @yentur 

Thanks for taking a look. The fix is the other way

- `cutlass/_mlir/_mlir_libs/__init__.py` should not modify env to communicate with other parts of the codebase in the first place.
- `get_prefix_dsl_libs` should still error out if user sets `CUTE_DSL_LIBS` to a non-exist path.

I'll take some of tests you added here. :)

PR #3479, opened 2026-08-20
fix(pycute): logical_product with an int tiler computed logical_divide

### LINKED ISSUE #3478: [BUG] pycute: logical_product with an integer tiler computes logical_divide
**Describe the bug**

In `python/pycute/layout.py`, `logical_product` dispatches an integer tiler to `logical_divide`, so it returns a divide result under a product name. The `Layout`-wrapped form is unaffected, so the two spellings of the same call disagree.

`python/pycute/layout.py:312-316`:

```python
def logical_product(layoutA, layoutB):
  if layoutB is None:
    return layoutA
  elif is_int(layoutB):
    return logical_divide(layoutA, Layout(layoutB))   # <-- should be logical_product
```

The identical line at `python/pycute/layout.py:301` is correct: it belongs to `logical_divide` directly above.

The C++ reference does the right thing at `include/cute/layout.hpp:1670`:

```cpp
} else if constexpr (is_integral<Tiler>::value) {
    return logical_product(block, make_layout(tiler));
}
```

```
cd python
python -c "
from pycute import *
A = Layout((2,5),(5,1))
print('product Layout(4):', logical_product(A, Layout(4)))
print('product int 4    :', logical_product(A, 4))
print('divide  int 4    :', logical_divide(A, 4))
"
```

Output:

```
product Layout(4): ((2, 5), 4):((5, 1), 10)     <- correct, matches C++
product int 4    : ((2, 2), 3):((5, 1), 2)      <- wrong
divide  int 4    : ((2, 2), 3):((5, 1), 2)      <- identical to the line above
```

The size is the clearest tell: `size(A)` is 10, the product should be 40, and the returned layout has size 12.

**Expected behavior**

`logical_product(A, 4)` should equal `logical_product(A, Layout(4))`.

**Scope**

Everything reaching the integer branch is affected:

| Call | Affected |
|---|---|
| `logical_product(A, 4)` | yes |
| `logical_product(A, (4,2))` | yes, the tuple branch recurses per mode into the int branch |
| `zipped_product(A, 4)` | yes, `hier_unzip(logical_product, ...)` bottoms out there |
| `tiled_product(A, 4)` | yes, built on `zipped_product` |
| `logical_product(A, Layout(4))` | no |

`operators/cutlass/operators/fusion/pycute/layout.py:339` is a copy of the same file and has the same line.

There is currently no test for `logical_product` or `logical_divide` in `test/python/pycute/`.

**Environment details**

- Environment location: Bare-metal
- Pure Python, no GPU, CUDA or build needed
- Reproduced on `main` at the time of filing

**Additional context**

Note this is `python/pycute`, the Python reference implementation, not the CuTe DSL. The issue-form dropdown has no option for it.

I have a fix and a regression test ready and will open a PR referencing this issue.

arcusbuilds:
Fixes #3478.

`logical_product(layout, <int>)` called `logical_divide`, so it returned a divide result. Recursing with `Layout(tiler)` reaches the correct branch, which is what the C++ reference does at `include/cute/layout.hpp:1670`.

Before:

```
product Layout(4): ((2, 5), 4):((5, 1), 10)
product int 4    : ((2, 2), 3):((5, 1), 2)
```

After:

```
product Layout(4): ((2, 5), 4):((5, 1), 10)
product int 4    : ((2, 5), 4):((5, 1), 10)
```

The tuple-tiler, `zipped_product` and `tiled_product` paths bottom out in the same branch and are fixed with it. After the change the integer path delegates to the `Layout` path, so the two forms cannot diverge again.

Adds `test/python/pycute/test_product.py`. There was no test for `logical_product` before. It ports the post-conditions from `test/unit/cute/core/logical_product.cpp` over the same cases, and pins the integer and tuple forms against their `Layout` equivalents. `run_all_tests.py` discovers it automatically; the suite goes from 10 to 14 tests.

The nine `Layout`-tiler cases in `test_logical_product` pass both before and after the fix, since a `Layout` tiler skips the integer branch. They are the control: the other three tests failing while that one stays green shows the failure is the integer path and not the test file.

The second commit applies the same one-word fix to `operators/cutlass/operators/fusion/pycute/layout.py`, a copy of the same module. It is kept separate. Say the word and I will drop it if that tree is regenerated from an internal source.

### DIFF
--- a/operators/cutlass/operators/fusion/pycute/layout.py
+++ b/operators/cutlass/operators/fusion/pycute/layout.py
@@ -336,7 +336,7 @@ def logical_product(layoutA, layoutB):
     if layoutB is None:
         return layoutA
     elif is_int(layoutB):
-        return logical_divide(layoutA, Layout(layoutB))
+        return logical_product(layoutA, Layout(layoutB))
     elif is_tuple(layoutB):
         assert len(layoutA) >= len(layoutB)
         return make_layout(
--- a/python/pycute/layout.py
+++ b/python/pycute/layout.py
@@ -313,7 +313,7 @@ def logical_product(layoutA, layoutB):
   if layoutB is None:
     return layoutA
   elif is_int(layoutB):
-    return logical_divide(layoutA, Layout(layoutB))
+    return logical_product(layoutA, Layout(layoutB))
   elif is_tuple(layoutB):
     assert len(layoutA) >= len(layoutB)
     return make_layout(chain((logical_product(layoutA[i], layoutB[i]) for i in range(           0,len(layoutB))),
--- a/test/python/pycute/test_product.py
+++ b/test/python/pycute/test_product.py
@@ -1,0 +1,90 @@
+#################################################################################################
+#
+# Copyright (c) 2023 - 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: BSD-3-Clause
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+# 1. Redistributions of source code must retain the above copyright notice, this
+# list of conditions and the following disclaimer.
+#
+# 2. Redistributions in binary form must reproduce the above copyright notice,
+# this list of conditions and the following disclaimer in the documentation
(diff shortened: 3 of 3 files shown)


ccecka:

Issue #3077, opened 2026-02-27
[BUG] TiledMMA modifier can't be set in different jit function


HarryWu99:
### Which component has the problem?

CuTe DSL

### Bug Report

**Describe the bug**
tiled_mma, tCrA, tCrB, tCrC

**Steps/Code to reproduce bug**
```
```

**Expected behavior**

**Environment details (please complete the following information):**
 - Environment location: [Bare-metal, Docker, Cloud(specify cloud provider)] 

**Additional context**

alecco:
## Quick fix:  in the callee remove `@cute.jit` (but then mind issues with parent-owned tensors, so check thoroughly if everything is OK, YMMV)

It is not GEMM. But it is something about conditionals.

Here is a smaller repro for a very similar case I found:

```
#!/usr/bin/env python3
"""Minimal CuTeDSL SM120 dominance repro.

This is the smallest clean A/B repro found so far:

- `one_site_kernel` compiles
- `two_sites_kernel` fails with `operand #0 does not dominate this use`

The helper does not read global memory and does not depend on input shapes. It
only writes a constant BF16 value into a `1x1` shared tile.

The trigger is structural:

- shared `sV` is wrapped in an attribute-bearing lowered object
- one helper call site is inside `range_constexpr(1)`
- the second helper call site is any later syntactic reuse of the same object

This is a verifier/lowering bug, not a runtime kernel bug.
"""

from types import SimpleNamespace

import cuda.bindings.driver as cuda
import cutlass
import cutlass.cute as cute
import cutlass.utils as utils
import torch

NUM_THREADS = 128


# NOTE: removing the decorator there is no issue
@cute.jit
def write_helper(smem_params: SimpleNamespace):
    tidx, _, _ = cute.arch.thread_idx()
    if 0 >= 0:
        smem_params.sV[0, 0] = cutlass.BFloat16(0.0)


@cute.kernel
def one_site_kernel(sV_layout: cute.Layout, SharedStorage: cutlass.Constexpr):
    sV = utils.SmemAllocator().allocate(SharedStorage).sV.get_tensor(sV_layout)
    smem_params = SimpleNamespace(sV=sV)
    if 0 >= 0:
        write_helper(smem_params)


@cute.kernel
def two_sites_kernel(sV_layout: cute.Layout, SharedStorage: cutlass.Constexpr):
    sV = utils.SmemAllocator().allocate(SharedStorage).sV.get_tensor(sV_layout)
[...]

PR #3480, opened 2026-08-21
[CuTeDSL] Fix "operand does not dominate this use" ICE: thread mutated call arguments out of dynamic `if`/`while` regions

zkyue:
## TL;DR

Objects passed as plain **call arguments** to a nested `@cute.jit` helper inside dynamic
(runtime-valued) control flow are not threaded through the generated `scf` region, unlike
assignments and method-call receivers. If the helper mutates the argument in place (e.g.
`TiledMma.set`), the object is left bound to a region-local SSA value and the next use fails
MLIR verification with `operand #0 does not dominate this use`. This PR marks plain-name
positional/keyword call arguments as potentially mutated in the region analyzer, and scopes
the threading exemptions in `is_read_only_object` so the wider sweep neither demotes
compile-time constants nor disturbs region-result threading. Regression tests cover both
directions. This is the same bug class as #3077 (SM90, bot-closed as stale, never fixed — it
still reproduces verbatim on the 4.7.0 wheel; the repro below is its SM100 flavor).

## Symptom and minimal repro

```python
import cutlass
import cutlass.cute as cute
from cutlass.cute.nvgpu import OperandMajorMode, tcgen05
import cutlass.utils.blackwell_helpers as sm100_utils
from cutlass.cute.typing import Int32, Float32, BFloat16


@cute.jit
def inner_set(tiled_mma: cute.TiledMma):
    tiled_mma.set(tcgen05.Field.ACCUMULATE, False)


class Repro:
    @cute.kernel
    def kern(self, tiled_mma: cute.TiledMma, flag: Int32):
        if flag > 0:                # dynamic if -> outlined into an scf.if region
            inner_set(tiled_mma)    # nested jit helper mutates the argument in place
        tiled_mma.set(tcgen05.Field.ACCUMULATE, True)   # any later use -> ICE

    @cute.jit
    def run(self, flag: Int32):
        tiled_mma = sm100_utils.make_trivial_tiled_mma(
            BFloat16, BFloat16, OperandMajorMode.K, OperandMajorMode.K,
            Float32, tcgen05.CtaGroup.ONE, (64, 32),
        )
        self.kern(tiled_mma, flag).launch(grid=(1, 1, 1), block=(32, 1, 1))


cutlass.cuda.initialize_cuda_context()
cute.compile(Repro().run, Int32(1))
```

On `nvidia-cutlass-dsl` 4.6.1 and 4.7.0 (B200, CUDA 13) this fails with:

```
error[INTERNAL]: The compiler could not build valid IR for this code.
  error: IR verification failed: error: unknown: operand #0 does not dominate this use
note: unknown: see current operation: %3 = "cute_nvgpu.atom.set_value"(%5, %2)
      <{field = #cute_nvgpu.atom_mma_field_sm100<accum_c>}> : (!cute.tiled_mma<...>, i1) -> ...
note: unknown: operand defined here (op in a child region)
```

All three ingredients are required; removing any one of them makes it compile:
1. a dynamic (non-constexpr) `if` (or `while`),
2. the mutation happening inside a **nested `@cute.jit` helper call** (inlining the
   `.set` compiles fine, because method-call receivers *are* threaded),
3. any later use of the object after the region.

## Root cause

Dynamic `if`/`while` bodies are outlined and mutated variables are threaded through the
`scf` op as region yields/results; after the op is built, the threaded Python objects are
rebound to the op results (`pack_from_irvalue`). Which variables get threaded is decided
statically by `DSLPreprocessor.analyze_region_variables` in
`python/CuTeDSL/cutlass/base_dsl/ast_preprocessor.py`:

- assignments (`x = ...`) are collected into `write_args`;
- method-call receivers are collected into `invoked_args` by `RegionAnalyzer.visit_Call`
  (`ast_preprocessor.py:1114-1131` at current main): *"Classes are mutable by default. Mark
  them as write."*;
- **plain call arguments are not collected at all** — `visit_Call` records the callee name
  but never looks at `node.args` / `node.keywords`.

A nested `@cute.jit` helper is traced inline at the current insertion point — inside the
`scf.if` child region. `TiledMma.set` mutates in place
(`self.value = _cute_nvgpu_ir.atom_set_value(...)`,
`python/CuTeDSL/cutlass/cute/nvgpu/tcgen05/mma.py:456`), so the shared Python object ends up
holding an SSA value defined in the child region. Since the object is not in the yield set,
nothing rebinds it when the region closes, and the next use references the region-local
value. The MLIR verifier correctly rejects the IR; the invalid IR is constructed entirely by
the Python frontend.

## The fix

**1. `ast_preprocessor.py`, `RegionAnalyzer.visit_Call` (+20 lines): the actual bug fix.**
Plain-name positional and keyword arguments are added to `invoked_args`, mirroring the
existing method-receiver rule (a callee can mutate an argument in place exactly like a
method can mutate its receiver). The existing `self`-under-`support_pyir` carve-out is
mirrored for consistency. `*args`/`**kwargs` container names are deliberately **not**
marked: threading the container would rebind only the container name and silently detach
the mutated element from the variable the caller still holds (verified in IR: the later use
kept consuming the pre-region value while the `scf.if` result went dead — a silent
mutation drop). Those forms keep failing loudly, exactly like current main.

**2. `cutlass_dsl/cutlass.py`, `is_read_only_object` (+85/−5 lines): scoped threading
exemptions.** `is_read_only_object` is the single choke point (`_filter_readonly_objects` /
`insert_readonly_objects` / `unpack_to_irvalue` / `pack_from_irvalue`) that exempts objects
from threading. The wider sweep makes two new object classes reachable, and threading either
one is a behavior regression; but the exemptions must not leak beyond variable captures.
Each rule is pinned by the concrete regression it prevents (all covered by the new tests):

- **Immutable Python values** (ints, floats, strings, enum members, tuples of such,
  functions, modules, types) stay read-only. A callee cannot observably mutate them, and
  threading them **demotes compile-time constants to staged Numerics** (`_tree_flatten`
  falls back to `as_numeric(x).ir_value()`). Without this rule, a Python `int` passed to a
  helper inside a dynamic `if` and later used as a `cutlass.range_constexpr` bound fails
  with `error[ARG_NON_CONSTANT]` (test: `test_constexpr_int_arg_stays_constexpr`; hit in a
  production sparse-attention kernel during validation).

- **Objects carrying no MLIR values** (no dynamic leaves under `tree_flatten`, or
  unflattenable) stay read-only. Threading exists solely to rebind region-defined SSA
  values; a plain Python config object passed to a helper inside a dynamic `if` must not be
  pushed through region flattening (test: `test_plain_python_object_arg_not_threaded`).

- **The pre-existing `self` opt-in rule keeps precedence**: an opted-in receiver
  (`_dsl_thread_self_in_staged_cf`) is always threaded regardless of its current contents.

- **Both rules apply only to named variable captures.** Region *results* flow through the
  same machinery (`_ifexp_execute_dynamic` -> `scf_execute_dynamic`) and must always be
  threaded: filtering an immutable branch value out of a dynamic if-expression produces a
  zero-result `scf.if` whose packing silently restores the then-value — `1 if flag > 0
  else 2` evaluated to `1` for `flag == 0` (caught by a runtime probe during pre-submission
  review of an earlier draft of this patch). Results are now tagged with the new
  `REGION_RESULT_NAME` placeholder (an invalid identifier, so it cannot collide with a real
  variable name) instead of `"unknown"`, and the two rules skip it; callers providing no
  name information are likewise exempt. Tests: `test_classification_is_scoped_to_named_captures`
  (unit) and `test_dynamic_ternary_of_literals_selects_at_runtime` (runtime, checks both
  branch values).

**3. `cutlass_dsl/cutlass_ast_decorators.py` (1 line):** `_ifexp_execute_dynamic` passes
`REGION_RESULT_NAME` as the placeholder argument name for ternary results.

With the fix, the minimal repro produces exactly the expected threading
(`CUTE_DSL_PRINT_IR=1`):

```mlir
%1 = scf.if %0 -> (!cute.tiled_mma<...>) {
  %3 = cute_nvgpu.atom.set_value<accum_c>(%arg0, %false) ...
  scf.yield %3 : ...
} else {
  scf.yield %arg0 : ...
}
%2 = cute_nvgpu.atom.set_value<accum_c>(%1, %true) ...   // consumes the scf.if result
```

## Validation

Everything below was run on a B200 (SM100) against the 4.7.0 wheel (whose copies of the
touched files are byte-identical to current main) with only the patched files overlaid.

| Gate | Unpatched | Patched |
|---|---|---|
| Minimal repro above | ICE | compiles |
| jit-helper `.set` in dynamic `if` + `cute.gemm` after | ICE | compiles |
| same, inside a traced `while` (yield as the use site) | ICE | compiles |
| #3077 shape (jit `.set` one branch, jit `gemm` other branch) | ICE | compiles |
| #3077 shape inside a traced `while` | ICE | compiles |
| 12-variant control ladder (inline `.set`, no-later-use, constexpr `if`, unconditional helper, ...) | compiles | compiles (unchanged) |
| New regression test file (10 tests) | 5 fail: 4 mutation shapes with the dominance ICE + 1 new-API unit test; 5 controls pass | 10 pass |
| Dynamic ternary of literals, runtime, both flag values | correct (2 / 1) | correct (2 / 1) |
| Existing `test/python/CuTeDSL/` suite | — | failure set byte-identical before/after (pre-existing wheel-vs-main drift only) |
| Sanity kernel (compile + launch + value check) | OK | OK |
| Production sparse-attention kernels, forward + backward (compile + run) | OK | OK |
| Production numerics, fixed seed, unpatched vs patched | — | 3 of 4 outputs bitwise-equal; the 4th uses non-deterministic float atomics and stays inside its run-to-run band (max abs diff 7.6e-5 vs 6.1e-5–6.9e-5 across unpatched re-runs) |

The new tests were verified in both directions with exact counts. The `TiledMma.set` tests
skip on non-SM10x devices; the guard tests are arch-neutral; the classification unit test
needs no GPU; the runtime ternary test is additionally gated on torch + CUDA availability.

## Prior art

#3077 ("TiledMMA modifier cannot be set in different jit function", SM90, filed 2026-02) is
this exact bug class. It was closed by the stale bot with no fix; the construct still
reproduces on 4.7.0. This PR fixes the underlying region-threading gap and covers that shape
in the regression tests (`test_helpers_in_both_branches`).

## Scope and limitations

- **Non-Name argument forms** (`helper(obj.attr)`, `helper(xs[0])`, literal `*`/`**`
  containers) are not marked and keep failing exactly like current main. This matches the
  name granularity of the existing analyzer (the receiver rule reduces to a base name the
  same way).
- **`*args`/`**kwargs` names** are deliberately excluded (see above): loud failure was
  preferred over silent mutation drop through the aliased element.
- **Mixed static/dynamic containers**: a container like `[dyn, 4]` passed to a helper is
  now threaded, which demotes the static leaf — a later `range_constexpr(xs[1])` fails
  loudly. This is the DSL's established semantics for threaded containers: the identical
  error already occurs on current main when the same container is threaded by assignment
  inside the region (verified). Keeping constants out of mutated containers (or
  `const_expr`) is the existing remedy.
- **Callee-changed dynamic structure**: if a helper changes which leaves of a threaded
  container are dynamic (e.g. replaces a staged value with a Python literal), compilation
  fails loudly with a structure-mismatch assertion. On current main such code only
  "compiled" because the trace-time mutation escaped region discipline entirely (the
  branch effect applied unconditionally).
- **PyIR preprocessor mode** (non-default) shares `analyze_region_variables` but threads
  values through a different mechanism; the underlying bug is not fixed for PyIR mode by
  this PR and our validation does not exercise it. The new sweep mirrors the existing
  `support_pyir` special case for `self`.
- `_carries_mlir_values` runs `tree_flatten` once per candidate object per classification
  site at trace time. If this shows up in compile-time profiles it can be cached; the patch
  was kept minimal.

### DIFF
--- a/python/CuTeDSL/cutlass/base_dsl/ast_preprocessor.py
+++ b/python/CuTeDSL/cutlass/base_dsl/ast_preprocessor.py
@@ -1127,6 +1127,26 @@ class DSLPreprocessor(ast.NodeTransformer):
                     else:
                         invoked_args.add(base_name)
 
+                # Arguments passed to a call may be mutated in place by the
+                # callee (e.g. a nested @jit helper mutating a TiledMma via
+                # `.set`), exactly like the receiver of a method call above.
+                # Mark plain-name arguments as invoked so the mutated value is
+                # threaded out of the scf region instead of leaking a
+                # region-local SSA value ("operand does not dominate this use").
+                # `*args`/`**kwargs` container names are deliberately NOT
+                # marked: threading the container rebinds only the container
+                # name, silently detaching the mutated element from the
+                # variable the caller still holds, so those forms must keep
+                # failing loudly instead of miscompiling.
+                arg_exprs = list(node.args) + [
+                    kw.value for kw in node.keywords if kw.arg is not None
+                ]
+                for arg in arg_exprs:
+                    if isinstance(arg, ast.Name):
+                        if arg.id == "self" and support_pyir:
+                            continue
+                        invoked_args.add(arg.id)
+
                 self.generic_visit(node)
 
         analyzer = RegionAnalyzer()
--- a/python/CuTeDSL/cutlass/cutlass_dsl/cutlass.py
+++ b/python/CuTeDSL/cutlass/cutlass_dsl/cutlass.py
@@ -2324,13 +2324,74 @@ class KernelLauncher:
 # =============================================================================
 # Utils
 # =============================================================================
+
+# Placeholder passed as the "argument name" when filtering region RESULTS
+# (e.g. dynamic if-expression branch values) rather than named variable
+# captures. Results are always yielded out of the region, so the
+# mutability-based read-only rules for captured variables must not apply to
+# them. An invalid Python identifier is used so it can never collide with a
+# real variable name.
+REGION_RESULT_NAME = "<region result>"
+
+
+def _is_immutable_python_value(item: Any) -> bool:
+    """True if `item` is an immutable Python value that a callee cannot
+    mutate in place (so it never needs to be threaded through staged
+    control-flow regions)."""
+    import enum as _enum
+    import types as _types
+
+    if item is None or isinstance(item, (int, float, complex, str, bytes)):
+        return True
+    if isinstance(item, _enum.Enum):
+        return True
+    if isinstance(item, (tuple, frozenset)):
+        return all(_is_immutable_python_value(e) for e in item)
+    if isinstance(
+        item,
+        (
+            _types.FunctionType,
+            _types.BuiltinFunctionType,
+            _types.MethodType,
+            _types.ModuleType,
+            type,
+        ),
+    ):
+        return True
+    return False
+
+
+def _carries_mlir_values(item: Any) -> bool:
+    """True if `item` (possibly a container or dataclass) holds any MLIR SSA
+    values that staged control flow would need to thread through region
+    boundaries."""
+    try:
+        leaves, _, _ = tree_flatten(item, return_ir_values=False)
+    except Exception:
+        return False
+    return any(is_dynamic_expression(leaf) for leaf in leaves)
+
+
 def is_read_only_object(item: Any, arg_name: Optional[str] = None) -> bool:
     """
-    Check if an item is a read-only object.
+    Check if an item is a read-only object (exempt from being threaded
+    through staged control-flow regions).
 
-    A read-only object is either a frozen dataclass or a method receiver
-    (``self``) whose class does not opt in to ``self`` threading via the
-    ``_dsl_thread_self_in_staged_cf`` class attribute.
+    An item is read-only when it is:
+
+    * a frozen dataclass;
+    * a method receiver (``self``) whose class does not opt in to ``self``
+      threading via the ``_dsl_thread_self_in_staged_cf`` class attribute
+      (an opted-in receiver is always threaded);
+    * any other named variable capture (``arg_name`` is a real variable
+      name) that a callee cannot observably mutate in place: an immutable
+      Python value, or an object carrying no MLIR SSA values.
+
+    The mutability-based rules apply only to named captures: region results
+    (``arg_name`` is :data:`REGION_RESULT_NAME`) and callers providing no
+    name information are always threaded as before, since e.g. dynamic
+    if-expression branch values must be yielded even when they are plain
+    Python literals.
     """
     if is_frozen_dataclass(item):
         return True
@@ -2340,8 +2401,25 @@ def is_read_only_object(item: Any, arg_name: Optional[str] = None) -> bool:
         # SSA values, which sibling regions then read ("operand does not
         # dominate this use"). Classes that mutate state reachable through
         # `self` inside dynamic regions set the flag (e.g. task_scheduling's
-        # Task).
+        # Task). This rule takes precedence over the mutability-based rules
+        # below so an opted-in receiver stays threaded regardless of its
+        # current contents.
         return not getattr(type(item), "_dsl_thread_self_in_staged_cf", False)
+    if arg_name is not None and arg_name != REGION_RESULT_NAME:
+        if _is_immutable_python_value(item):
+            # Immutable Python values (constexpr ints, strings, enum members,
+            # tuples of such, ...) cannot be observably mutated by a callee.
+            # Threading them through staged control flow would demote them
+            # from compile-time constants to staged Numerics (breaking e.g.
+            # range_constexpr bounds), so keep them read-only.
+            return True
+        if not _carries_mlir_values(item):
+            # Threading exists solely to rebind region-defined SSA values
+            # after the region closes. An object holding no MLIR values has
+            # nothing to rebind; threading it would either demote
+            # compile-time state to staged values or fail flattening
+            # entirely (plain Python objects).
+            return True
     return False
 
 
--- a/python/CuTeDSL/cutlass/cutlass_dsl/cutlass_ast_decorators.py
+++ b/python/CuTeDSL/cutlass/cutlass_dsl/cutlass_ast_decorators.py
@@ -853,7 +853,7 @@ def _ifexp_execute_dynamic(
         op_type_name="ifexp",
         mix_iter_args=mix_iter_args,
         full_write_args_count=0,
-        mix_iter_arg_names=["unknown" for _ in mix_iter_args],
+        mix_iter_arg_names=[cutlass_dsl.REGION_RESULT_NAME for _ in mix_iter_args],
         create_op_func=create_if_op,
         region_builders=region_builders,
     )
--- a/test/python/CuTeDSL/test_call_args_in_if.py
+++ b/test/python/CuTeDSL/test_call_args_in_if.py
@@ -1,0 +1,329 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: LicenseRef-NvidiaProprietary
+#
+# Use of this software is governed by the terms and conditions of the
+# NVIDIA End User License Agreement (EULA), available at:
+# https://docs.nvidia.com/cutlass/latest/media/docs/pythonDSL/license.html
+#
+# Any use, reproduction, disclosure, or distribution of this software
+# and related documentation outside the scope permitted by the EULA
+# is strictly prohibited.
+
+"""
+Unit tests for mutable objects passed as *arguments* to nested @cute.jit
+helpers inside DSL dynamic control flow.
+
+Dynamic `if`/`while` bodies are outlined into scf regions, and any object a
+callee may mutate in place must be threaded through the region boundary
+(yielded and rebound) so uses after the region see the region result rather
+than a region-local SSA value. The region analyzer marks assignments and
+method-call receivers, and it must mark plain-name call arguments the same
+way: a nested helper such as
+
+    @cute.jit
+    def disable_accumulate(tiled_mma: cute.TiledMma):
+        tiled_mma.set(tcgen05.Field.ACCUMULATE, False)
+
+called inside a dynamic `if` otherwise leaves the mutated TiledMma bound to
+a value defined in the child region, and the next use of it fails MLIR
+verification with "operand #0 does not dominate this use" (NVIDIA/cutlass
+issue #3077 is the SM90 flavor of the same failure).
+
+The extra threading must stay scoped to named variable captures that can
+actually observe a mutation:
+
+* immutable Python values (e.g. constexpr ints) and plain Python objects
+  carrying no MLIR values must stay read-only, or they would be demoted to
+  staged values (breaking e.g. `cutlass.range_constexpr` bounds) or fail
+  flattening at the region boundary;
+* region *results* (dynamic if-expression branch values) must keep being
+  threaded even when they are plain Python literals, or
+  `1 if pred else 2` silently constant-folds to its then-value.
+
+The TiledMma mutation tests require a tcgen05-capable (SM 10.x) device
+because `TiledMma.set` is the canonical in-place mutator; the read-only
+guard tests are architecture-neutral, and the classification tests need no
+GPU at all.
+"""
+
+import unittest
+
+import cutlass
+import cutlass.cute as cute
+import cutlass.utils.blackwell_helpers as sm100_utils
+from cutlass.base_dsl.runtime.cuda import get_compute_capability_major_minor
+from cutlass.cute.nvgpu import OperandMajorMode, tcgen05
+from cutlass.cute.typing import BFloat16, Float32, Int32
+
+try:
+    _CC_MAJOR, _ = get_compute_capability_major_minor()
+except Exception:  # no GPU / broken driver: only device-less tests can run
+    _CC_MAJOR = None
+
+try:
+    import torch
+
+    _HAS_TORCH_CUDA = torch.cuda.is_available()
+except Exception:
+    _HAS_TORCH_CUDA = False
+
+requires_tcgen05 = unittest.skipUnless(
+    _CC_MAJOR == 10,
+    "TiledMma.set tests require a tcgen05-capable (SM 10.x) device",
+)
+
+requires_gpu_runtime = unittest.skipUnless(
+    _CC_MAJOR is not None and _HAS_TORCH_CUDA,
+    "runtime check requires a CUDA device and torch",
+)
+
+
+@cute.jit
+def disable_accumulate(tiled_mma: cute.TiledMma):
+    tiled_mma.set(tcgen05.Field.ACCUMULATE, False)
+
+
+@cute.jit
+def enable_accumulate(tiled_mma: cute.TiledMma):
+    tiled_mma.set(tcgen05.Field.ACCUMULATE, True)
+
+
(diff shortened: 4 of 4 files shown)


python/CuTeDSL/cutlass/cutlass_dsl/cutlass.py
```diff
@@ -2324,13 +2324,74 @@ def __call__(self, *args: Any, **kwargs: Any) -> Any:
...
+            _types.MethodType,
+            _types.ModuleType,
+            type,
+        ),
+    ):
+        return True
+    return False
+
+
+def _carries_mlir_values(item: Any) -> bool:
```
brandon-yujie-sun, line 2364: I guess the following snippet might still not work?

```
  @cute.jit
  def fill(acc, x):
      acc.append(x)          # mutates the caller's list in place

  @cute.jit
  def kern(n: cutlass.Int32, pred: cutlass.Boolean):
      acc = []               # read-only
      if pred:
          fill(acc, n)       # [n] -> _carries_mlir_values True, and then mismatches at the child_tree defs length check
      cute.printf("done\n")

  kern(cutlass.Int32(1), cutlass.Boolean(True))
```

brandon-yujie-sun (commented):
Thanks for contributing the codes! This is a known limitation of DSL that mutated passed in args are not fully supported in this case. Like you mentioned, the workaround is inlining. We do have some work in progress for better support on DSL frontend which is planned for future release. That said, would it work to live with the WAR for now?

brandon-yujie-sun:
@anakinxc @grypp

zkyue:
@brandon-yujie-sun Thanks for taking a look, and for the probe — it was a good one to check.

**On the snippet:** your prediction is exactly right for it as written: with this PR, `_carries_mlir_values` flips after the in-place `append`, the entry-side and yield-side flattenings disagree on arity, and it stops at the `child_treedefs` length assert (bare `AssertionError` at `tree_utils.py:1033`, at trace time, with the user's `if` in the traceback). Two data points from running it on B200 that may be useful context:

- On current `main` the same snippet compiles and prints `done`, but the conditional append actually escapes the region: with `pred=False` at runtime, `len(acc)` is still 1 afterwards (the mutation applies unconditionally). And if a value computed inside the region is consumed after the `if`, it becomes the same dominance ICE this PR addresses. So this shape is broken on `main` either way; the PR changes the failure to a loud trace-time one.
- The non-empty starting variant (`acc = [Int32(0)]`, append changes only the length) doesn't hit the bare assert: it's caught earlier by the per-arg region-result check and reports a proper `error[TYPE_UNSTABLE_JOIN]: 'acc' has type 'list<1>' on one path and 'list<2>' on another`, pointing at the `if`.

Since a runtime-branch-dependent container length can't be expressed as a static `scf.if` yield arity, erroring is the intended behavior for this class (it's the "callee-changed dynamic structure" limitation in the PR body). If useful, I'm happy to push a small follow-up that upgrades the bare assert in the empty-container case to the same `TYPE_UNSTABLE_JOIN` diagnostic and adds both of your variants as negative tests.

**On living with the inlining workaround:** understood, and no objection to the frontend rework superseding this — you know the roadmap. Two small requests in the meantime:

1. Until the rework lands, would you consider a user-facing diagnostic for this pattern (a trace-time error suggesting inlining the mutation)? The current failure is an MLIR-level ICE with no hint of the workaround, and as the data point above shows, some shapes don't error at all. If you'd rather not take the semantic fix now, I'd be glad to reduce this PR to just the diagnostic plus the regression tests.
2. It would be great if the rework tracks this case (and #3077, the SM90 twin) — the minimal repros and tests here are yours to reuse as regression coverage.

Happy to keep the PR open as a reference, reduce it as above, or close it in favor of your internal tracking — whichever you prefer.

anakinxc:
Hi @zkyue 

Are you guys using CuTe DSL 4.7+?

If so, you can try this already with `cute.compile[cute.FrontendNext](...)`, although it's still early testing, but feedbacks are welcome :)

python/CuTeDSL/cutlass/base_dsl/ast_preprocessor.py
```diff
@@ -1127,6 +1127,26 @@ def visit_Call(self, node: ast.Call) -> None:
                     else:
                         invoked_args.add(base_name)
 
+                # Arguments passed to a call may be mutated in place by the
```
grypp, line 1130: comments are unnecessary. let's keep code clean

python/CuTeDSL/cutlass/base_dsl/ast_preprocessor.py
```diff
@@ -1127,6 +1127,26 @@ def visit_Call(self, node: ast.Call) -> None:
...
+                # region-local SSA value ("operand does not dominate this use").
+                # `*args`/`**kwargs` container names are deliberately NOT
+                # marked: threading the container rebinds only the container
+                # name, silently detaching the mutated element from the
+                # variable the caller still holds, so those forms must keep
+                # failing loudly instead of miscompiling.
+                arg_exprs = list(node.args) + [
+                    kw.value for kw in node.keywords if kw.arg is not None
+                ]
+                for arg in arg_exprs:
```
grypp, line 1144: IIUC, if a symbol escapes to another function, preprocessor assumes here that it's write_args.

python/CuTeDSL/cutlass/cutlass_dsl/cutlass.py
grypp: This program won't work, because `acc` is meta program list, so we cannot `append` inside control-flow. this PR could catch this error out?

zkyue:
Thanks for the pointer — we took `FrontendNext` for a spin. Setup: released
wheels `nvidia-cutlass-dsl[cu13]` 4.7.0 and 4.7.1 (results identical), B200
(SM100), compiling with `cute.compile[cute.FrontendNext](fn, *args)`. We ran
the full regression set from this PR (both the used-to-ICE shapes and the
must-keep-working controls) plus the list-append probe from the earlier
review discussion. Only adaptation: swapping the compile selector; case
bodies unchanged.

Results:

| case | default frontend (4.6.1/4.7.0/4.7.1) | FrontendNext (4.7.0/4.7.1) |
|---|---|---|
| helper mutates TiledMma arg in dynamic `if` (positional / kwarg / both-branches #3077 / `while`) | ICE "operand #0 does not dominate this use" | **still the same ICE** |
| inline `tiled_mma.set(...)` in dynamic `if` (this PR's positive control) | compiles | **ICE (FrontendNext-only regression)** |
| read-only guards (constexpr int arg, plain Python object arg, element-wise list mutation) | compile | compile |
| `1 if pred else 2` with dynamic pred (runtime select) | correct (2/1) | correct (2/1) |
| list `append` via nested helper in dynamic `if` (empty list, unused after) | accepted, but the trace-time append escapes the region | loud, clean rejection |
| same, `len(acc)` read after the `if`, runtime pred=False | **silent wrong: prints `len=1`** | loud rejection |
| same, region-computed element consumed after the `if` | ICE (dominance) | loud rejection |

So, three takeaways from our side:

1. The list-mutation family is genuinely resolved by FrontendNext, and in
   the right way: a clean trace-time diagnostic ("In-place mutation
   `acc.append(...)` of a meta Python list is not allowed inside staged
   control flow", with a suggestion). Since a length-changing container
   across a dynamic `if` has data-dependent structure, rejection is the
   only sound outcome — this fully replaces the default frontend's
   silent-wrong/ICE behavior for that family. Nice.

2. The dominance pathology this PR targets is **not** resolved. All four
   nested-helper TiledMma shapes still fail under FrontendNext with the
   byte-identical `operand #0 does not dominate this use` on the same
   `cute_nvgpu.atom.set_value` op ("operand defined here (op in a child
   region)"). Moreover FrontendNext regresses the inline form, which the
   default frontend compiles fine:

   ```python
   @cute.kernel
   def kern(tiled_mma: cute.TiledMma, flag: Int32):
       if flag > 0:
           tiled_mma.set(tcgen05.Field.ACCUMULATE, False)
       tiled_mma.set(tcgen05.Field.ACCUMULATE, True)

   @cute.jit
   def entry(flag: Int32):
       tiled_mma = sm100_utils.make_trivial_tiled_mma(
           BFloat16, BFloat16, OperandMajorMode.K, OperandMajorMode.K,
           Float32, tcgen05.CtaGroup.ONE, (64, 32))
       kern(tiled_mma, flag).launch(grid=(1, 1, 1), block=(32, 1, 1))

   cute.compile[cute.FrontendNext](entry, Int32(1))   # ICE on 4.7.0 and 4.7.1
   cute.compile(entry, Int32(1))                      # compiles
   ```

   Unconditional `TiledMma.set` (no control flow) compiles fine under
   FrontendNext, so this looks like the region threading for
   value-carrying objects, not TiledMma support itself.

3. Adoption data point: we also tried FrontendNext on one of our existing
   production CuTe DSL kernels. It is rejected at trace time — FrontendNext
   enforces "variables used inside staged control flow must be defined
   before the region" for a phase variable assigned in one staged region
   and consumed in a later `mbarrier_wait`, a pattern the default frontend
   traces fine. We also noticed `warning[PHASE_AUTO_PROMOTED_TO_STAGED]`
   firing from inside `cutlass/utils/tmem_allocator.py:485`
   (`self._num_allocated_columns`) itself — flagging in case that warning
   in the DSL's own utils is unintended.

Given that FrontendNext (as of 4.7.1) doesn't yet cover the dominance bug
and the default frontend remains the shipping path, we'd suggest this PR is
still useful as-is; happy to rebase or trim it however you prefer, and to
file the FrontendNext findings above (items 2 and 3) as separate issues
with the minimal repros if that's helpful.

anakinxc:
Hi @zkyue 

Just tested both repro here and #3077 with latest internal version, with `FrontendNext` both compiles as expected now.

wrt your fix, this does fix the issue you observed here, this is actually something I tried a while ago, but this introduced a major compatibility breakage. Thinking about the following case:

```python
import cutlass
import cutlass.cute as cute


@cute.jit
def magic_print(object: MyObject):
    cute.printf(...)
    // lots of debug print


@cute.jit
def kern(debug_flag):
    obj = MyObject(...)
    if debug_flag > 0:      # dynamic if -> outlined into an scf.if region
        magic_print(obj)    # nested jit helper
    obj.whatever()          # any later use

```

With this fix, this code will no longer works if `MyObject` does not implement `DynamicExpressionProtocol`

zkyue:
@anakinxc That's great news — thanks for testing both repros against the internal build, and the `MyObject` example makes the compatibility cost of the tracer-side fix concrete: threading plain-name args requires region-carried objects to be flattenable, which silently raises the bar for every nested helper that today just borrows an object for side effects. Given that you already explored this exact approach and rejected it for that reason, and that FrontendNext resolves both this PR's shapes and #3077 at the root, we agree the new frontend is the right home for this class of fix rather than patching the legacy tracer. Closing this PR in favor of FrontendNext.

Two small offers before we go:

- The regression suite from this PR (nested-helper mutation via positional/kwarg/both-branches/`while`, the read-only guard controls, the runtime-select control) plus the list-mutation probe family from the review discussion made a decent torture set — on released 4.7.0/4.7.1 wheels they caught the inline `tiled_mma.set` case, which you've since confirmed fixed internally. If a test-only contribution of those cases would be useful for FrontendNext validation, happy to send it as a separate PR or attach them to an issue.
- We'll re-run the full set on the next released wheel that carries the internal frontend fixes and report back on this thread if anything still reproduces.

Thanks @brandon-yujie-sun, @grypp and @anakinxc for the careful reviews.

PR #3482, opened 2026-08-24
pycute: fix logical_product with an integer tiler

### LINKED ISSUE #3478: [BUG] pycute: logical_product with an integer tiler computes logical_divide

VaggelisGian:
Fixes #3478

`pycute.logical_product` dispatched an integer tiler to `logical_divide`, so `logical_product(A, n)` returned a divide result under a product name and disagreed with the equivalent spelling `logical_product(A, Layout(n))`. The tuple branch recurses into the integer branch per mode, which made `zipped_product` and `tiled_product` wrong for integer tilers as well. The same line exists in the vendored copy under `operators/cutlass/operators/fusion/pycute` and is fixed there too.

The C++ reference in `include/cute/layout.hpp` converts an integral tiler with `make_layout(tiler)` and calls `logical_product`; the Python branch now does the same.

Adds `test/python/pycute/test_logical_product.py` covering the integer, Layout, tuple and None spellings plus the zipped/tiled forms; three of its four cases fail on the unfixed code.

Test Plan:
```
...
----------------------------------------------------------------------
Ran 14 tests in 0.007s
OK
```
14 tests = 10 pre-existing + 4 new; with the fix stashed, 3 of the new ones fail (`AssertionError: 12 != 40` on the size post-condition).

Both modified files byte-compile: `python -m py_compile python/pycute/layout.py operators/cutlass/operators/fusion/pycute/layout.py`.

Expected outputs were also cross-checked against the C++ implementation compiled host-side (g++ against include/cute/layout.hpp): identical on every valid case tested, including the direct int overload.

### DIFF
--- a/operators/cutlass/operators/fusion/pycute/layout.py
+++ b/operators/cutlass/operators/fusion/pycute/layout.py
@@ -336,7 +336,7 @@ def logical_product(layoutA, layoutB):
     if layoutB is None:
         return layoutA
     elif is_int(layoutB):
-        return logical_divide(layoutA, Layout(layoutB))
+        return logical_product(layoutA, Layout(layoutB))
     elif is_tuple(layoutB):
         assert len(layoutA) >= len(layoutB)
         return make_layout(
--- a/python/pycute/layout.py
+++ b/python/pycute/layout.py
@@ -313,7 +313,7 @@ def logical_product(layoutA, layoutB):
   if layoutB is None:
     return layoutA
   elif is_int(layoutB):
-    return logical_divide(layoutA, Layout(layoutB))
+    return logical_product(layoutA, Layout(layoutB))
   elif is_tuple(layoutB):
     assert len(layoutA) >= len(layoutB)
     return make_layout(chain((logical_product(layoutA[i], layoutB[i]) for i in range(           0,len(layoutB))),
--- a/test/python/pycute/test_logical_product.py
+++ b/test/python/pycute/test_logical_product.py
@@ -1,0 +1,104 @@
+#################################################################################################
+#
+# Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: BSD-3-Clause
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+# 1. Redistributions of source code must retain the above copyright notice, this
+# list of conditions and the following disclaimer.
+#
+# 2. Redistributions in binary form must reproduce the above copyright notice,
+# this list of conditions and the following disclaimer in the documentation
+# and/or other materials provided with the distribution.
+#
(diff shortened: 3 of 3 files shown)


ccecka:

VaggelisGian:
Understood, thanks for the direction. Since this copy of pycute is scheduled for removal in favor of NVlabs/CuTe, I am closing the fix and the matching regression tests rather than adding review load on code that will be deleted. Happy to re-apply any of it against NVlabs/CuTe if that tree turns out to share one of these bugs.

PR #3490, opened 2026-08-24
Fix pycute Swizzle recursion on zero shift

VaggelisGian:
`pycute.shiftr`/`shiftl` guarded on `s > 0` and delegated the else case to each other with a negated shift, so shift 0 ping-ponged between them until `RecursionError`. `Swizzle(0, M, 0)` passes the constructor assert (`abs(0) >= 0`), is legal in C++ where `Swizzle<0,4,3>` is literally the default swizzle, and applies as the identity; in pycute it was unusable:

```python
>>> Swizzle(0, 4, 0)(123)
RecursionError: maximum recursion depth exceeded
```

This uses `s >= 0` like the C++ reference in `include/cute/numeric/math.hpp` (`shiftl`/`shiftr`). The same two lines exist in the vendored copy under `operators/cutlass/operators/fusion/pycute` and are fixed there too.

Adds `test/python/pycute/test_swizzle.py`: shift-zero identity, application tables for four patterns cross-checked against compiled `cute::Swizzle` (host-side g++ against include/cute/swizzle.hpp), and full-domain involutivity checks.

Test Plan:
```
$ PYTHONPATH=python python test/python/pycute/test_swizzle.py
Ran 3 tests ... OK          (RecursionError x2 on unfixed code)
Ran 13 tests ... OK
```
C++ golden check (docker g++, include/cute/swizzle.hpp):
```
S0_4_0 : 0 1 2 7 8 9 15 16 63 64 255 256
SW23_3 : 0 1 2 7 72 73 79 144 255 64 63 256
SW101  : 0 1 3 6 8 9 14 16 62 64 254 256
SW323  : 0 1 2 7 8 9 15 16 59 72 227 256
SW212_ : 0 1 10 31 8 9 23 16 39 64 231 256
```
identical to the fixed Python implementation on every probe point.

### DIFF
--- a/operators/cutlass/operators/fusion/pycute/swizzle.py
+++ b/operators/cutlass/operators/fusion/pycute/swizzle.py
@@ -38,11 +38,11 @@ from .layout import *
 
 
 def shiftr(a, s):
-    return a >> s if s > 0 else shiftl(a, -s)
+    return a >> s if s >= 0 else shiftl(a, -s)
 
 
 def shiftl(a, s):
-    return a << s if s > 0 else shiftr(a, -s)
+    return a << s if s >= 0 else shiftr(a, -s)
 
 
 ## A generic Swizzle functor
--- a/python/pycute/swizzle.py
+++ b/python/pycute/swizzle.py
@@ -38,11 +38,11 @@ from .layout import *
 
 
 def shiftr(a, s):
-  return a >> s if s > 0 else shiftl(a, -s)
+  return a >> s if s >= 0 else shiftl(a, -s)
 
 
 def shiftl(a, s):
-  return a << s if s > 0 else shiftr(a, -s)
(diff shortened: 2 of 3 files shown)


ccecka:

VaggelisGian:
Understood, closing along with the other pycute fixes given the plan to remove this copy in favor of NVlabs/CuTe. The zero-shift recursion and the idx2crd broadcast fix are worth a quick check against the NVlabs/CuTe tree before it ships.

VaggelisGian:
Follow-up on the check suggested above: NVlabs/CuTe's pycute is clean for all three bugs this copy had - shiftr/shiftl already use the >= 0 guard, idx2crd has a different signature without the stride-0 broadcast case, and logical_product with an integer tiler produces a proper product. Nothing to port.

PR #3531, opened 2026-08-25
Fix packed uint1b logical array reductions

### LINKED ISSUE #3510: uint1b_t ReduceArray specializations are broken (logical_or never compiles / always true; logical_and tests bytes not bits)
### Description

The `uint1b_t` specializations of `ReduceArrayOperation` in include/cutlass/reduction/thread/reduction_operators.h (lines ~157-204) are broken in three independent ways.

#### logical_or: wrong parameter type AND constant-true result

```cpp
template <int N>
struct ReduceArrayOperation<logical_or<uint1b_t>, uint1b_t, N> {
  ...
  uint1b_t operator()(
    logical_and<uint1b_t> const &reduction_op,   // <-- copy-paste: accepts only logical_and
    ArrayType const &array) const {
    ...
    bool item = true;                            // <-- identity of OR is false
    for (...) { bits = ptr[byte]; item = (item || bits); }
    return uint1b_t{item};                       // <-- always true
```

Consequences, both verified against the current tree:

1. The public helper does not compile:
   `cutlass::reduction::thread::detail::ReduceArray(logical_or<uint1b_t>{}, array)` fails with *"operator() cannot be called with the given argument list"* because the parameter type names `logical_and`.
2. Even calling it directly with a `logical_and` argument compiles but ignores its input entirely: `item` starts `true` and is only ever OR-ed into, so the result is constant `true`.

#### logical_and: tests whole bytes, not bits

```cpp
bool item = false;
for (int byte = 0; byte < (N + 7) / 8; ++byte) {
  uint8_t bits = ptr[byte];
  item = (item || !bits);      // "some byte is entirely zero"
}
return uint1b_t{!item};
```

This returns "no byte is entirely zero", not "all elements are set". With `N = 8` and element 7 cleared, byte 0 is `0x7F`, no byte is zero, and the reduction reports true:

```
AND [1]*7+[0]: got=1 expected=0
```

There is also no masking of padding bits when N is not a multiple of 8.

Neither specialization has in-tree callers today, so both are latent public-API defects.

### Suggested fix

For logical_or: change the parameter type to `logical_or<uint1b_t>` and initialize `bool item = false;`.
For logical_and: loop over the ceil(N/8) bytes, mask the final byte's padding bits, and test per-byte zero as intended - or simply iterate elements via the array API.

wei-core:
## Summary

- fix the `uint1b_t` logical OR reduction specialization to accept `logical_or`
- use the correct identities and packed-byte checks for logical AND and OR
- mask padding bits in partial trailing bytes
- add regression coverage across logical bit and storage boundaries

## Background

The logical OR specialization currently cannot be called through `ReduceArray` because its operator takes `logical_and`. Even when called directly through the specialization, its `true` identity makes an all-zero array reduce to true.

The logical AND specialization checks whether each storage byte is nonzero rather than whether every valid bit is set. For example, `0x7f` incorrectly reduces to true. Partial trailing bytes also need a mask so padding bits do not participate in either reduction.

This change keeps the packed-byte path: full bytes are reduced directly, while only the valid low bits of a partial trailing byte are considered.

## Contribution roles

I led the contribution direction and acceptance criteria, reviewed and approved
the final design, and reviewed every changed line. I understand the change and
own the final design decision and the submission. AI assistance was used for
repository research, root-cause analysis, implementation preparation, and
automated validation execution.

## Testing

- reproduced the public logical OR helper compilation failure on the unmodified main branch
- reproduced incorrect runtime results on main: AND of `[1,1,1,1,1,1,1,0]` returned 1, and OR of an all-zero array returned 1
- manually compiled and ran the new regression source with MSVC 19.44, CUDA 12.9 headers, CCCL headers, and GoogleTest v1.14.0
- 2/2 tests passed, exhaustively checking each logical position for `N = 1, 7, 8, 9, 15, 16, 32`, including padding-bit isolation
- `git diff --check`

Fixes #3510

### DIFF
--- a/include/cutlass/reduction/thread/reduction_operators.h
+++ b/include/cutlass/reduction/thread/reduction_operators.h
@@ -160,19 +160,23 @@ struct ReduceArrayOperation<logical_and<uint1b_t>, uint1b_t, N> {
 
   CUTLASS_HOST_DEVICE
   uint1b_t operator()(
-    logical_and<uint1b_t> const &reduction_op, 
+    logical_and<uint1b_t> const &,
     ArrayType const &array) const {
 
-    uint8_t const *ptr = reinterpret_cast<uint8_t const *>(&array);
-    bool item = false;
+    uint8_t const *ptr = reinterpret_cast<uint8_t const *>(array.raw_data());
+    bool item = true;
 
     CUTLASS_PRAGMA_UNROLL
-    for (int byte = 0; byte < (N + 7) / 8; ++byte) {
-      uint8_t bits = ptr[byte];
-      item = (item || !bits);
+    for (int byte = 0; byte < N / 8; ++byte) {
+      item = item && (ptr[byte] == uint8_t(0xff));
     }
 
-    return uint1b_t{!item};
+    if (N % 8) {
+      uint8_t mask = uint8_t((1u << (N % 8)) - 1);
+      item = item && ((ptr[N / 8] & mask) == mask);
+    }
+
+    return uint1b_t{item};
   }
 };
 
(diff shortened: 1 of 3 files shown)


wei-core:
Maintenance check: GitHub still reports this PR as mergeable, with no unresolved review threads or failing checks, but review has not started yet. The change is scoped to #3510 and includes exhaustive packed-bit boundary coverage.

Could a maintainer triage it or point me to the appropriate reviewer? I am happy to refresh the branch or run any additional validation that would help.

PR #3565, opened 2026-08-28
[CuTeDSL] Preserve deterministic link library ordering

### LINKED ISSUE #3564: [BUG] CuTe DSL: _merge_gpu_link_libraries emits link-libraries in Python set order, so the compile pipeline string varies per process
## Describe the bug

`DSL._merge_gpu_link_libraries` collects each `gpu.module`'s `link-libraries` attribute into a Python `set` and joins it unsorted into the `LinkLibraries` compile option (`python/CuTeDSL/cutlass/base_dsl/dsl.py:2127-2134`, `main` @ dc45f97):

```python
sources = set(
    x.value for x in gpu_module.attributes.get("link-libraries", set())
)
link_libraries = (
    link_libraries
    + ("," if link_libraries and len(sources) > 0 else "")
    + ",".join(sources)          # <- set iteration order
)
```

Iteration order of a `set` of strings depends on `PYTHONHASHSEED`, which is randomized per process. Whenever a program links two or more bitcode libraries (one `link-libraries` entry is added per extern bitcode source, `ffi.py:603-611`), the rendered pipeline token `link-libraries='...'` (`StringCompileOption.serialize`, `compiler.py:413-416`) differs from process to process for the same program.

The attribute's writer already establishes a canonical order: `ffi.py:609-611` writes the ArrayAttr **sorted**. The `set()` in `_merge_gpu_link_libraries` then discards that order (and adds no dedup that `ffi.py` has not already done), so the sortedness maintained on the attribute never reaches the compiler invocation.

Order demonstration (pure CPython semantics, no GPU needed):

```console
$ for s in 0 1 2 3; do PYTHONHASHSEED=$s python -c "print(','.join(set(['/opt/libdevice.10.bc','/opt/libuser_ops.bc','/opt/libmath_ext.bc'])))"; done
/opt/libuser_ops.bc,/opt/libdevice.10.bc,/opt/libmath_ext.bc
/opt/libdevice.10.bc,/opt/libmath_ext.bc,/opt/libuser_ops.bc
/opt/libuser_ops.bc,/opt/libmath_ext.bc,/opt/libdevice.10.bc
/opt/libuser_ops.bc,/opt/libmath_ext.bc,/opt/libdevice.10.bc
```

## Why it matters

- The effective compiler invocation is not reproducible across processes for the same source, so byte-identical rebuilds cannot be relied on, and any consumer that records or hashes the effective pipeline string sees spurious per-process variation.
- Link order is semantically load-bearing in the usual first-definition-wins model: if two linked bitcode libraries ever provide the same symbol, which definition wins becomes a per-process choice. I have not identified a shipped pair that conflicts, so with today's inputs the practical effect may be limited to non-reproducible builds; filing accordingly rather than as a correctness bug.
- Process-scoped compile nondeterminism is a failure signature that is costly to debug downstream (for one open example of the class, flashinfer-ai/flashinfer#4704: a CuTe-DSL kernel's tests flip between all-pass and all-fail only across processes). I am not claiming this site explains that report; it is one confirmed instance of the class in the frontend.

## Expected behavior

The pipeline string for a fixed program should not depend on `PYTHONHASHSEED`:

```python
+ ",".join(sorted(sources))
```

mirroring `ffi.py:610`, or iterate the already-sorted ArrayAttr directly instead of re-wrapping it in a set.

## Environment / verification

- All line references verified against NVIDIA/cutlass `main` @ dc45f97 (`python/CuTeDSL` frontend source).
- Not executed against the installed `nvidia-cutlass-dsl` wheel: the order-dependence follows from CPython set semantics and the code path above; I could not run the DSL end-to-end here (no CUDA device).

migarci2:
## Summary

- keep the canonical order already stored in each GPU module's `link-libraries` attribute
- retain deduplication without relying on Python set iteration
- add a CPU-only regression test that preserves existing options and detects the old unordered path

## Testing

- `python test/python/CuTeDSL/test_link_libraries_order.py -v`
- `ruff check test/python/CuTeDSL/test_link_libraries_order.py`
- `ruff format --check test/python/CuTeDSL/test_link_libraries_order.py`
- `python3.12 -m py_compile python/CuTeDSL/cutlass/base_dsl/dsl.py test/python/CuTeDSL/test_link_libraries_order.py`
- `git diff --check`

## Contribution disclosure

AI assistance was used for repository research, implementation, independent review, and running the checks above.

Fixes #3564

### DIFF
--- a/test/python/CuTeDSL/test_link_libraries_order.py
+++ b/test/python/CuTeDSL/test_link_libraries_order.py
@@ -1,0 +1,52 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: LicenseRef-NvidiaProprietary
+#
+# Use of this software is governed by the terms and conditions of the
+# NVIDIA End User License Agreement (EULA), available at:
+# https://docs.nvidia.com/cutlass/latest/media/docs/pythonDSL/license.html
+#
+# Any use, reproduction, disclosure, or distribution of this software
+# and related documentation outside the scope permitted by the EULA
+# is strictly prohibited.
+
+import unittest
+from types import SimpleNamespace
+from unittest import mock
+
+from cutlass.base_dsl import dsl as dsl_module
+from cutlass.base_dsl.compiler import LinkLibraries
+
+
+class _ReverseSet(set):
+    def __iter__(self):
+        return iter(sorted(super().__iter__(), reverse=True))
+
+
+class TestLinkLibrariesOrder(unittest.TestCase):
+    def test_preserves_attribute_order_and_deduplicates(self):
+        paths = ["device.bc", "math.bc", "user.bc", "math.bc"]
+        gpu_module = SimpleNamespace(
+            name="gpu.module",
+            attributes={
+                "link-libraries": [SimpleNamespace(value=path) for path in paths]
+            },
+        )
+        module = SimpleNamespace(body=SimpleNamespace(operations=[gpu_module]))
+        compile_options = SimpleNamespace(
+            options={LinkLibraries: LinkLibraries("existing.bc")}
+        )
+        owner = SimpleNamespace(compile_options=compile_options)
+
+        # Make the old set-based path fail deterministically, independent of
+        # the interpreter's hash seed. The fixed path does not use this name.
+        with mock.patch.object(dsl_module, "set", _ReverseSet, create=True):
+            dsl_module.BaseDSL._merge_gpu_link_libraries(owner, module)
+
+        self.assertEqual(
+            compile_options.options[LinkLibraries].value,
+            "existing.bc,device.bc,math.bc,user.bc",
+        )
+
+
+if __name__ == "__main__":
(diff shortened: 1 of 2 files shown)


test/python/CuTeDSL/test_link_libraries_order.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+class _ReverseSet(set):
+    def __iter__(self):
+        return iter(sorted(super().__iter__(), reverse=True))
+
+
+class TestLinkLibrariesOrder(unittest.TestCase):
+    def test_preserves_attribute_order_and_deduplicates(self):
+        paths = ["device.bc", "math.bc", "user.bc", "math.bc"]
```
brandon-yujie-sun: This is already alphabetical sorted. Maybe change the order a bit so the assert gates both the order and dedup?

brandon-yujie-sun (approved):
Thanks for contributing the codes! LGTM with a minor comment.

brandon-yujie-sun (approved):
LGTM, thanks!

brandon-yujie-sun:
@Junkai-Wu to help further process thanks!

PR #3572, opened 2026-08-30
Fix cute.composition rejecting dynamic-shaped RHS modes with a ScaledBasis stride

### LINKED ISSUE #3470: [BUG] [CuTeDSL] cute.composition throws when rhs has a dynamic shape

CuTe DSL

### Bug Report

**Describe the bug**
Invoking cute.composition on a RHS layout that contains one mode with a dynamic size causes an exception to be raised. At first I thought that this was by design, because if cute.composition were to verify that RHS's codomain falls within LHS's domain, it needs to know RHS's domain. But apparently such a check isn't actually performed (not even for fully-static layouts).

```py
import torch
import cutlass.cute as cute


@cute.jit
def repro(tensor: cute.Tensor):
    lhs = cute.make_layout((4096, 8192), stride=(1, 4096))
    rhs = cute.make_layout(
        (32, tensor.shape[1] // 8, 8, 128),
        stride=(
            cute.E(0),
            8 * cute.E(1),
            cute.E(1),
            32 * cute.E(0),
        ),
    )
    print(cute.composition(lhs, rhs))


tensor = from_dlpack(torch.empty((4096, 8192), device="cuda"))
tensor = tensor.mark_compact_shape_dynamic(mode=1)
repro(tensor)
```

Raises:
```
cutlass._mlir._mlir_libs._site_initialize.<locals>.MLIRError: Operation creation failed:
error: unknown: unable to compute the following composition: '!cute.layout<"(4096,8192):(1,4096)">' o '!cute.layout<"(32,?,8,128):(1@0,8@1,1@1,32@0)">'
```

CuTeDSL version 4.7.0

saiyambharara:
## Summary
cute.composition raised an MLIRError when the RHS layout had a ScaledBasis stride (cute.E(...)) on a mode with a dynamic shape.

The check causing this required the mode's shape to be a static int, but that shape is never read in the ScaledBasis branch before the check. It is forwarded unchanged into a recursive composition_impl call, and every branch that call can land in already tolerates a dynamic shape: one copies the shape into the result without inspecting it, the other gates its static divisibility checks behind holds_int() and skips them when an operand is dynamic. Removing the check lets composition compute the result the same way it already does in every other dynamic-shape case.

Fixes #3470.

## Test plan
- Added two cases to CompositionTest.Basic in cg_composition_test.cpp, one minimal ScaledBasis plus dynamic shape case, one matching the layouts from the issue.
- Ran the full cutegen unit suite (24 binaries, about 200 tests) against the patch. All pass.

### DIFF
--- a/cutlass_compiler/cutegen/include/cutegen/layout.hpp
+++ b/cutlass_compiler/cutegen/include/cutegen/layout.hpp
@@ -1674,7 +1674,8 @@ layout_t<TDynTraits> composition_impl(const cute_shape_t<TDynTraits>&  lhs_shape
     // Special case for a RHS ScaledBasis stride
     if(holds_scaled_basis(rhs_stride))
     {
-        if(!holds_int(rhs_shape)) return cg_error_t{};
+        // rhs_shape may be dynamic: it's forwarded as-is to composition_impl below,
+        // whose remaining cases already tolerate a dynamic shape.
         using sb_t                 = typename find_scaled_basis_in_rec_var<stride_t>::value;
         const auto& sb             = std::get<sb_t>(rhs_stride);
         const auto& v              = sb.modes();
--- a/cutlass_compiler/cutegen/test/cg_composition_test.cpp
+++ b/cutlass_compiler/cutegen/test/cg_composition_test.cpp
@@ -415,6 +415,28 @@ TEST(CompositionTest, Basic)
         EXPECT_EQ(r, layout(shape(dyn_t{}), stride(dyn_t{})));
     }
     // -------------------------------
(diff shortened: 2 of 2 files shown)


saiyambharara:
@ccecka , @brandon-yujie-sun , @Junkai-Wu  this fixes the composition bug from #3470. Would appreciate a review when you have a chance.

Junkai-Wu:
@ccecka to take a look.

ccecka (approved):

saiyambharara:
ccecka has approved. @Junkai-Wu could you take a final look and merge when you get a chance? I am not authorized to push to main 
image

Issue #3527, opened 2026-08-24
[BUG | QST] CPU overhead regression with TVM-FFI


kainzhong:
### Which component has the problem?

CuTe DSL

### Bug Report

**Describe the bug**
In v4.6 update, the following is changed in `python/CuTeDSL/cutlass/cutlass_dsl/cutlass.py`:
```
-                if kwargs_wrapper_spec.kwonly_names or kwargs_wrapper_spec.arg_defaults:
+                # Route through the kwargs-capable compiled class whenever
+                # the signature has kwonly/defaults OR any positional arg
+                # carries a dataclass instance (detected by the spec
+                # converter, including Union[...] over dataclasses). The
+                # kwargs wrapper is the only place tvm-ffi's
+                # ``map_dataclass_to_tuple`` unpack hook fires.
+                if (
+                    kwargs_wrapper_spec.kwonly_names
+                    or kwargs_wrapper_spec.arg_defaults
+                    or kwargs_wrapper_spec.arg_names
+                    or map_dataclass_to_tuple
+                ):
```
So basically if a @cute.jit function satisfies one of the following:
- `kwonly_names`: have keyword only arguments (indicated by *)
- `arg_defaults`: have default arguments
- `arg_names`: have any argument that is not a keyword only argument
- `map_dataclass_to_tuple`: have dataclass arguments

With these restrictions (especially `kwonly_names` and `arg_names`), basically only a CuTeDSL function with no arguments can go to the else branch and become a `TVMFFIJitCompiledFunction` which is a subclass of `tvm_ffi.Function` so they get to be called in C++ and bypass python completely. For CuTeDSL functions with arguments, they now have to go through the `TVMFFIJitCompiledFunctionWithKwargs` path.

This will slightly slow down calling a CuTeDSL function from python due to extra python layers from in `TVMFFIJitCompiledFunctionWithKwargs`, which isn't a big deal since python is slow anyway. However, if you register the CuTeDSL function via tvm-ffi and call it from C++, the overhead becomes obvious because you will leave C++ and enter python (`TVMFFIJitCompiledFunctionWithKwargs.__call__`) and then enter the compiled CuTeDSL function. This makes the CPU overhead regress compared to pre-4.6 version since there wasn't going through python with GIL and argument wrapper stuff.

[...]

PR #3589, opened 2026-09-06
fix(CuTeDSL): optimize positional-only functions with TVMFFIJitCompiledFunction to eliminate CPU overhead (#3527)  

Ammar-Alnagar:
#### 1. What was the Issue? (#3527)
In CUTLASS v4.6, the routing condition in `python/CuTeDSL/cutlass/cutlass_dsl/cutlass.py` was updated to check:

```python
if (
    kwargs_wrapper_spec.kwonly_names
    or kwargs_wrapper_spec.arg_defaults
    or kwargs_wrapper_spec.arg_names
    or map_dataclass_to_tuple
):
    return TVMFFIJitCompiledFunctionWithKwargs(...)
```

Because `kwargs_wrapper_spec.arg_names` includes positional parameter names for every argument, any `@cute.jit` function taking 1 or more arguments evaluated to `True` and was routed to `TVMFFIJitCompiledFunctionWithKwargs`.

`TVMFFIJitCompiledFunctionWithKwargs` wraps invocation in `kwargs_wrapper.make_kwargs_wrapper`. When registered via TVM-FFI and called from C++, execution left C++ to enter Python (GIL + argument wrapper parsing), causing CPU overhead regression compared to direct C++ TVM-FFI function execution.

---

#### 2. What We Did

Positional-only parameters (`def func(a, b, c, /)`) cannot be called with keyword arguments in Python grammar. Therefore, functions with strictly positional-only arguments do not require keyword-argument wrapping.

1. **`python/CuTeDSL/cutlass/base_dsl/jit_executor.py`**:
   - Added `has_pos_or_kw` to `KwargsWrapperSpec`.
   - Updated `get_kwargs_wrapper_spec` to track whether any `Parameter.POSITIONAL_OR_KEYWORD` parameter exists.

2. **`python/CuTeDSL/cutlass/cutlass_dsl/cutlass.py`**:
   - Changed `kwargs_wrapper_spec.arg_names` check in `_make_compiled_func` to `kwargs_wrapper_spec.has_pos_or_kw`.
   - Functions defined as `def func(a, b, c, /)` without defaults or dataclasses now return `TVMFFIJitCompiledFunction` (a direct subclass of `tvm_ffi.Function`).

3. **Unit Tests & Documentation**:
   - Added unit test suite `test_tvm_ffi_kwargs_wrapper_spec.py`.
   - Updated documentation in `tvm_ffi_compilation.rst:64`.

---

#### 3. Comparison & Verification of Fix

| Function Signature          | has_pos_or_kw | arg_defaults | kwonly_names | Class Returned                        | CPU Overhead Behavior                                      |
|-----------------------------|---------------|--------------|--------------|---------------------------------------|------------------------------------------------------------|
| `def func()`                | False         | `()`         | `[]`         | `TVMFFIJitCompiledFunction`           | 0 Python Overhead (Direct C++ `tvm_ffi.Function`)          |
| `def func(a, b, c, /)` (Fix)| False         | `()`         | `[]`         | `TVMFFIJitCompiledFunction`           | 0 Python Overhead (Direct C++ `tvm_ffi.Function`, bypasses GIL/wrapper) |
| `def func(a, b, c)`         | True          | `()`         | `[]`         | `TVMFFIJitCompiledFunctionWithKwargs` | Handles Python kwargs calls (`func(a=1, b=2)`)             |
| `def func(a, b=1, /)`       | False         | `(1,)`       | `[]`         | `TVMFFIJitCompiledFunctionWithKwargs` | Handles positional default value binding                   |
| `def func(a, /, *, k=1)`    | False         | `()`         | `['k']`      | `TVMFFIJitCompiledFunctionWithKwargs` | Handles keyword-only argument binding                      |

##### Verification Summary

- Evaluated signature parsing with `inspect.signature` across positional-only, positional-or-keyword, and mixed argument patterns.
- Confirmed `def func(a, b, c, /)` correctly evaluates `check_routing()` to `False`, bypassing `TVMFFIJitCompiledFunctionWithKwargs` and directly instantiating `TVMFFIJitCompiledFunction`.
```

### CODE REVIEW

**Comment by Ammar-Alnagar** [comment]
Hey @kainzhong / @tqchen,

Following up on the earlier discussion ([https://github.com/NVIDIA/cutlass/issues/3527]):

> Hi @kainzhong, thanks for reaching out! This is a known issue for us. Your understanding is correct - the condition you mention is generally true, except the trivial functions without any args. But per our benchmark, the overhead it introduces is < 1us, while the overall launch overhead is typically <10 us. We think it is fine to keep this overhead for compatibility of diversed inputs. And we agree to have more documents here. Any PR is appreciated.
> cc: @tqchen

This PR addresses the routing issue so that pure positional-only functions (`def func(a, b, c, /)`) now correctly bypass `TVMFFIJitCompiledFunctionWithKwargs` and use the direct `TVMFFIJitCompiledFunction` path (zero Python/GIL overhead).

Would appreciate a review when you have a chance — thanks!

File: media/docs/pythonDSL/guides/tvm_ffi_compilation.rst [hunk 1]
```diff
@@ -61,6 +61,7 @@ To maximize performance benefits, we recommend setting up your workflow as follo
 - **Use the environment stream flag** to implicitly pass the current PyTorch stream.
 - **Rely on compiled argument validation** instead of Python-side attribute validation,
   as TVM FFI functions perform fast compiled checks.
+- **Use positional-only arguments (`def func(a, b, c, /)`)** when defining `@cute.jit` functions to bypass Python wrapper overhead completely and allow direct execution via `tvm_ffi.Function`.
```
kainzhong: I think adding some explanation here would be better. Also I think you also need to say don't use default values for params

test/python/CuTeDSL/test_tvm_ffi_kwargs_wrapper_spec.py
```diff
@@ -0,0 +1,44 @@
+# Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
kainzhong, line 1: I think a more proper way to test this is
- Create different variants of the `@cute.jit` function and compile it via tvm-ffi
- Assert which path they should take, like if the function object is `TVMFFIJitCompiledFunctionWithKwargs` or `TVMFFIJitCompiledFunction`
- Actually call that function (I think only doing this in python is enough) to make sure this whole path works

Also I feel it should be put under somewhere like `test/python/CuTeDSL`? (I'm not a CuTeDSL guy so maybe I'm wrong)

media/docs/pythonDSL/guides/tvm_ffi_compilation.rst
Ammar-Alnagar: good point.
 i expanded this section to explain that TVM FFI’s native ABI is positional-only, so a fully positional-only @cute.jit signature can use TVMFFIJitCompiledFunction directly and avoid the Python kwargs wrapper/GIL overhead.
i also documented the full fast-path contract: defaults, positional-or-keyword parameters, keyword-only parameters, and top-level dataclass arguments all require TVMFFIJitCompiledFunctionWithKwargs and  updated the primary guide example to use / as well.

test/python/CuTeDSL/test_tvm_ffi_kwargs_wrapper_spec.py
Ammar-Alnagar: i wasnt really sure at first  but i do agree with you on this  , the previous coverage only validated KwargsWrapperSpec parsing and would not catch a regression in _make_compiled_func. so i added the tests under test/python/CuTeDSL and now compile several real @cute.jit variants with --enable-tvm-ffi.
each test asserts the selected compiled-function class and invokes it with the intended calling convention: direct positional calls for the fast path, kwargs for named parameters, omitted defaults for positional-only defaults, and explicit keyword-only arguments. The focused suite passes (7 passed).

kainzhong:
Hi @cyx-6 can you take a look at this PR? My opinion is
- Should we make this behavior even more explicit? Now adding `/` can change the CPU overhead feels a bit black magic and I wonder if it would be better to add a param to `cute.compile` like `optimize_for_pos_only`
- I don't think `test/python/CuTeDSL/test_tvm_ffi_cpp_registry.py` is needed
- `test_kwargs_wrapper_spec_positional_only`, `test_kwargs_wrapper_spec_positional_or_keyword` and `test_kwargs_wrapper_spec_mixed_posonly_and_pos_or_kw` seem unnecessary as well

But I think that should be up to cutlass people to decide. What do you think?

cyx-6:
@kainzhong Thanks for contribution! This change looks good to me. And yes, I agree that some semantic changes should be discussed and decided by cutlass people.

Ammar-Alnagar:
> Hi @cyx-6 can you take a look at this PR? My opinion is
> 
> * Should we make this behavior even more explicit? Now adding `/` can change the CPU overhead feels a bit black magic and I wonder if it would be better to add a param to `cute.compile` like `optimize_for_pos_only`
> * I don't think `test/python/CuTeDSL/test_tvm_ffi_cpp_registry.py` is needed
> * `test_kwargs_wrapper_spec_positional_only`, `test_kwargs_wrapper_spec_positional_or_keyword` and `test_kwargs_wrapper_spec_mixed_posonly_and_pos_or_kw` seem unnecessary as well
> 
> But I think that should be up to cutlass people to decide. What do you think?

I can see where your point in the black box change and an input from one of the cutlass people would be much appreciated but i think the test cases don't hurt and are more of an assurance , but thats just my opinion and might not be the best flow for the project .
Curious on what you think? @cyx-6 @kainzhong

kainzhong:
Hi @Ammar-Alnagar unfortunately I'm not a cutlass developer. I think you probably should request their review instead since they are the ones who can approve it.

Ammar-Alnagar:
> Hi @Ammar-Alnagar unfortunately I'm not a cutlass developer. I think you probably should request their review instead since they are the ones who can approve it.

no worries , thanks alot for the help and comments !!!

Ammar-Alnagar:
@Junkai-Wu @brandon-yujie-sun @fengxie  , can i get a review on this whenever you get a chance ? , thanks!!!

PR #2917, opened 2026-01-01
New RMS Norm example with unit tests

bkryu:
The example showcases a memory bound rmsnorm kernel achieving above 6.5 TB/s on B200.

Unit test results:
```
$ pytest test/examples/CuTeDSL/sm_100a/test_rmsnorm.py -W ignore::pytest.PytestRemovedIn9Warning
==================================== test session starts ====================================
platform linux -- Python 3.10.12, pytest-9.0.2, pluggy-1.6.0
...
collected 79 items                                                                          

test/examples/CuTeDSL/sm_100a/test_rmsnorm.py ....................................... [ 49%]
........................................                                              [100%]

==================================== 79 passed in 5.64s =====================================
```

Benchmark results from calling the script on B200
```
$ python examples/python/CuTeDSL/blackwell/rmsnorm.py --M 32768 --N 16384 --dtype BFloat16 --benchmark
Running RMSNorm test with:
  M: 32768, N: 16384
  dtype: BFloat16
  has_weight: True
  eps: 1e-06
  SM version: 100
  cluster_n: 1
  threads_per_row: 128
  rows_per_block: 1
Correctness check passed!

Benchmarking with 2 warmup, 100 iterations...
Kernel execution time: 329.54 us
Memory throughput: 6516.67 GB/s
PASS

$ python examples/python/CuTeDSL/blackwell/rmsnorm.py --M 32768 --N 32768 --dtype BFloat16 --benchmark
Running RMSNorm test with:
  M: 32768, N: 32768
  dtype: BFloat16
  has_weight: True
  eps: 1e-06
  SM version: 100
  cluster_n: 2
  threads_per_row: 128
  rows_per_block: 1
Correctness check passed!

Benchmarking with 2 warmup, 100 iterations...
Kernel execution time: 612.90 us
Memory throughput: 7007.68 GB/s
PASS
```

PR #3593, opened 2026-09-07
[CuTeDSL] reduce example: replace inline PTX with cute.arch.store_async_dsmem

zkyue:
## Summary

The Blackwell `reduce.py` example still hand-rolls two cluster helpers, `set_block_rank` and `store_shared_remote`, that emit `mapa.shared::cluster` and `st.async.shared::cluster.mbarrier::complete_tx::bytes.f32` via `llvm.inline_asm`. That was necessary when the example landed (#2917), but `cute.arch` now exposes the same functionality as built-ins:

* `cute.arch.map_dsmem_ptr` — `mapa.shared::cluster`, returns a dsmem `Pointer`
* `cute.arch.store_async_dsmem` — maps destination + mbarrier to the peer CTA and issues `st.async` with `mbarrier::complete_tx::bytes`

This PR removes the inline PTX and calls the built-in directly from `cluster_reduce`.

## Changes

* Delete `set_block_rank` / `store_shared_remote` and the `cutlass._mlir.dialects.llvm` / `T` imports.
* `cluster_reduce` now calls
  `cute.arch.store_async_dsmem(elem_pointer(...), val.bitcast(Int32), mbar_ptr, peer_cta_rank=lane_idx)`.
  The built-in stores raw `b32` words, so the `Float32` is reinterpreted as `Int32` with `Numeric.bitcast` (an `arith.bitcast`, no data conversion). The transaction-byte accounting (`num_warps * cluster_n * 4`) and the mbarrier protocol are unchanged.
* Refresh the module docstring and the section comment; the Quack attribution is kept for the overall scheme.

Net: `1 file changed, 18 insertions(+), 96 deletions(-)`, and the example no longer contains any inline PTX.

## Notes

* `store_async_dsmem` was added in the v4.8 dev update, so the example now requires a 4.8+ `nvidia-cutlass-dsl` for the `cluster_n > 1` path. `map_dsmem_ptr` (4.6+) is not called directly because `store_async_dsmem` performs both `mapa` translations internally; keeping `set_block_rank` as a wrapper around it would have been dead code.
* `elem_pointer` is untouched.

## Testing

Run on a B200 (sm_100a, CUDA 13.3) with `nvidia-cutlass-dsl==4.8.0.dev0`:

```
CUDA_VISIBLE_DEVICES=7 python -m pytest test/examples/CuTeDSL/sm_100a/test_rmsnorm.py -v
======================= 79 passed, 22 warnings in 6.48s ========================
```

This includes `TestRMSNormClusterPath::test_cluster_path_correctness[32768]` and `[65536]`, which exercise exactly the changed path (`cluster_n > 1` in the RMSNorm example, which imports `row_reduce` from this file). All non-cluster RMSNorm tests pass as well.

PTX of the cluster kernel (`CUTE_DSL_KEEP=ptx`) shows the same instruction sequence as the removed inline asm, with the built-in emitting a `.b32` store instead of `.f32` (same bytes moved):

```
mapa.shared::cluster.u32   %r321, %r320, %r9;
mapa.shared::cluster.u32   %r323, %r322, %r9;
st.async.shared::cluster.mbarrier::complete_tx::bytes.b32  [%r321], %r8, [%r323];
```

Also checked statically against `cute.arch.store_async_dsmem`'s contract: both pointers are SMEM `Pointer`s from `SmemAllocator`, the value is a single `Int32` (4-byte alignment requirement satisfied by the `Float32` buffer), and `peer_cta_rank` is the lane index as before.

The two follow-up commits are doc-only (mbarrier arrival count is 1, lanes `0..cluster_n-1` each store, the usage example now mirrors the caller's init-fence / cluster-sync sequence).

### DIFF
--- a/examples/python/CuTeDSL/cute/blackwell/kernel/reduce/reduce.py
+++ b/examples/python/CuTeDSL/cute/blackwell/kernel/reduce/reduce.py
@@ -295,10 +295,16 @@ def cluster_reduce(
                 Float32
             )
 
-            # Initialize mbarrier (once per kernel); arrival count is 1 because
-            # a single elected thread does the arrive_and_expect_tx
+            # Initialize mbarrier (once per kernel, by a single thread); the
+            # arrival count is 1 because one elected thread does the
+            # arrive_and_expect_tx. Fence the init and sync the cluster before
+            # any CTA may store into a peer's buffer / mbarrier.
             mbar = cute.make_smem_tensor(cute.make_layout((1,)), cute.arch.Mbarrier)
-            cute.arch.mbarrier_init(mbar.iterator, 1)
+            if cute.arch.thread_idx()[0] == 0:
+                cute.arch.mbarrier_init(mbar.iterator, 1)
+            cute.arch.mbarrier_init_fence()
+            cute.arch.cluster_arrive_relaxed()
+            cute.arch.cluster_wait()
 
             # Perform cluster reduction
             result = cluster_reduce(
--- a/examples/python/CuTeDSL/cute/blackwell/kernel/reduce/reduce.py
+++ b/examples/python/CuTeDSL/cute/blackwell/kernel/reduce/reduce.py
@@ -81,7 +81,8 @@ Mbarrier Requirements (Cluster Reduction)
 -----------------------------------------
 For cluster reduction, the caller must:
 1. Allocate an mbarrier in shared memory
-2. Initialize it with `cute.arch.mbarrier_init(mbar_ptr, thread_count)`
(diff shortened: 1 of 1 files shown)


zkyue:
Tested on B200 with `nvidia-cutlass-dsl==4.8.0.dev0`: `test/examples/CuTeDSL/sm_100a/test_rmsnorm.py` passes (79 passed), including both `test_cluster_path_correctness` cases that go through the changed `cluster_reduce` path. PTX dump of the cluster kernel shows `mapa.shared::cluster.u32` x2 + `st.async.shared::cluster.mbarrier::complete_tx::bytes.b32`, i.e. the same sequence the inline asm used to emit. Updated the Testing section of the description accordingly.

PR #3607, opened 2026-09-10
[CuTeDSL] Prevent domain_offset overflow with static strides

### LINKED ISSUE #3604: [BUG] domain_offset overflows runtime Int32 coordinate with static stride

CuTe DSL

### Bug Report

**Describe the bug**

`cute.domain_offset` can overflow a runtime `Int32` coordinate when the tensor has a static stride. The element offset is calculated in 32 bits before conversion to a 64-bit byte offset, which can cause `CUDA_ERROR_ILLEGAL_ADDRESS`. The equivalent dynamic-stride layout uses 64-bit arithmetic and succeeds.

```python
import argparse
import os

os.environ.setdefault("CUDA_LAUNCH_BLOCKING", "1")

import torch
import cutlass
import cutlass.cute as cute


ROWS = 2**29 + 1
COLS = 4
LAST_ROW = ROWS - 1


# IMA: from_dlpack(out) preserves the fully static compact layout:
#   !cute.layout<"(536870913,4):(4,1)">
@cute.kernel
def no_make_layout_static_stride(tensor: cute.Tensor, row: cutlass.Int32):
    # IR emitted by domain_offset:
    #   %idx = cute.crd2idx(%coord, %lay)
    #       : (...) -> !cute.int_tuple<"?{div=4}">
    #   %ptr = cute.add_offset(%iter, %int_tuple)
    shifted = cute.domain_offset((row, 0), tensor)
    shifted[(0, 0)] = cutlass.BFloat16(1.0)


# Works: the runtime row stride remains i64.
@cute.kernel
def make_layout_dynamic_stride(tensor: cute.Tensor, row: cutlass.Int32):
    layout = cute.make_layout(
        (tensor.shape[0], COLS),
        stride=(tensor.stride[0], 1),
    )
    # IR for layout:
    #   %lay_3 = cute.make_layout(%shape, %stride)
    #       : !cute.layout<"(?,4):(?{i64},1)">
    rebuilt = cute.make_tensor(tensor.iterator, layout)
    # IR emitted by domain_offset:
    #   %idx = cute.crd2idx(%coord, %lay_3)
    #       : (...) -> !cute.int_tuple<"?{i64}">
    #   %ptr = cute.add_offset(%iter, %int_tuple)
    shifted = cute.domain_offset((row, 0), rebuilt)
    shifted[(0, 0)] = cutlass.BFloat16(1.0)


# IMA: make_layout replaces the dynamic stride with the static stride (4, 1).
@cute.kernel
def make_layout_static_stride(tensor: cute.Tensor, row: cutlass.Int32):
    layout = cute.make_layout(
        (tensor.shape[0], COLS),
        stride=(COLS, 1),
    )
    # IR for layout:
    #   %lay_0 = cute.make_layout(%shape, %stride)
    #       : !cute.layout<"(?,4):(4,1)">
    rebuilt = cute.make_tensor(tensor.iterator, layout)
    # IR emitted by domain_offset:
    #   %idx = cute.crd2idx(%coord, %lay_0)
    #       : (...) -> !cute.int_tuple<"?{div=4}">
    #   %ptr = cute.add_offset(%iter, %int_tuple)
    shifted = cute.domain_offset((row, 0), rebuilt)
    shifted[(0, 0)] = cutlass.BFloat16(1.0)


@cute.jit
def launch_no_make_layout_static_stride(tensor: cute.Tensor, row: cutlass.Int32):
    no_make_layout_static_stride(tensor, row).launch(grid=[1, 1, 1], block=[1, 1, 1])


@cute.jit
def launch_make_layout_dynamic_stride(tensor: cute.Tensor, row: cutlass.Int32):
    make_layout_dynamic_stride(tensor, row).launch(grid=[1, 1, 1], block=[1, 1, 1])


@cute.jit
def launch_make_layout_static_stride(tensor: cute.Tensor, row: cutlass.Int32):
    make_layout_static_stride(tensor, row).launch(grid=[1, 1, 1], block=[1, 1, 1])


cases = (
    "no_make_layout_static_stride",
    "make_layout_dynamic_stride",
    "make_layout_static_stride",
)
parser = argparse.ArgumentParser()
parser.add_argument("case", choices=cases)
args = parser.parse_args()

out = torch.zeros((ROWS, COLS), device="cuda", dtype=torch.bfloat16)
row = cutlass.Int32(LAST_ROW)

if args.case == "no_make_layout_static_stride":
    launch_no_make_layout_static_stride(from_dlpack(out), row)
elif args.case == "make_layout_dynamic_stride":
    launch_make_layout_dynamic_stride(from_dlpack(out).mark_layout_dynamic(), row)
else:
    launch_make_layout_static_stride(from_dlpack(out).mark_layout_dynamic(), row)

torch.cuda.synchronize()
assert float(out[0, 0]) == 0.0
assert float(out[-1, 0]) == 1.0
```

```console
$ python repro.py no_make_layout_static_stride
CUDA_ERROR_ILLEGAL_ADDRESS (error code: 700)

$ python repro.py make_layout_dynamic_stride
# succeeds

$ python repro.py make_layout_static_stride
CUDA_ERROR_ILLEGAL_ADDRESS (error code: 700)
```

**Expected behavior**

All three cases should write `out[-1, 0]` without an illegal memory access.

**Environment details**

- Docker
- `nvidia-cutlass-dsl==4.8.0a0+20260828210702.38664cd`
- CUDA 13.4
- GB300 (`sm_103a`)

**Additional context**

The static-stride path lowers the address calculation as:

```ptx
shl.b32      %r2, %r1, 2;
mul.wide.s32 %rd2, %r2, 2;
add.s64      %rd3, %rd1, %rd2;
```

For `row = 2**29`, the first instruction produces `0x80000000`; signed widening then generates a `-4 GiB` byte offset. BF16 is used here only to expose the incorrect index-width propagation through element-to-byte lowering; the coordinate-to-index issue is not BF16-specific. The dynamic-stride case keeps the index in 64 bits:

```ptx
mul.lo.s64 %rd4, %rd2, %rd3;
shl.b64    %rd5, %rd4, 1;
add.s64    %rd6, %rd1, %rd5;
```

kzos:
Fixes #3604.

## What this fixes

`cute.domain_offset` can calculate the wrong global-memory address even when
the runtime coordinate itself fits in `Int32`. A static stride does not widen
the coordinate before multiplication, so the **element offset** can overflow
before it is converted to a 64-bit **byte offset** for pointer addition.

This change widens dynamic coordinate leaves before `crd2idx` for global/generic
memory pointers. It fixes the reported static-stride and rebuilt-static-stride
cases without requiring callers to make their layouts dynamic or manually cast
coordinates to `Int64`. The code change is confined to `tensor.py`; this PR also
adds 10 CPU IR regression tests.

## Reproducer and failure mechanism

The [full standalone reproducer in #3604](https://github.com/NVIDIA/cutlass/issues/3604)
allocates a BF16 tensor with shape `(2**29 + 1, 4)` and compact strides `(4, 1)`,
then uses a runtime `Int32(2**29)` to address the last row. Its essential operation
inside the kernel is:

```python
# tensor: global BF16 tensor, layout (2**29 + 1, 4):(4, 1)
# row: runtime cutlass.Int32, value 2**29
shifted = cute.domain_offset((row, 0), tensor)
shifted[(0, 0)] = cutlass.BFloat16(1.0)
```

The coordinate is valid and fits in signed 32 bits. The intermediate product
does not:

| Address-calculation step | Intended value | Unpatched static-stride path |
| --- | --- | --- |
| Runtime row | `2**29` | `2**29` |
| Element offset: row times stride 4 | `2**31` | Overflows signed 32-bit arithmetic to `-2**31` |
| Byte offset: element offset times 2 for BF16 | `+2**32` (+4 GiB) | Signed wide multiplication produces `-2**32` (-4 GiB) |

The resulting store goes before the allocation rather than to its last row.
Casting the result of `crd2idx` to `Int64` would be too late: the narrow product
has already lost its intended value. BF16 exposes the byte-address consequence;
the coordinate-to-index overflow is not specific to BF16.

The issue's dynamic-stride variant succeeds because its runtime stride is
already `i64`, causing the coordinate/stride calculation to stay wide. Rebuilding
a layout with the static stride `(4, 1)` removes that wide operand and restores
the failure. It is the arithmetic width, not the presence of `make_layout`, that
distinguishes these cases.

## Why this change is placed in `domain_offset`

`domain_offset` is the point where the coordinate becomes an offset for a
particular tensor iterator. For a global/generic `Pointer`, the patch recursively
promotes narrow dynamic coordinate leaves to `Int64` **before** calling
`crd2idx`, so the coordinate/stride product is formed at the required width.

The implementation preserves the surrounding contracts:

- Python integer constants and `None` leaves stay unchanged, preserving static
  coordinate information and constant folding.
- Integer leaves already at least 64 bits wide are left alone. Signed narrow
  coordinates sign-extend; unsigned narrow coordinates zero-extend.
- `get_divisibility` and `assume` carry existing divisibility facts across the
  promotion, preserving the derived pointer-alignment information.
- `transform_leaf` handles nested coordinates rather than only flat tuples.
- Shared-memory pointers and non-pointer iterators, including identity tensors,
  retain their existing arithmetic. General-purpose `crd2idx` and CuTe integer
  algebra are not changed globally.

This addresses overflow introduced within this `domain_offset` calculation. It
cannot recover a coordinate that the caller has already overflowed, and it does
not make out-of-bounds coordinates valid. No performance improvement is claimed;
the effect of wider arithmetic on arbitrary kernels has not been benchmarked.

## Check against the executed native code

For the BF16 row/stride case above, the executed baseline kernel emits:

```ptx
shl.b32      %r2, %r1, 2;
mul.wide.s32 %rd2, %r2, 2;
```

Here the first instruction computes the element offset in 32 bits; the wide
multiply only widens that already-overflowed result. The fixed kernel instead
combines the stride and element size into a wide multiply of the original row:

```ptx
mul.wide.s32 %rd2, %r1, 8;
```

Disassembly of 60 repeated baseline/fixed binaries confirms the corresponding
32-bit shift is removed and the wide multiply changes from factor 2 to 8.
A semantically neutral source variant retains identical native instructions
within each arm and does not fix the baseline failure. Baseline memcheck reports
a 2-byte global store 4 GiB before the nearest allocation, matching the predicted
negative byte offset.

## Regression tests included in this PR

Run with an environment importing the changed CuTeDSL source:

```console
python test/python/CuTeDSL/test_domain_offset.py
```

These tests build and verify MLIR on the CPU; they do not allocate a multi-GiB
tensor or require a GPU. **10/10 pass with the fix, versus 5/10 on the baseline.**
They check widening before `crd2idx`, global/generic pointers, nested coordinates,
divisibility/alignment preservation, unsigned extension, static constants,
already-wide coordinates, dynamic-i64 strides, and unchanged shared-memory and
identity-tensor behavior. These are IR regressions, not a substitute for runtime
validation.

## Additional B300 validation

A separate runtime harness exercised the real installed package on B300 SM103,
with source hashes checked against the baseline and this patch. Each case used
a fresh process and private compile/cache directory. Baseline failures were
isolated because an illegal address invalidates the CUDA context.

- Static-stride, rebuilt-static-layout, and neutral-source variants each produced
  **10/10 baseline failures and 10/10 fixed passes**.
- A 29-entry fixed matrix ran both normally and under Compute Sanitizer. It
  covers BF16/FP32/Int32 elements; row offsets `0`, `2`, `2**29 - 1`, `2**29`,
  `2**29 + 1`, and `2**30`; and dynamic-i64, explicit-i64, nested, constant,
  neutral, and valid negative-offset cases. This is a selected matrix, not the
  Cartesian product of every option.
- All **29 fixed memcheck cases reported zero errors**. A separate baseline
  memcheck reproduced the invalid global store described above.
- **590 fixed-path graph replays and 40 baseline-control replays passed**.
  Each replay poisoned the tensor buffer and checked that only the intended
  element changed. Captured scalar arguments stayed fixed; mutation of those
  arguments was not tested.

Across the campaign: **90 fixed cases passed, 5 baseline controls passed, and
32 expected baseline failures were reproduced (127 total cases)**. Memcheck
cases and graph replays are included in those runs, not extra independent cases.
The GPU harness/results are additional validation, not tests added to this PR
or a report of upstream CI passing.

<details>
<summary>Toolchain and existing suite limitations</summary>

Validation used public `nvidia-cutlass-dsl==4.8.0.dev0` with matching base/core/cu12
libraries, Torch 2.11.0+cu130, cuda-bindings 13.3.1, B300 driver 580.126.09, and
Compute Sanitizer 2025.4.1. This is not the exact dated alpha wheel in the report.
The unpatched wheel's `tensor.py` is byte-identical to the base source at
`147295a3`; the fixed installed module is byte-identical to this PR's version.

The broader public CuTeDSL Python suite is **39/43 passing with this change,
versus 34/43 on the baseline**. Both retain the same four unrelated failures/errors:

- `test_exec_defined_function_raises_user_error`
- `test_uncaught_dsl_error_does_not_kill_interactive_session`
- `test_preserves_attribute_order_and_deduplicates`
- `test_fastmath_flag`

No unrelated changes are included. Both final GPU arms used lazy CUDA module
loading; earlier eager-loading attempts timed out in PyTorch CUDA initialization
before CuTe compilation/launch and are not counted as kernel results.

</details>


kzos:
Hi @hwu36, could you help route this `domain_offset` fix for CuTeDSL review? It widens dynamic coordinates before static-stride multiplication for global/generic pointers, with shared-memory and non-pointer behavior unchanged. The PR includes 10 CPU IR regression tests and records separate baseline/fixed B300, memcheck, and CUDA Graph validation, including the remaining broader-suite limitations. Happy to address review feedback; no performance improvement is claimed. Thanks!

hwu36:
@brandon-yujie-sun

Issue #3640, opened 2026-09-17
[BUG] Operand initialized with a constant doesn't work with `inline_ptx` as an operand


kainzhong:
### Which component has the problem?

CuTe DSL

### Bug Report

**Describe the bug**
In CuTeDSL, if you define a variable as a constant like `zero = Int32(0)`, `inline_ptx` lowers `Int32(0)` to the immediate operand 0. This causes ptxas to reject instructions whose operand grammar requires a register, such as `max.xorsign.abs.bf16x2`. There is currently no apparent way to request register materialization for a constant passed through `read_only_args`.

**Steps/Code to reproduce bug**
```python
import cutlass
from cutlass import cute
from cutlass import Int32


@cute.kernel
def kernel():
    value = cute.arch.thread_idx()[0]
    zero = Int32(0)
    cute.arch.inline_ptx(
        "max.xorsign.abs.bf16x2 {$w0}, {$r0}, {$r1};",
        write_only_types=[Int32],
        read_only_args=[zero, value],
    )

@cute.jit
def launch():
    kernel().launch(grid=(1, 1, 1), block=(32, 1, 1))

if __name__ == "__main__":
    cute.compile(launch)
```

Outputs:
```
error: PTX assembly failed

  --> /home/kainingz/GitHub/TransformerEngine/repro_cutedsl_inline_ptx_constant.py:21:0
   |
  19 | 
  20 | 
> 21 | @cute.kernel
     | ^
  22 | def kernel():
  23 |     value = cute.arch.thread_idx()[0]
  error: ptxas rejected the PTX generated for this kernel while compiling it to SASS.
  note: target architecture: sm_100
  note: ptxas log:
        ptxas application ptx input, line 22; error   : Arguments mismatch for instruction 'max'
        ptxas fatal   : Ptx assembly aborted due to errors
  suggestion: read the ptxas log above; re-run with CUTE_DSL_KEEP=ptx and assemble the dumped PTX
              with `ptxas -arch=sm_100 <dumped>.ptx` to iterate on the failure
```

Generated PTX:
```assembly
//
// Generated by NVIDIA NVVM Compiler
//
// Compiler Build ID: CL-38501229
// Cuda compilation tools, release 13.4, V13.4.46
// Based on NVVM 23.0.0
//

.version 9.4
.target sm_100
.address_size 64

	// .globl	kernel_cutlass_kernel_0

.visible .entry kernel_cutlass_kernel_0()
.reqntid 32, 1, 1
[...]

PR #3642, opened 2026-09-18
[CuTe DSL] Allow selected read-only inline_ptx operands to be materialized into a register

0z5a:
## Summary

`cute.arch.inline_ptx` lowers a compile-time constant read-only operand to the immediate constraint `n` regardless of what the target instruction accepts. Some PTX instructions take only a register in a given operand slot, so such a constant is rejected by ptxas. This adds `force_register_args`, an opt-in that names the read-only operands which must be moved into a register first.

Refs #3640.

## Minimal failure

```python
@cute.kernel
def kernel():
    value = cute.arch.thread_idx()[0]
    zero = Int32(0)
    cute.arch.inline_ptx(
        "max.xorsign.abs.bf16x2 {$w0}, {$r0}, {$r1};",
        write_only_types=[Int32],
        read_only_args=[zero, value],
    )
```

Generated PTX on an L20 (sm_89, `nvidia-cutlass-dsl 4.8.0.dev0`):

```ptx
mov.u32 	%r3, %tid.x;
max.xorsign.abs.bf16x2 %r1, 0, %r2;      // immediate operand
```

```
ptxas application ptx input, line 32; error   : Arguments mismatch for instruction 'max'
ptxas fatal   : Ptx assembly aborted due to errors
```

## Where the constraint is chosen

Not in this repository. `NVVM_InlinePtxOp` carries no per-operand register request (only `readOnlyArgs`, `readWriteArgs`, `ptxCode`, `predicate`, `memoryClobber`). The constraint comes from the NVVM to LLVM conversion, `mlir/lib/Dialect/LLVMIR/IR/BasicPtxBuilderInterface.cpp`:

```cpp
static FailureOr<char> getRegisterType(Value v, Location loc) {
  if (v.getDefiningOp<LLVM::ConstantOp>())
    return 'n';
  return getRegisterType(v.getType(), loc);
}
```

The value-level overload returns `n` for any constant without consulting the instruction, so the generated `llvm.inline_asm` gets the constraint string `"=r,n,r"`.

## What this changes

`inline_ptx(..., force_register_args=(0,))` materializes the named read-only operands with an explicit `mov` before the target instruction runs:

```ptx
mov.b32 %r2, 0;
max.xorsign.abs.bf16x2 %r1, %r2, %r3;
```

which is the output #3640 asks for. The operand the target instruction sees is produced by an instruction rather than by a constant op, so the existing type-based register constraint applies. Nothing else changes: the default is `None`, unnamed operands keep today's behavior, and indices refer to `read_only_args` only — out-of-range, negative, bare-int and non-integer indices raise.

## Why not disable immediates globally

A constant is legal, and preferable, in plenty of operand slots (`add.s32`, `mbarrier.init.b64`, ...). Making `getRegisterType(Value)` return `r` for every constant would be a semantics change well beyond this issue. The opt-in is therefore per call site, and the existing immediate path is untouched — covered by `test_default_path_is_unchanged` and by the `add.s32` cases in the verification below.

## Verification

Hardware: 1× NVIDIA L20, compute capability 8.9, driver 570.86.10, `CUTE_DSL_ARCH=sm_89`.
Environment: `nvidia-cutlass-dsl 4.8.0.dev0` (cu12 libraries), PTX generated by CUDA 13.4.46 / NVVM 23.0.0. Baseline is `f614dc40e17fb3ddb1b7b474318b48a8d5d21a2c`.

End-to-end path (Python DSL → lowering → PTX → ptxas → launch → host readback), 32 lanes of packed bf16 covering ±0, ±1, ±2 and ±0.5, compared as raw bit patterns:

| Variant | Compile | Bit-exact readback |
|---|---|---|
| `Int32(0)` as reported | FAIL — `Arguments mismatch for instruction 'max'` | — |
| constant read from memory (pre-change workaround) | PASS | 32/32 |
| `llvm.inline_asm` with explicit `=r,r,r` | PASS | 32/32 |
| `force_register_args=(0,)` | PASS | 32/32 |

Regression matrix, all executed on the device: literal zero; non-zero and negative finite constants; a constant produced by folding; the constant in the second read-only position; two identical constants with only one opted in; a read-only placeholder repeated twice; dynamic operands with and without the opt-in; an instruction where an immediate is legal, with and without the opt-in; two write-only results with a forced constant; and the argument-validation cases. The new test file covers the same IR shapes on CPU, so it needs neither a GPU nor ptxas.

Timing was measured with CUDA events in an interleaved paired run (60 rounds memory-bound, 40 rounds register-resident) including an A/A arm that recompiles and re-measures the patched variant under a second label. The fixed kernel matches the existing workarounds within the A/A noise band (about ±1.2%); the materialized `mov` is hoisted out of the loop and absorbed by register allocation. This is a correctness change, not a throughput change.

Two failures observed while testing are **not** addressed here, and reproduce without this change: `inline_ptx(predicate=...)` emits `@0` instead of the predicate's operand index, and a float read-only constant fails with `constraint 'n' expects an integer constant expression`. Float operands are therefore rejected by `force_register_args` with an explicit error rather than silently ignored.

## Scope and open questions

This is a CUTLASS-side wrapper. The complete fix belongs in LLVM/MLIR — a per-operand register request on `NVVM_InlinePtxOp` honoured by `BasicPtxBuilderInterface` — which then has to reach a release wheel. This repository has no copy of that C++, and the wheel ships a statically linked compiler, so that path cannot be validated from here.

The parameter name is a proposal. Feedback on the name, on whether the `mov` materialization is the right shape, or on waiting for the compiler-side attribute instead, is welcome.

Not tested: sm_100 (the reported target; this host is sm_89), and NaN/subnormal bit-level semantics.


0z5a:
### Timing

This is a correctness fix, so the point of these numbers is that it costs nothing at runtime — not that it is faster.

Hardware: 1× NVIDIA L20, compute capability 8.9, driver 570.86.10, `CUTE_DSL_ARCH=sm_89`. Workloads:

- `mem` — memory-bound elementwise kernel over 16,777,216 packed-bf16x2 carriers (64 MiB read + 64 MiB write), grid 8192 × block 256, 8 carriers per thread.
- `alu` — register-resident loop, 262,144 threads × 8 accumulators × 2048 iterations = 4.29e9 instructions.

Timing uses CUDA events. Every arm is compiled up front, then measured in **interleaved rounds** — each round visits all arms once in a rotating order — 60 rounds for `mem`, 40 for `alu`. `patched_aa` is the patched variant recompiled and measured under a second label: an A/A control that bounds the noise floor of the harness itself.

| Arm | `mem` median (ms) | `mem` effective BW (GB/s) | `alu` median (ms) | `alu` throughput (Gops/s) |
|---|---|---|---|---|
| constant read from memory (pre-change workaround) | 0.198656 | 675.6 | 0.345568 | 12 428.7 |
| `llvm.inline_asm` with explicit `=r,r,r` | 0.199168 | 673.9 | 0.345088 | 12 446.0 |
| `force_register_args=(0,)` | 0.199680 | 672.2 | 0.345088 | 12 446.0 |
| `patched_aa` — A/A control, same code as the row above | 0.199680 | 672.2 | 0.345088 | 12 446.0 |

Paired per-round ratios against the workaround arm:

| Comparison | `mem` geomean ratio | `alu` geomean ratio |
|---|---|---|
| explicit `=r,r,r` / workaround | 1.00321 | 1.00006 |
| `force_register_args` / workaround | 1.00469 | 0.99687 |
| **A/A control / workaround** | **1.00957** | **1.01157** |

The A/A control deviates from the reference arm by *more* than the patched variant does (+0.96% / +1.16% versus +0.47% / −0.31%), so none of the difference is attributable to this change — everything sits inside roughly a ±1.2% noise band. That matches expectation: the materialized `mov` is loop-invariant and absorbed by register allocation.

The host is shared with other workloads, which is why the arms are paired and interleaved rather than measured in sequence. Outputs were compared as raw bit patterns, not with a tolerance; see the readback table in the description.

python/CuTeDSL/cutlass/cute/arch/nvvm_wrappers.py
```diff
@@ -4590,6 +4667,7 @@ def inline_ptx(
     read_only_args: Optional[list[ScalarArg]] = None,
     read_write_args: Optional[list[ScalarArg]] = None,
     predicate: Optional[Boolean] = None,
+    force_register_args: Optional[Sequence[int]] = None,
```
kainzhong, line 4670: I feel like if you do this wouldn't it be neater?
So instead of `force_register_args` which uses indices to represent args, probably we can just ask user to write something like `"max.xorsign.abs.bf16x2 {$w0}, {$n0}, {$r0};"` with an extra param: `immediate_args=[]`,  where stuff like `{$nX}` means the X-th immediate arguments and `{$rX}` means the X-th read registers. I think this is more explicit and it should be compatible to your current change.

python/CuTeDSL/cutlass/cute/arch/nvvm_wrappers.py
```diff
@@ -4611,6 +4689,14 @@ def inline_ptx(
     :type read_write_args: list[ScalarArg], optional
     :param predicate: Optional Boolean value for conditional execution (corresponds to @$p prefix).
     :type predicate: Boolean, optional
+    :param force_register_args: Optional indices into ``read_only_args`` whose
```
kainzhong: I think you probably also need to state the restrictions of "only i1, i16, i32, i64 and pointers" are supported here, and I think the description here could be more concise

test/python/CuTeDSL/test_inline_ptx_register_operand.py
```diff
@@ -0,0 +1,228 @@
...
+
+from cutlass._mlir import ir
+from cutlass._mlir.dialects import func
+from cutlass import Int32, Int64
+from cutlass.cute.arch.nvvm_wrappers import inline_ptx
+
+_PTX_MAX = "max.xorsign.abs.bf16x2 {$w0}, {$r0}, {$r1};"
+
+
+def _build(body, arg_types=("i32", "i32")):
```
kainzhong: I think you should just test end to end instead of using this `_build`, so you just use `@cute.jit` to compile real kernels and be sure that the whole path is safe.

kainzhong (changes requested):
Apart from my comments, I think you also need to expose this new interface to https://docs.nvidia.com/cutlass/latest/media/docs/pythonDSL/primitives.html#cutlass.experimental.primitives.inline_ptx which is the one listed in CuTeDSL's documentation.

However, I'm not a cutlass dev so you probably should @ one of them to get an approval

python/CuTeDSL/cutlass/cute/arch/nvvm_wrappers.py
0z5a: I checked whether this is reachable from Python, and it is not. `{$wX}` / `{$rX}` / `{$rwX}` are substituted by the `InlinePtxOp` lowering, and the op takes exactly three operand groups - `inline_ptx(write_only_args, read_only_args, read_write_args, ptx_code, *, predicate=None)`. So `{$nX}` plus `immediate_args` is a fourth group on the op: an ODS change and a matching update to the NVVM-to-LLVM conversion, plus a rebuild, none of which this Python file can carry.

The constraint an operand gets is chosen in that same lowering, from the defining op (constant -> `n`), so spelling the string differently cannot force `r`. The only Python-side lever is to make the operand stop being a constant op, which is what the `mov` behind `force_register_args` does.

Two ways forward, your call:
1. Keep `force_register_args` for this PR as the Python-only opt-in, and I open a follow-up for the `{$nX}` / `immediate_args` group - once that lands, the parameter can be deleted.
2. Hold this PR and land the MLIR group first, if you would rather not add a Python-only parameter that the group would replace.

python/CuTeDSL/cutlass/cute/arch/nvvm_wrappers.py
0z5a: Done: the docstring now names the operand types that can be materialized (i1, i16, i32, i64 and pointers) and says the rest in three lines instead of six.

test/python/CuTeDSL/test_inline_ptx_register_operand.py
0z5a: Rewritten in 46c2035: the IR-shape tests and `_build` are gone. The file now compiles and launches real `@cute.jit` kernels - the issue's `max.xorsign.abs.bf16x2` case with the constant in the register-only slot, including host readback of a bit-exact result, plus a `mov.b64` case for a wider operand. The index-validation cases stay, as plain calls that assert the raise.

Ran on an RTX 5090 with `CUTE_DSL_ARCH=sm_120`: **6 passed**. One caveat about what that run covers - the wheel matching this branch (4.8.0.dev0) is not on that host, so I ran against the installed 4.6.2 DSL with this PR's `nvvm_wrappers` change ported into it. That validates the test file and the fix's Python side end to end through ptxas, not the exact wheel. Happy to re-run on the right wheel if you can point me at one, or to leave it to CI.

python/CuTeDSL/cutlass/cute/arch/nvvm_wrappers.py
kainzhong: I may not have explained my suggestion clearly. I’m not proposing any change to `InlinePtxOp` or involving any MLIR stuff.
Since this wrapper is executing Python code, immediate_args could be a virtual Python-level group. For example:
```
{$w0}, {$n0}, {$r0}
read_only_args=[x],
immediate_args=[y],
```
could be rewritten in Python to:
```
{$w0}, {$r0}, {$r1}
read_only_args=[y, x],
force_register_args=[1],
```
so this is just a Python interface-level desugaring.

Using indices to specify register operands feels a bit fragile to me so I wonder if it's better to be more explicit.
However I’m not a CUTLASS developer, so I think we should let them decide. This is only a proposal and I may be overlooking something.

(Separately these responses feel very claude-ish to me. I'm fine with agents but I think the author should at least show some participation to indicate they know what their agents are doing -- that's just my personal thought

0z5a:
@Junkai-Wu PTAL.

